PR #10, opened 2017-05-19
fixes #2 consolidate / reorganize config data

### LINKED ISSUE #2: consolidate / reorganize config data
right now, stuff is in package.scala, application.conf, config-template.conf, and config.xml

Options:

- put everything in config-template.conf 
- keep separation between beam stuff and akka stuff (but keep config files in same directory)
- keep separation between application config (e.g. akka but also stuff related to deployment at scale) and model scenario config (e.g. parameters / inputs)

dserdyuk:
- Added new consolidated beam.conf that is environment specific and contains beam-specific props, Akka config and MATSim module configuartion properties
- Added initialization code of MATSim config from typesafe config. You can add any MATSim module to beam.conf
- beam.conf has been loaded as classpath resourse (sf-bay added as additional source folder in build.gradle)
- overall refactoring of ConfigModule and configuration injection via Guice


colinsheppard:
Instead of injecting BeamConfig into actors (e.g. see PersonAgent), please add "beamConfig" as a member to the AgentsimServices companion object which is accessible from anywhere with an import.

dserdyuk:
ok. I will revert as it was before.

dserdyuk:
@colinsheppard.  Btw, how about to keep beamConfig in  ConfigModule companion object ( as it is now) and import ConfigModule instead of AgentsimServices if you need beamConfig ?

colinsheppard:
I think we can get the best of both here at the slight cost of purity (I'm personally ok with impure code if it's useful). Let's keep BeamConfig as an injected object, but also add it to AgentsimServices companion. So in the BeamConfig companion:

`  var beamConfig: BeamConfig = _ `

Then in the class:

```
case class AgentsimServices @Inject()(protected val injector: Injector, theBeamConfig: BeamConfig) extends ActorInject {
  val matsimServices: MatsimServices = injector.getInstance(classOf[MatsimServices])
  AgentsimServices.beamConfig: BeamConfig = theBeamConfig
}
```

Then we can remove actually injecting this into actors.

Can you also do the same thing for matsim services? Thanks!

PR #11, opened 2017-05-24
Update users.rst

zishanbilal:
removed broken link:
You can download [templates of the required input files here](https://www.dropbox.com/s/hya6dlw969d3jfw/beam-inputs-template.zip?dl=1).


colinsheppard:
The dropbox link you added is broken.

zishanbilal:
@colinsheppard i removed the link, i added it in comment just for reference. If we have any fresh link please share so i can update that. Otherwise we may remove the link.

PR #13, opened 2017-05-30
Point to point

ahmar-nadeem-nor:


ahmar-nadeem-nor:
Point to Point functionality added to the beam project. There are not conflicts and can be merged

zishanbilal:
I have already added R5 as dependency, @ahmar-nadeem-nor please use that as library to perform build process. And you don't need to add any third-party code. If you need to change something in R5 or any thirdparty then please discuss first.
Please remove the code from R5 and then update the PR.

src/main/java/com/conveyal/r5/analyst/GridCache.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.zip.GZIPInputStream;
+
+/**
+ * Cache opportunity density grids on S3, with a local cache for performance when reusing the same grids. Each instance
+ * works with a single S3 bucket.
+ */
+public class GridCache {
+    private static final Logger LOG = LoggerFactory.getLogger(GridCache.class);
+
+    private AmazonS3 s3 = new AmazonS3Client();
```
dserdyuk: I would pass client outside to be able to control credentials

ahmar-nadeem-nor:
Well, that has been updated. At first I tried adding the dependency but my gradle got messed up and it never built my project(neither in eclipse nor IntelliJ). So, I moved the classes there. But now my workspace is clean and gradle working smooth.

src/main/java/beam/playground/point2point/PointToPointMain.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+public class PointToPointMain {
+
+    private static final Logger LOG = LoggerFactory.getLogger(PointToPointMain.class);
+
+    public static void main(String directoryPath[]) {
+
+        final boolean inMemory = false;
+
+        File dir = new File(directoryPath[0]);
```
zishanbilal: Shouldn't check if there is any argument, it could cause potentially null pointer exception.

zishanbilal: In case of null argument some default value should automatically be considered.

src/main/java/beam/playground/point2point/PointToPointMain.java
```diff
@@ -0,0 +1,37 @@
...
+
+        if (!dir.isDirectory() && dir.canRead()) {
```
zishanbilal: +1

src/main/java/beam/playground/point2point/PointToPointMain.java
```diff
@@ -0,0 +1,37 @@
...
+import com.conveyal.r5.transit.TransportNetwork;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import java.io.File;
+
+/**
+ * Created by ahmar.nadeem on 5/28/2017.
+ */
+public class PointToPointMain {
```
zishanbilal: Shouldn't the name to be more specific, e.g; NetworkBuilder or something that reflect the purpose.

colinsheppard:
So is the only new contribution the PointToPointMain class? 

If so, please do not merge this branch. Rather, backup the new class and just directly add to r5-router branch. Then delete this branch which now contains all of the data associated with the R5 classes which will only bloat our repository.

ahmar-nadeem-nor:
Agreed with Colin. I will just cut a new branch off r5-router and add the main class to it. Will create the PR for that branch.

colinsheppard:
Ok thanks Ahmar.

PR #15, opened 2017-06-01
added network builder functionality

ahmar-nadeem-nor:


zishanbilal:
Add some test for this.

zishanbilal:
As discussed please break into two public functions

- buildNetwork
take path and write network on disk
- varifyNetwork
check if network.dat and mapdb already exist at provided path, and return true if finds

main method use them and do the rest.

ahmar-nadeem-nor:
Do you think in that case we will still need the main method? I prefer removing the main method and call the functionality from a test class.

PR #27, opened 2017-06-21
Dserdiuk/vehicle step2

### LINKED ISSUE #18: Vehicles Part 2 - Managers and inputs
Develop an input file format. For now, let's keep it simple and specify as 2 CSV files, vehicles.csv and vehicle-relations.csv. The vehicles would cover basic properties of each vehicle and the vehicle-relations.csv would assign vehicles to their owners.

Vehicles can only be owned by one VehicleManager, which is an abstract trait inherited by households and fleet managers (which are also BeamAgents). These classes need to be scaffolded as well.

dserdyuk:
Fixes #18 
* added initial version of  vehiclemanager and household 
* refactoring of BeamVehicle to meet VehicleManager requirements
* added simple app to see how we can initialize vehicle network


src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -1,46 +1,122 @@
...
+    * MATSim vehicle vehicle implementation class
+    * @return
+    */
+  def vehicleClassName: String
+}
+
+object BeamVehicle {
 
-trait BeamVehicle[T] extends Identifiable[T]{
+  def energyByType(vehicleTypeId: Id[VehicleType]): Double = {
```
colinsheppard: What does the Double here represent?

colinsheppard: Ok, I see now, it's energy per unit length. Can you rename to "energyPerUnitByType"?

PR #31, opened 2017-07-02
for #18 Vehicles Part 2 - Managers and inputs

dserdyuk:
* associate the appropriate PersonAgent ActorRefs
* reinit beamvehicles and household actors on each iteration

### DIFF
--- a/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicleAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicleAgent.scala
@@ -15,26 +15,29 @@ import scala.concurrent.ExecutionContext.Implicits.global
   */
 object BeamVehicleAgent {
 
-  def props(vehicleId: Id[Vehicle], matSimVehicle: Vehicle, trajectory: Trajectory, powertrain: Powertrain) = {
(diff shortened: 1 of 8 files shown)


src/main/scala/beam/agentsim/agents/vehicles/BeamVehicleAgent.scala
```diff
@@ -1,16 +1,28 @@
...
+
 import scala.concurrent.ExecutionContext.Implicits.global
 
 /**
   *
   * @author dserdiuk
   */
+object BeamVehicleAgent {
+
+  def props(vehicleId: Id[Vehicle], matSimVehicle: Vehicle, trajectory: Trajectory, powertrain: Powertrain) = {
```
colinsheppard: Yes, trajectory will be set by the driver during execution, not on construction.

src/main/scala/beam/sim/BeamServices.scala
```diff
@@ -29,4 +31,6 @@ case class BeamServices @Inject()(protected val injector: Injector) extends Acto
   var schedulerRef: ActorRef =_
   var taxiManager: ActorRef = _
   var popMap: Option[Map[Id[Person], Person]] = None
```
colinsheppard: please refactor popMap to "persons" and remove the Option wrapper to make consistent with vehicles and households.

PR #43, opened 2017-08-04
Implemented DrivingMode trait and defined transit vehicle reservation protocol

dserdyuk:


src/main/scala/beam/agentsim/agents/modalBehaviors/DrivingMode.scala
```diff
@@ -0,0 +1,132 @@
...
+  * @author dserdiuk on 7/29/17.
+  */
+object DrivingMode {
+  case class CompleteLegTrigger(tick: Double, beamLeg: BeamLeg) extends Trigger
+  implicit val beamLegOrdering: Ordering[BeamLeg] = Ordering.by(_.startTime)
+
+}
+
+trait DrivingMode extends BeamAgent[PersonData] with TriggerShortcuts with HasServices with AggregatorFactory {
+  this: PersonAgent =>
```
colinsheppard: If possible I'd like this to be independent of PersonAgent, we will want non-persons to be able to drive. E.g. for now our transit drivers won't be persons and we will likely want to create AI drivers in the future as well.

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivingMode.scala
```diff
@@ -0,0 +1,132 @@
...
+
+  type PassengerLogStorage = mutable.TreeMap[BeamLeg, mutable.ListBuffer[ReservationLogEntry]]
+
+  case class ReservationLogEntry(departFrom: BeamLeg, arriveAt: BeamLeg,  passenger: ActorRef) {
```
colinsheppard: This isn't the intended meaning of a BeamLeg. We need to use something that signifies a transit stop instead.

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivingMode.scala
```diff
@@ -0,0 +1,132 @@
+package beam.agentsim.agents.modalBehaviors
```
colinsheppard: Please rename DrivingMode to "DrivesVehicle"

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivingMode.scala
colinsheppard: I realize that functionally they can be used to represent the same thing, but BeamLegs might contain relatively large amounts of data (because they can have a GraphPath in them) which will make equality comparisons MUCH more expensive than a simple Id which is all we really need.

colinsheppard:
@dserdyuk I just deleted a couple of comments in my initial review They were off the mark especially after having spent some additional time thinking about our traveling and driving protocols.

But I will have more feedback on the DrivesVehicle trait soon so stay tuned.

colinsheppard:
@dserdyuk I have substantially revised the protocols around traveling and driving. See:

http://beam.readthedocs.io/en/latest/protocols.html

We may want to discuss this. 

@zishanbilal note that this will require changes again to BeamLeg.

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivingMode.scala
dserdyuk: Basically, I need to have beamvehicle (actorref or id) to be able to send EnterVehicleTrigger and LeaveVehicleTrigger to it. 
That's why I restricted to PersonAgent to get currentVehicle from agentData. 
Alternatively, I could try to grab vehicleId from completedLeg because it's optional.  That should give more flexibility in base class/trait for DrivesVehicle but requires some additional (not-defined yet) logic to handle case when completedLeg  doesn't have beamVehicleId

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivingMode.scala
dserdyuk: Done

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivingMode.scala
dserdyuk: Alternative implementation is done.

dserdyuk:
@colinsheppard  I guess we can abandon this since you have merged these changes in your branch, right ?

colinsheppard:
Agreed. All of your changes fed into the PR I made.

PR #44, opened 2017-08-06
Agentsim drives vehicle

colinsheppard:
I've done some significant work on DrivesVehicle and BeamVehicle to make things consistent with the revised protocol I published on Friday. I will keep adding to this PR over the next day or two, but wanted to give you a chance to review before I merge into agentsim.


src/main/scala/beam/agentsim/agents/modalBehaviors/DrivesVehicle.scala
```diff
@@ -0,0 +1,186 @@
...
+
+  }
+
+  private def processNextLegOrCompleteMission() = {
+    val theTriggerId = _currentTriggerId.get
+    _currentTriggerId = None
+    val theTick = _currentTick.get
+    _currentTick = None
+
+    passengerSchedule.schedule.remove(passengerSchedule.schedule.firstKey)
```
dserdyuk: why do we do this ? Do you want to remove visited/completed stop  from schedule ?

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -47,160 +58,153 @@ trait VehicleData extends BeamAgentData with Vehicle {
...
-  def passengers: List[ActorRef]
+  var carrier: Option[ActorRef] = None
+  var passengers: ListBuffer[Id[Vehicle]] = ListBuffer()
+  var trajectory: Option[Trajectory] = None
 
-  def data: VehicleData
-
-  def trajectory: Trajectory
-
-  def powerTrain: Powertrain
```
dserdyuk: I would leave def instead of var in trait to have better immutability. Concrete implementation will manage state of trajectory internally by handling appropriated event

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivesVehicle.scala
```diff
@@ -0,0 +1,186 @@
...
+  this: BeamAgent[T] =>
+
+  //TODO: double check that mutability here is legit espeically with the schedules passed in
+  protected var passengerSchedule: PassengerSchedule = PassengerSchedule()
+
+  protected var _currentTriggerId: Option[Long] = None
+  protected var _currentTick: Option[Double] = None
+  protected var _currentLeg: Option[BeamLeg] = None
+  protected var _currentVehicle: Option[BeamVehicleIdAndRef] = None
+  protected val _awaitingBoardConfirmation: HashSet[Id[Vehicle]] = HashSet()
```
dserdyuk: Correct me if I'm wrong.. So these hashsets will contain ids of vehicle ( including humanbodyvehicle)  only of the passengers that's is about to board/alight rather than all passengers during trip ? My point is there is no way we could end up with id replacement in those hashsets

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -47,160 +58,153 @@ trait VehicleData extends BeamAgentData with Vehicle {
...
       } else {
-        // is it allowed to travel without driver ? auto pilot?
-        stay()
+        val leftSeats = fullCapacity - passengers.size
+        val beamAgent = sender()
+        beamAgent ! VehicleFull(id)
       }
+      stay()
+    case Event(UpdateTrajectory(newTrajectory), info) =>
+      trajectory = Some(newTrajectory)
```
dserdyuk: I thought we need to append each leg to trajectory rather then reset it ?

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -47,160 +58,153 @@ trait VehicleData extends BeamAgentData with Vehicle {
 }
 
 object BeamVehicle {
+  val ActorPrefixName = "vehicle-"
+
+  def props(vehicleId: Id[Vehicle], matSimVehicle: Vehicle, powertrain: Powertrain) = {
+    Props(classOf[BeamVehicle], vehicleId, VehicleData.vehicle2vehicleData(matSimVehicle), powertrain, None)
```
dserdyuk: are you sure you can use trait in props ?

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -195,8 +195,8 @@ class BeamSim @Inject()(private val actorSystem: ActorSystem,
 
   private def initVehicleActors(iterId: Option[String] = None) = {
     val actors = services.vehicles.map { case (vehicleId, matSimVehicle) =>
-      val props = BeamVehicleAgent.props(vehicleId, matSimVehicle, new Powertrain(BeamVehicle.energyPerUnitByType(matSimVehicle.getType.getId)))
-      val beamVehicle = actorSystem.actorOf(props, BeamVehicle.buildActorName(vehicleId, iterId))
+      val props = BeamVehicle.props(vehicleId, matSimVehicle, new Powertrain(BeamVehicle.energyPerUnitByType(matSimVehicle.getType.getId)))
```
dserdyuk: I guess we need to add ref into beamServices.agentRefs here

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivesVehicle.scala
colinsheppard: Yeah, it makes it easier to reason about the schedule if we remove & forget about completed legs. Also will help conserve on memory footprint.

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
colinsheppard: My thinking here is to avoid code duplication. Instead of treating the trait like an interface I'm treating it like a superclass and storing / handling data that all vehicles need here.

src/main/scala/beam/agentsim/agents/modalBehaviors/DrivesVehicle.scala
colinsheppard: Yes, the list of all passengers is in passengerSchedule.schedule.get(BeamLeg).manifest.riders.

These hashsets are just for the very temporary purpose of keeping track of which boarders / alighters have yet to confirm.

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
colinsheppard: Good point, I like that idea better.

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
colinsheppard: You're probably right on this. Either we only define props in concretes or we pass the class in here.

src/main/scala/beam/sim/BeamSim.scala
colinsheppard: Yes, I left that chore for this week while we work toward getting this code to actually run...

Issue #48, opened 2017-08-06
Complete Traveling Implementation


colinsheppard:
See:
http://beam.readthedocs.io/en/agentsim-drives-vehicle/protocols.html

And note that work is already under way under PR:

https://github.com/LBNL-UCB-STI/beam/pull/44

GanttStart: 2017-08-06
GanttDue: 2017-08-11

colinsheppard:
New subtasks:

1. Re-integrate Reservation request handling in BeamVehicle (from earlier work by @dserdyuk), should forward to the driver if a driver exists and buffer requests if no driver present (then unwind the buffer when a driver appears).
2. Re-integrate Reservation handling in DrivesVehicle (also based on earlier work by @dserdyuk). 
3. Append to vehicle trajectory rather than replace
4. Fix Actor creation "props" in BeamVehicle to take the concrete class, not trait.
5. Make sure all indices of Objects and ActorRefs in BeamServices are actually being built.

dserdyuk:
@colinsheppard subtasks have been done. The only problem was with personId-> vehicleId convertion.. I had to add "body-" prefix for humanbody vehicleId because the is overlap between person's and real vehicle ids. It leads to actor name duplication error.

dserdyuk:
@colinsheppard please review and close it if there is no addition subtask. it looks finished for me.

colinsheppard:
@dserdyuk 

I can now confirm that we have agents moving around successfully in the model, but only for walking and driving. I need to work through any issues for public transit, which will also coincide with resolving issue #51, so getting close.

PR #61, opened 2017-08-12
for #48 Complete Traveling Implementation

dserdyuk:
* set carrier
* send NotifyLegEnd/NotifLegStart directly to PersonAgent

### DIFF
--- a/src/main/scala/beam/agentsim/agents/PersonAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/PersonAgent.scala
@@ -186,6 +186,13 @@ class PersonAgent(val beamServices: BeamServices, override val id: Id[PersonAgen
       logError(s"Unrecognized message ${msg}")
       goto(Error)
   }
+  whenUnhandled{
+    case ev@Event(_, _) =>
+      handleEvent(AnyState, ev)
+    case msg@_ =>
+      logError(s"Unrecognized message ${msg}")
+      goto(Error)
+  }
 
   chainedWhen(Uninitialized){
     case Event(TriggerWithId(InitializeTrigger(tick), triggerId), _) =>
--- a/src/main/scala/beam/agentsim/agents/PersonAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/PersonAgent.scala
@@ -308,9 +308,9 @@ class PersonAgent(val beamServices: BeamServices, override val id: Id[PersonAgen
             }else{
               processedData.nextVehicleAssignment.beamVehicleId
             }
+            //TODO the following needs to find all subsequent legs in currentRoute for which this agent is driver and vehicle is the same...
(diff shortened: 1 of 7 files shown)


src/main/scala/beam/agentsim/agents/modalBehaviors/DrivesVehicle.scala
```diff
@@ -47,17 +47,12 @@ trait DrivesVehicle[T <: BeamAgentData] extends  TriggerShortcuts with HasServic
...
           _awaitingAlightConfirmation ++= manifest.alighters
-          manifest.riders.foreach(passenger => beamServices.vehicleRefs(passenger) ! NotifyLegEnd)
+          manifest.riders.foreach(pv => beamServices.personRefs.get(pv.personId).foreach( _  ! NotifyLegEnd))
           stay()
         case None =>
           log.error(s"Driver ${id} did not find a manifest for BeamLeg ${_currentLeg}")
           goto(BeamAgent.Error)
       }
-    case Event(ReservationRequestWithVehicle(req, vehicleIdToReserve), _) =>
-      require(passengerSchedule.schedule.nonEmpty, "Driver needs to init list of stops")
```
colinsheppard: This requirement is not general enough and highlights an issue with how ReservationRequest and handleVehicleReservation are currently working. When a Person makes a ReservationRequest, they should be passing some number of BeamLegs with the request, then we search the PassengerSchedule with the BeamLegs to determine whether there is sufficient capacity.

This process needs to work regardless of whether driving a Bus (which has a fixed schedule) or is a Taxi where the schedule is totally dynamic... so if the BeamLeg doesn't already exist in the PassengerSchedule, we should have an abstract method that validates whether it's ok to add that unrecognized BeamLeg.

Then we implement the logic of that method in the concrete class because the constraints around adding new BeamLegs will vary depending on the type of driver (e.g. a TransitDriver may never allow new BeamLegs but a PersonAgent may always allow since this is simply how we populate the PassengerSchedule with the data necessary to drive the vehicle).

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -298,9 +308,12 @@ class PersonAgent(val beamServices: BeamServices, override val id: Id[PersonAgen
             }else{
               processedData.nextVehicleAssignment.beamVehicleId
             }
-            passengerSchedule.addPassenger(vehicleId,Vector(processedData.nextLeg))
+            //TODO the following needs to find all subsequent legs in currentRoute for which this agent is driver and vehicle is the same...
```
colinsheppard: Please note this TODO hasn't been completed but should... just scan all upcoming Legs until the BeamVehicleAssignment either says asDriver==F or the beamVehicleId does not match the vehicle she is about to BecomeDriver of.

PR #67, opened 2017-08-16
Agentsim traveling

colinsheppard:
All, I'm going to do this merge in the next day but wanted to give you a chance to see everything that has changed (a lot) and give feedback.

At this point, the agent sim runs to completion (at least using the very small population and inputs inside "dev" scenario) and agents successfully route a trip, become driver of their Body and then subsequently become driver of their CAR which is managed by their household, and travel from their first to second activity.

There's a lot left to do but this should be helpful for others to run a somewhat functioning model and develop against.


src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -78,8 +66,8 @@ object BeamVehicle {
       0.0
   }
 
-  def buildActorName(vehicleId: Id[Vehicle]): String = {
-    s"$ActorPrefixName${vehicleId.toString}"
+  def buildActorName(matsimVehicle: Vehicle): String = {
+    s"$ActorPrefixName${matsimVehicle.getType.getDescription}-${matsimVehicle.getId.toString}"
```
dserdyuk: @colinsheppard  Do we really need description in actor name ? it may be to long

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -155,58 +153,120 @@ trait BeamVehicle extends Resource with  BeamAgent[VehicleData] with TriggerShor
...
+
   chainedWhen(Idle) {
     case Event(BecomeDriver(tick, newDriver, newPassengerSchedule), info) =>
-      if(driver.isEmpty) {
-        driver = Some(beamServices.agentRefs(newDriver))
-        driver.get ! BecomeDriverSuccess(newPassengerSchedule)
+      if(driver.isEmpty || driver.get == beamServices.agentRefs(newDriver.toString)) {
+        if (driver.isEmpty) {
+          driver = Some(beamServices.agentRefs(newDriver.toString))
+          if (newDriver.isInstanceOf[Id[Person]]) beamServices.agentSimEventsBus.publish(MatsimEvent(new PersonEntersVehicleEvent(tick, newDriver.asInstanceOf[Id[Person]], id)))
```
dserdyuk: good catch!!

src/main/scala/beam/agentsim/agents/vehicles/HumanBodyVehicle.scala
```diff
@@ -1,50 +1,79 @@
...
+  def createId(personId: Id[Person]) : Id[Vehicle] = {
+    Id.create("body-" + personId.toString, classOf[Vehicle])
+  }
 
-object HumanBodyVehicle {
-  //TODO make HumanDimension come from somewhere
-  def props(services: BeamServices, personId: Id[PersonAgent]) = Props(classOf[HumanBodyVehicle],services,
-    HumanBodyVehicleData(personId, HumanDimension(1.7, 60.0)), new Trajectory(), None)
+  def isHumanBodyVehicle(beamVehicleId: Id[Vehicle]) = beamVehicleId.toString.toLowerCase.contains("body")
+  val placeHolderBodyVehilceId: Id[Vehicle] = Id.create("body",classOf[Vehicle])
```
dserdyuk: Could you give more details on this ? why is it without numeric id ? I assume we can failed to parse personId from such placeholder down the line

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -126,15 +120,19 @@ class BeamSim @Inject()(private val actorSystem: ActorSystem,
   }
 
   private def cleanupVehicle() = {
-    for ( (_, actorRef) <- vehicleActors) {
-      logger.debug(s"Stopping ${actorRef.path.name} ")
+    logger.info(s"Stopping  BeamVehicle actors")
```
dserdyuk: info is too much here, we will have tons of events

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
colinsheppard: We can use the "description" field however we want, so I'm planning to use it like a name or code field (e.g. "CAR", "BIKE").

src/main/scala/beam/agentsim/agents/vehicles/HumanBodyVehicle.scala
colinsheppard: You're helping me change my mind on this. 

My original thinking was that since we're already treating HumanBody and walking as a special case during traveling, we can avoid the extra overhead of passing the BodyVehicleId from the person through the RoutingRequest only to be folded into the EmbodiedBeamTrip and passed back to the person.... instead we just use a dummy placeholder that the Person can look for and substitute their body Id.... but now that I've completed the extra work of folding car Id and bike Id into the RoutingResponse, I'm realizing that passing the Body vehicle doesn't really add anything and would remove the confusion of this dummy Id.... so I will change as a TODO.

src/main/scala/beam/sim/BeamSim.scala
colinsheppard: Agreed, there is a lot of logging everywhere that falls into the same category. But I'd rather clean up logging after we've gotten our near-term features implemented since they aid in debugging...

Issue #54, opened 2017-08-06
Transit Vehicle Synthesis


colinsheppard:
Generate transit vehicles from R5 TransportNetwork.

GanttStart: 2017-08-14
GanttDue: 2017-08-25

colinsheppard:
@neverforgit  FYI - I need transit vehicles in order to test other functionality, so I'm going to start pushing commits toward this end and will keep you posted. Let me know if you end up ready to work on this in the meantime.

colinsheppard:
I'm assigning this to @dserdyuk because completing this is critical for him to continue working through transit vehicle reservation protocol. 

FYI - I partially made it through the creation of transit vehicles but didn't complete. On agentsim branch, see R5Routingworker.initTransitVehicles()

The PassengerSchedule is being created properly (though now fully tested) and so the next step is to actually create the vehicles and the TransitDriverAgents and add them to the appropriate indexes in BeamServices (vehicles/vehicleRefs/agentRefs).

Then you should be able to the finalize the Reservation request protocol which should be initiated from ChoosingMode after the choice has been made and before the Person schedules PersonDepartureTrigger. If the result comes back Failure, the Person should make the choice again with a new Trip (though as a first step it's ok to just send the Person to Error state).

Finally, be aware that once a Person chooses a transit trip and starts traveling, there may be some debugging to do to get them to move through the traveling protocol correctly. In particular, I haven't tested out the NotifyLegStart/NotifyLegEnd logic, so it could have some mistakes.

PR #69, opened 2017-08-21
for  #54 Transit Vehicle Synthesis

dserdyuk:
* initial version
* create partial transit by sending message to the Router
* beamAgent state transition fixes

### DIFF
--- a/src/main/scala/beam/agentsim/agents/BeamAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/BeamAgent.scala
@@ -75,9 +75,6 @@ trait BeamAgent[T <: BeamAgentData] extends LoggingFSM[BeamAgentState, BeamAgent
   private val chainedStateFunctions = new mutable.HashMap[BeamAgentState, mutable.Set[StateFunction]] with mutable.MultiMap[BeamAgentState,StateFunction]
   final def chainedWhen(stateName: BeamAgentState)(stateFunction: StateFunction): Unit = {
     chainedStateFunctions.addBinding(stateName,stateFunction)
-    when(stateName, null) { case event =>
-      handleEvent(stateName, event)
-    }
   }
 
   def handleEvent(state: BeamAgentState, event: Event): State = {
--- a/src/main/scala/beam/agentsim/agents/BeamAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/BeamAgent.scala
@@ -73,8 +73,12 @@ trait BeamAgent[T <: BeamAgentData] extends LoggingFSM[BeamAgentState, BeamAgent
   protected var _currentTick: Option[Double] = None
 
   private val chainedStateFunctions = new mutable.HashMap[BeamAgentState, mutable.Set[StateFunction]] with mutable.MultiMap[BeamAgentState,StateFunction]
-  final def chainedWhen(stateName: BeamAgentState)(stateFunction: StateFunction): Unit =
+  final def chainedWhen(stateName: BeamAgentState)(stateFunction: StateFunction): Unit = {
     chainedStateFunctions.addBinding(stateName,stateFunction)
+    when(stateName, null) { case event =>
+      handleEvent(stateName, event)
+    }
+  }
 
   def handleEvent(state: BeamAgentState, event: Event): State = {
     var theStateData = event.stateData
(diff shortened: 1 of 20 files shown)


src/main/scala/beam/agentsim/agents/BeamAgent.scala
```diff
@@ -73,8 +73,12 @@ trait BeamAgent[T <: BeamAgentData] extends LoggingFSM[BeamAgentState, BeamAgent
   protected var _currentTick: Option[Double] = None
 
   private val chainedStateFunctions = new mutable.HashMap[BeamAgentState, mutable.Set[StateFunction]] with mutable.MultiMap[BeamAgentState,StateFunction]
-  final def chainedWhen(stateName: BeamAgentState)(stateFunction: StateFunction): Unit =
+  final def chainedWhen(stateName: BeamAgentState)(stateFunction: StateFunction): Unit = {
     chainedStateFunctions.addBinding(stateName,stateFunction)
+    when(stateName, null) { case event =>
+      handleEvent(stateName, event)
+    }
```
colinsheppard: Please revert this addition. The point of the _chainedWhen_ is as a replacement to _when_, not a wrapper for it. If multiple partials are registered via _when_ to the same state, only one the matching cases will run. _chainedWhen_ is designed so that all matching cases in the registered partials will run. 

If there are new code blocks not running when you expected them, you may be missing the boilerplate in the class or trait that handles events for a particular state, like this:

```
when(Initialized){
    case ev @ Event(_,_) =>
      handleEvent(stateName, ev)
  }
```

src/main/scala/beam/agentsim/agents/TransitDriverAgent.scala
```diff
@@ -30,25 +34,15 @@ class TransitDriverAgent(val beamServices: BeamServices,
...
-  }
-
   chainedWhen(Uninitialized){
-    case Event(TriggerWithId(InitializeTrigger(tick), triggerId), info: BeamAgentInfo[TransitDriverData]) =>
-      beamServices.schedulerRef ! CompletionNotice(triggerId)
-      goto(Idle)
+    case Event(InitializeTrigger(tick), info: BeamAgentInfo[TransitDriverData]) =>
+      logDebug(s" $id has been initialized, going to Waiting state")
+      data.vehicleUnderControl.ref ! BecomeDriver(tick, id, Option(data.passengerSchedule))
+      val firstStop = data.passengerSchedule.getStartLed()
```
colinsheppard: typo, should be "getStartLeg"

src/main/scala/beam/sim/BeamServices.scala
```diff
@@ -36,14 +38,14 @@ case class BeamServices @Inject()(protected val injector: Injector) extends Acto
...
   var taxiManager: ActorRef = _
-  var persons: Map[Id[Person], Person] = Map()
-  var personRefs: Map[Id[Person], ActorRef] = Map()
-  var vehicles: Map[Id[Vehicle], Vehicle] = Map()
-  var vehicleRefs: Map[Id[Vehicle], ActorRef] = Map()
-  var households : Map[Id[Household], Household] = Map()
-  var householdRefs : Map[Id[Household], ActorRef] = Map()
-  var agentRefs: Map[String, ActorRef] = Map()
-  var transitVehiclesByBeamLeg: Map[BeamLeg, Id[Vehicle]] = Map()
+  val persons: mutable.Map[Id[Person], Person] = new ConcurrentHashMap[Id[Person], Person]().asScala
```
colinsheppard: I'm unfamiliar with ConcurrentHashMap but why create a Java object and immediately convert to mutable.Map? Why no just create a mutable.Map?

dserdyuk: those maps can be used/mutated ( and  agentRefs and vehiclerefs are been mutated) from different actors ( hence different threads) 

transitVehiclesByBeamLeg is exception, it's been populated in one actor and been used in read-only mode

dserdyuk: @colinsheppard  yeah, Scala has basically only wrapper around java concurrent map

dserdyuk:
@colinsheppard  Code review changes has been applied

sfwatergit:
I had some further ideas on improving the functionality of
chainedWhen/handle Event that could make error reporting more informative
as well as better express the intent of this computation. Essentially, this
would involve using the Validation trait from scalaz to process the state
chain and collect errors sequentially. This is a form of the Either
construct (monad), so can be used similarly. I can play around with some
implementations and see if I can figure anything useful out.

Issue #49, opened 2017-08-06
Transit Fares


colinsheppard:
Figure out how to make transit fares get loaded and used in R5 and configure this for all transit agencies in the SF Bay Area.

Sum these fares and include as cost of BeamTrip.

GanttStart: 2017-08-06
GanttDue: 2017-08-18

colinsheppard:
I'm getting NoSuchElement exception after merging your latest updates into agentsim. It looks like the fares aren't created with non-transit modes, so when we run EmbodiedBeamTrip.embodyWithStreetVehicles it is attempting to access elements of the fares data that aren't present. At the least we should be adding fares with 0.0 as the value.

Also, I'm ok with using tuples for locally associating data, but when we are passing data from one method to another, (e.g. from buildResponse up to calcRoute) the use of tuples is very difficult to follow. Please create a new data structure with descriptive names so it is clear what is there and why.

In this case, the output of buildResponse could be something like: TripsAndFares(trips: Vector[BeamTrip], faresByLeg: Vector[Map[Int,Option[Double]]])

PR #70, opened 2017-08-21
#49 Router transit fares

zishanbilal:
Use GTFS feeds to load fares and its rules for transit fare calculation.


src/main/scala/beam/router/RoutingModel.scala
```diff
@@ -92,6 +92,7 @@ object RoutingModel {
   case class BeamLeg(startTime: Long,
                      mode: BeamMode,
                      duration: Long,
+                     fare: Double = 0.0,
```
colinsheppard: Fare should be a part of an EmbodiedBeamLeg and _not_ a BeamLeg. I will explain my reasoning on this when I have a chance, but for now please move (or I think fare is already there).

zishanbilal: Integrated Leg cost with fare.

zishanbilal:
@colinsheppard please review, fare/cost removed from BeamLeg and moved to EmbodiedBeamLeg

Issue #76, opened 2017-08-29
Integration Test Framework


colinsheppard:
Get preliminary integration testing framework up and running. More details to come but for now the first 2 tests should just verify that the model runs to completion without any uncaught exceptions and that the events file is complete (i.e. contains both <events> and </events>) tags.

PR #80, opened 2017-08-30
#76 - Integration tests for running beam

fdariasm:
Running BEAM without errors, testing event file creation.
RunBeam object modified to reuse initialization logic in tests.


colinsheppard:
When I try to compile I get this exception:

```
/Users/critter/Dropbox/ucb/vto/beam-all/beam/src/test/scala/beam/integration/Integration.scala:16: not found: type RunBeam
class Integration extends WordSpecLike with Matchers with RunBeam{
```

colinsheppard:
Also, your commit is now behind some changes we made to the file path of beamville, which is now located under test/input/beamville. Can you please update?

fdariasm:
@colinsheppard, updated for new path for beamville. Not sure why you get that compilation error, I created a trait RunBeam at RunBeam.scala link: https://github.com/LBNL-UCB-STI/beam/blob/fdariasm/testing_framework/src/main/scala/beam/sim/RunBeam.scala, please check that IDE gets the change.

colinsheppard:
Ok, it compiles for me know, getting this when trying to run:

```
Unable to load a Suite class. This could be due to an error in your runpath. Missing class: beam.integration.Integration
java.lang.ClassNotFoundException: beam.integration.Integration
	at java.net.URLClassLoader.findClass(URLClassLoader.java:381)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:424)
	at sun.misc.Launcher$AppClassLoader.loadClass(Launcher.java:331)
	at java.lang.ClassLoader.loadClass(ClassLoader.java:357)
	at org.scalatest.tools.Runner$.$anonfun$doRunRunRunDaDoRunRun$6(Runner.scala:1213)
	at org.scalatest.tools.Runner$.$anonfun$doRunRunRunDaDoRunRun$6$adapted(Runner.scala:1211)
	at scala.collection.TraversableLike.$anonfun$filterImpl$1(TraversableLike.scala:248)
	at scala.collection.immutable.List.foreach(List.scala:378)
	at scala.collection.TraversableLike.filterImpl(TraversableLike.scala:247)
	at scala.collection.TraversableLike.filterImpl$(TraversableLike.scala:245)
	at scala.collection.AbstractTraversable.filterImpl(Traversable.scala:104)
	at scala.collection.TraversableLike.filter(TraversableLike.scala:259)
	at scala.collection.TraversableLike.filter$(TraversableLike.scala:259)
	at scala.collection.AbstractTraversable.filter(Traversable.scala:104)
	at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:1211)
	at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24(Runner.scala:1031)
	at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24$adapted(Runner.scala:1010)
	at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:1500)
	at org.scalatest.tools.Runner$.runOptionallyWithPassFailReporter(Runner.scala:1010)
	at org.scalatest.tools.Runner$.run(Runner.scala:850)
	at org.scalatest.tools.Runner.run(Runner.scala)
	at org.jetbrains.plugins.scala.testingSupport.scalaTest.ScalaTestRunner.runScalaTest2(ScalaTestRunner.java:138)
	at org.jetbrains.plugins.scala.testingSupport.scalaTest.ScalaTestRunner.main(ScalaTestRunner.java:28)
```

fdariasm:
Hi @colinsheppard, check if you have  Integration class at test/scala/beam/integration.

colinsheppard:
It's something about my configuration... IDEA is treating the whole test folder as Java only and not recognizing the scala tests... 

How are you running the test, by running the individual class or whole suite? Are you doing it through IDEA or through gradle?

fdariasm:
I'm running the individual class through IDEA, just right click and run, you might need to set PWD env variable if not loaded. Check if file is recognized as Scala class on IDEA.

colinsheppard:
Figured it out... in case others experience, I was configured to have IDEA delegate running to gradle but that was breaking when trying to run tests, so I unchecked that option and it works now:

image

colinsheppard:
Thanks @fdariasm I see the test run now. All 3 assertions are failing, I would only expect one of them to fail (the one about uncaught exceptions), the 2 about the events file should pass. Maybe the test is looking in the wrong place? I still have a BEAM_OUTPUTS environment variable set, so the test should be robust to that.

Finally, the test should probably also be robust to the fact that in our configuration settings, we can change the format of the events output (i.e. from "xml" to "xml.gz" or later on to "csv" or "csv.gz").

fdariasm:
@colinsheppard can you check if PWD value is correct for the test? if the first test fail the other two will probably also fail because it's looking for latest folder created in output directory to check for event file. Also, if you are on my branch, it shouldn't be failing I have tried both on Windows and Mac. What error do you get ?

colinsheppard:
If PWD was not set, then I would get an error much earlier on, but the simulation is running just fine until the JsonFriendlyEventWriterXML class tried to write outputs. I'm still getting a fail on all three tests.

Here is the console output:

https://www.dropbox.com/s/j4o9isxxvq4eo2h/integrationTestOutput.log?dl=0

colinsheppard:
When I run all 9 tests fail. This is because the method of finding the right sub-directory under ConfigModule.beamConfig.beam.outputs.outputDirectory is not robust to the outputs I have in my directory (from many different runs with different prefixes -- i.e. not just "beamville" but others).

Best would be to pull the actual output directory that is used inside the model run. But second best would be to filter only sub-directories that begin with the following plus an underscore.

ConfigModule.beamConfig.beam.agentsim.simulationName

PR #92, opened 2017-09-10
Dserdiuk/beam path transit

dserdyuk:


colinsheppard:
These commits have not been merged indirectly into application-sfbay and then upstream into agentsim and master.

PR #94, opened 2017-09-12
Made changes for event log controller

avinashchavanfidel:
Hi @colinsheppard ,

I have made changes in code to make the logger meaningful as well as I have shorter the event logs in xml file according to each level .
I have attached each level event log xml file for your reference. Please review and let me know your thoughts about each level with the attribute which I have covered.

[event_logs.zip](https://github.com/LBNL-UCB-STI/beam/files/1295828/event_logs.zip)


src/main/java/beam/agentsim/events/handling/BeamEventsWriterXML.java
```diff
@@ -48,22 +54,96 @@ public void init(final String outfilename) {
...
             this.out.append("\t<event ");
             Map<String, String> attr = event.getAttributes();
+
             for (Map.Entry<String, String> entry : attr.entrySet()) {
-                this.out.append(entry.getKey());
-                this.out.append("=\"");
-                this.out.append(encodeAttributeValue(entry.getValue()));
-                this.out.append("\" ");
+
+                if(LoggerLevel.equals(LoggerLevels.VERBOSE.toString())){
```
colinsheppard: The code duplication here is not acceptable.

avinashchavanfidel:
@colinsheppard  
We can use the data structure like list which will have the list of attribute. So for each different level we would have different set of attribute . 
Based on list we can add the attributes .Also if needed we can define list in conf file so there would not be any hard coding in the code.
Please confirm.

tanajilondhefidel:
Also please state which fields should go under each level.
Currently I have added as per shared example.code wise I will make changes.

src/main/java/beam/agentsim/events/handling/BeamEventsWriterXML.java
avinashchavanfidel: yes , I will rearrange condition.

avinashchavanfidel:
@colinsheppard  I have made changes and come up with different approach as below 
1. Currently using the List data structure.
2. Each level has its own attribute define in beam.conf file as below , so we need to add attribute which we wanted to display in xml file
beam.outputs.VERBOSELoggingLevel = ["time","type","person","mode","vehical","departure_time","link"]
beam.outputs.REGULARLoggingLevel = ["time","type","person","mode","vehical","departure_time"]
beam.outputs.SHORTLoggingLevel = ["time","type","person"]
3. Any can any custom attribute which they want to display in xml file.
4. Minimize the code now.
Please review and let me know your feedback.

colinsheppard:
So we now have control over MATSim logging which is great, but we also want the same kind of control over BEAM logging. Let's make 2 config parameters that do these 2 things:

//valid options INFO,WARN,DEBUG,TRACE,ERROR
beam.levels.beamLoggingLevels = [""]  // for all logging internal to BEAM
beam.levels.dependencyLoggingLevels = [""]  // for logging by dependencies like MATSim & R5

avinashchavanfidel:
@colinsheppard  made changes as per comments. Currently SLF4J dont provide the console appender so we can specify only one option to control the log. It is working perfectly fine.Also since there is no console so we need to specify the package name or full classname which we want to control the logs .


Below is the logger in beam.conf 
##################################################################
# Logger Filtering
##################################################################
//Specify log level,If logger empty no logs from MATSIM and other classes.
// you can specify more than one logger level like INFO,WARN,DEBUG,TRACE,ERRO
beam.levels.dependencyLoggingLevels = [""] // for logging by dependencies like MATSim

//Specify one valid option among INFO,WARN,DEBUG,TRACE,ERROR
beam.levels.beamLoggingLevels = "" // for all logging internal to BEAM+R5

//Specify the package or full classname which you want to control
beam.packages=["beam","com.conveyal","scala.beam.sim.BeamSim","beam.agentsim.agents.RideHailingManager"]

src/main/java/beam/LogSLF4JController.java
```diff
@@ -0,0 +1,37 @@
+package beam;
```
colinsheppard: Please move this into package beam.sim.config

test/input/beamville/beam.conf
```diff
@@ -76,7 +76,12 @@ beam.outputs.explodeEventsIntoFiles = false
 # Logging Levels:
 # >0 enables loggin, <=0 suppresses logging.
 # Any event types not explicitly listed take on Default.level
-beam.outputs.defaultLoggingLevel = 1
+# if defaultLoggingLevel is blank then it will treat as OFF.
+beam.outputs.defaultLoggingLevel = "SHORT" # valid options:VERBOSE,REGULAR,SHORT,OFF
+beam.outputs.VERBOSELoggingLevel = ["time","type","person","mode","vehical","departure_time","link"]
```
colinsheppard: this is better than before, but still needs to be revised. the problem is that the set of fields is different for each type of event and we will have 10-20 event types to specify, so I think it would be too much clutter to do this in the config file. rather we might do this in a class somewhere or from data.

test/input/beamville/beam.conf
```diff
@@ -117,8 +122,15 @@ beam.events.filterDist = 10000.0
 # Logger Filtering
 ##################################################################
 //Specify log level,If logger empty no logs from MATSIM and other classes.
-// you can specify more than one logger level like INFO,WARN
-beam.levels.loggerLevels = ["DEBUG"]  //valid options INFO,WARN,DEBUG,TRACE,ERROR
+// you can specify more than one logger level like INFO,WARN,DEBUG,TRACE,ERRO
+beam.levels.dependencyLoggingLevels = [""] // for logging by dependencies like MATSim
+
+//Specify one valid option among INFO,WARN,DEBUG,TRACE,ERROR
+beam.levels.beamLoggingLevels = "" // for all logging internal to BEAM+R5
```
colinsheppard: I'd like R5 to be lumped in with the dependencyLoggingLevels.

test/input/beamville/beam.conf
```diff
@@ -117,8 +122,15 @@ beam.events.filterDist = 10000.0
...
+
+//Specify the package or full classname which you want to control
+beam.packages=["beam","com.conveyal","scala.beam.sim.BeamSim","beam.agentsim.agents.RideHailingManager"]
```
colinsheppard: So, which loggingLevel param applies to these classes, beamLoggingLevels or dependencyLoggingLevels or both?

src/main/resources/beam-template.conf
```diff
@@ -71,7 +71,11 @@ beam.outputs.explodeEventsIntoFiles = false
 # Logging Levels:
 # >0 enables loggin, <=0 suppresses logging.
 # Any event types not explicitly listed take on Default.level
-beam.outputs.defaultLoggingLevel = 1
+# if defaultLoggingLevel is blank then it will treat as OFF.
+beam.outputs.defaultLoggingLevel = "SHORT" # valid options:VERBOSE,REGULAR,SHORT,OFF
+beam.outputs.VERBOSELoggingLevel = ["time","type","person","mode","vehical","departure_time","link"]
```
colinsheppard: see my comments in beam.conf

src/main/java/beam/agentsim/events/handling/BeamEventsWriterXML.java
```diff
@@ -48,22 +52,44 @@ public void init(final String outfilename) {
...
+
+                else if(LoggerLevel.equals(LoggerLevels.SHORT.toString())) {
+                    loggerinfotype = this.beamServices.beamConfig().beam().outputs().SHORTLoggingLevel();
+                }
+
+                this.out.append("\t<event ");
+                Map<String, String> attr = event.getAttributes();
+                for (Map.Entry<String, String> entry : attr.entrySet()) {
+
+                    if(loggerinfotype.contains(entry.getKey())) {
```
colinsheppard: this doesn't cover the case when we wrap a MATSim event in our own, but it's a good start.

src/main/java/beam/LogSLF4JController.java
```diff
@@ -0,0 +1,37 @@
...
+public class LogSLF4JController {
+    public void cutOff (List<String> classList,String loggerLevels){
+        LoggerContext root = (LoggerContext) LoggerFactory.getILoggerFactory();
+        scala.collection.Iterator classIterator = classList.iterator();
+        while (classIterator.hasNext()) {
+            String className = classIterator.next().toString();
+            Logger rootLogger = root.getLogger(className);
+            rootLogger.setLevel(Level.OFF);
+
+            if("INFO".equals(loggerLevels)){
```
colinsheppard: just use: Level.valueOf(loggerLevel)

avinashchavanfidel:
@colinsheppard 
1)Changed Beam.config with suitable comments.
2)Moved MATSIMandR5LoggerController.scala and BEAMSIMLoggerController.scala to beam.sim.config package
3)MATSIM and R5 logs now getting handled in one file and beam in another.
4)I found 8 different events of org.matsim.core.events implemented in different files, what I am looking for how to wrap or how we can control those, can you suggest me.
Already implemented way fromr beam.conf, another from code but it would again hard code ?
or if we can limit like REGULAR would be 7 attribute VERBOSE would be ALL , SHORT would be 3-4 attribute .
Please comment on this so that I can make the change ,rest issue are fixed

colinsheppard:
Ok thanks. I'll come up with a suggestion on how to wrap sometime today.

colinsheppard:
@avinashchavanfidel I just committed some code that moves the logic around filtering in or out fields based on SHORT vs VERBOSE to BeamEventsLogger class, which is where we -- for now -- specify what fields are added for VERBOSE and dropped for SHORT.  

I didn't complete the job, please do so yourself and test for both XML can CSV output formats.

avinashchavanfidel:
@colinsheppard , 
I am working on as per suggested way.

avinashchavanfidel:
@colinsheppard, Please have a look at following attached file.These are the output files of my changes in code for XML and CSV Logger Events.
[logger_CSV.zip](https://github.com/LBNL-UCB-STI/beam/files/1309960/logger_CSV.zip)
[logger_XML.zip](https://github.com/LBNL-UCB-STI/beam/files/1309959/logger_XML.zip)
Please let me know your feedback.

src/main/java/beam/agentsim/events/handling/BeamEventsLogger.java
```diff
@@ -153,8 +171,10 @@ public void setEventsFileFormats() {
 
     public Map<String,String> getAttributes(Event event) {
         Map<String,String> attributes = event.getAttributes();
-        if(getLoggingLevel(event) == LoggerLevels.SHORT && eventFieldsToAddWhenVerbose.containsKey(event.getClass())){
-            // remove fields from attributes heres
+        if(getLoggingLevel(event) == LoggerLevels.SHORT && eventFieldsToDropWhenShort.containsKey(event.getClass())){
+            for (Map.Entry<Class, String> entry : eventFieldsToDropWhenShort.entrySet()) {
```
colinsheppard: This is incorrect and will drop all fields across all classes that are keys in eventFieldsToDropWhenShort. Note see above that eventFieldsToDropWhenShort should be a multimap.

src/main/java/beam/agentsim/events/handling/BeamEventsLogger.java
```diff
@@ -51,8 +51,26 @@ public BeamEventsLogger(BeamServices beamServices, EventsManager eventsManager)
         allLoggableEvents.add(PersonLeavesVehicleEvent.class);
         allLoggableEvents.add(PersonArrivalEvent.class);
         allLoggableEvents.add(ActivityStartEvent.class);
-
+        //Adding attribute which should be removed when the logger level SHORT
         eventFieldsToDropWhenShort.put(PathTraversalEvent.class,PathTraversalEvent.ATTRIBUTE_VIZ_DATA);
+        eventFieldsToDropWhenShort.put(PathTraversalEvent.class,PathTraversalEvent.ATTRIBUTE_LINK_IDS);
```
colinsheppard: eventFieldsToDropWhenShort and eventFieldsToAddWhenVerbose should be multimaps. If you are unfamiliar with this data type:

http://google.github.io/guava/releases/23.0/api/docs/com/google/common/collect/Multimap.html

avinashchavanfidel:
@colinsheppard  please check changes.

avinashchavanfidel:
Hi @colinsheppard ,
I have made changes as per suggestions and got reviewed code from Rashid.
Also please find below the attached files for your reference.
[CSV_logger.zip](https://github.com/LBNL-UCB-STI/beam/files/1315530/CSV_logger.zip)
[XML_logger.zip](https://github.com/LBNL-UCB-STI/beam/files/1315531/XML_logger.zip)

Please review and let me know feedback.

PR #95, opened 2017-09-13
 ResourceManager and Household Protocol Updates

sfwatergit:
Implement resource management w/ households and RHM. Also (mostly) implement HouseHold protocol.


colinsheppard:
Just to confirm, from looking over the new Household, we're assigning vehicles to persons during init and then that vehicle sticks with that person re: availability.... there's no ability for a lower ranked person to get access to an unused vehicle? 

this is fine for now, I just want to put this on the queue for future work.

colinsheppard:
Also @sfwatergit it looks like the person still isn't tracking their tour?

sfwatergit:
That's right on both counts. Initial commit was to resolve any inconsistencies with most current version of AgentSim, since it changes up the resource model for RHA a little bit.

PR #101, opened 2017-09-14
fix reflection matsim config builder 

dserdyuk:


src/main/scala/beam/sim/config/MatSimBeamConfigBuilder.scala
```diff
@@ -34,10 +34,24 @@ class MatSimBeamConfigBuilder(beamConf: Config) {
...
+    val maybeParameterSets = MatSimBeamConfigBuilder.concreteClassesOfType[MatsimParameters].
+      collect { case clazz if MatSimBeamConfigBuilder.isExtends(clazz, classOf[ConfigGroup]) =>
+        try {
+          Option(clazz.newInstance())
+        } catch {
+          case e: IllegalAccessException =>
+            logger.warn(s"Couldn't instantiate MatsimParameters  '${clazz.getCanonicalName}'. It doesn't have default public constructor.Falling back to setAccessible(). Cause : " + e.getMessage)
+            Try {
+              val c = clazz.getDeclaredConstructor()
+              c.setAccessible(true)
```
dserdyuk: Some of MatSim classes have protected constructor, thus we have to apply this workaround

colinsheppard:
Thanks @dserdyuk. So when we get this warning, is it safe to ignore? If so, let's mute the warning:

```
08:41:04.290 WARN (MatSimBeamConfigBuilder.scala:43) Couldn't instantiate MatsimParameters  'org.matsim.core.config.groups.PlanCalcScoreConfigGroup.ModeParams'. It doesn't have default public constructor.Falling back to setAccessible(). Cause : Class beam.sim.config.MatSimBeamConfigBuilder$$anonfun$1 can not access a member of class org.matsim.core.config.groups.PlanCalcScoreConfigGroup$ModeParams with modifiers ""
08:41:04.298 WARN (MatSimBeamConfigBuilder.scala:43) Couldn't instantiate MatsimParameters  'org.matsim.run.OldToNewPlanCalcScoreConfigGroup.ModeParams'. It doesn't have default public constructor.Falling back to setAccessible(). Cause : Class beam.sim.config.MatSimBeamConfigBuilder$$anonfun$1 can not access a member of class org.matsim.run.OldToNewPlanCalcScoreConfigGroup$ModeParams with modifiers ""
```

dserdyuk:
@colinsheppard pretty much yes, it may be worth to change it to info level. It mean basically there is no public constructor. And the only thing we can do use setAccessible() to construct config param intance that will be filled with props from typesafe conf

colinsheppard:
Ok, but the presence of the public constructor will probably never change, so it's not necessary to remind ourselves of this fact every time we run the model, it just clutters the console.

dserdyuk:
ok. I will change it to debug

colinsheppard:
thanks.

PR #103, opened 2017-09-16
Fix beamville physsim integration

wrashid:
beamville matsim network fixed
r5<->osmId conversion removed


src/main/java/beam/physsim/jdeqsim/AgentSimToPhysSimPlanConverter.java
```diff
@@ -148,28 +140,15 @@ public void handleEvent(Event event) {
                     leg.setTravelTime(0);
                     List<Id<Link>> linkIds = new ArrayList<>();
                     int negCount = 0;
-                    for(String link : links.split(",")) {
 
-                        long osmLinkId = NetworkCoordinator.getOsmId(Integer.parseInt(link));
+                    links=links.replace("Vector(","");
```
dserdyuk: @wrashid real problem is on agentsim side and it has been fixed in https://github.com/LBNL-UCB-STI/beam/pull/100

wrashid: @dserdyuk will the recently introduced Vector(x,...,z) representation in linkIds (part of PathTraversalEvent) change in the near future?

PR #111, opened 2017-09-23
added log4j2 configuration, autoreload set to 30 secs

dserdyuk:


src/main/scala/beam/sim/RunBeam.scala
```diff
@@ -64,7 +64,7 @@ trait RunBeam {
 
     //TODO this line can be safely deleted, just for exploring structure of config class
     //  ConfigModule.beamConfig.beam.outputs.outputDirectory;
-    BeamLoggingSetup.configureLogs(ConfigModule.beamConfig)
+    //BeamLoggingSetup.configureLogs(ConfigModule.beamConfig)
```
colinsheppard: So this is where I was attempting to configure anything that goes to console to also go to a file. This needs to happen at runtime because the location of the file is in the dynamically created outputs directory.

dserdyuk:
@colinsheppard  It looks like outdated. Do we still  need this ?

Issue #121, opened 2017-10-06
Refactor from "output logging" to "events writing"


colinsheppard:
In the config file the naming is confusing and doesn't make it clear that those params change behavior of what events are written to file.

PR #146, opened 2017-11-09
Refactoring config params Issue #121

colinsheppard:
@michaz based on your comment a couple weeks ago, I've renamed the event writing config params, let me know if this makes better sense on first read...


michaz:
Hmm yes. I'm not quite sure how the integration with MATSim works / is planned to work -- is it only about the level of Events that are _written_, or also about the level of Events that are _produced_ / published on the event bus? The former, right? Then I think the new name is good.

colinsheppard:
@fdariasm this one is quick and simple, but I wanted you to be aware since you've developing against the config file. Let me know.

Issue #112, opened 2017-10-04
Protocols Documentation


colinsheppard:
Revise the protocol descriptions based on recent code revisions.

Create visualizations of each for reference.

PR #147, opened 2017-11-09
Updating docs #112

colinsheppard:
Issue #112


colinsheppard:
@sfwatergit can you check out the updated protocols which now reflect removing BeamVehicle from the loop and let me know if anything is inconsistent with your refactoring?

PR #148, opened 2017-11-09
Vehicle Demotion

colinsheppard:
Changing BeamVehicles from BeamAgents to a case class.


sfwatergit:
There are some additional commits that I haven't pushed yet.

Issue #149, opened 2017-11-09
When Using PErsonalVeh Don't Request Ride Hail


colinsheppard:
BUG!

PR #150, opened 2017-11-10
No ride hail if using personal vehicle

colinsheppard:
Issue #149


colinsheppard:
This branch was first merged into pull #164 and will get reviewed there.

PR #151, opened 2017-11-11
Replace blocking 'sleep' with non-blocking scheduling an Akka event.

michaz:
Now the single-threaded test works without hanging, and can replace the regular one.


colinsheppard:
Looks great!

PR #152, opened 2017-11-11
Simplify router

michaz:
- Removed many of the global variables in BeamServices
- Removed much indirectness (is that a word?) in the way the router keeps its transit legs which it returns in its answers. Note: The router unfortunately has to keep all the street-network-routed transit legs in a map, for the _only_ reason that passenger agents must have the exact same EmbodiedBeamLeg value as the transit driver, to reserve and travel it. That is because it is used as a key in many places. If that were not the case, the router could simply reply with a trip id and a (fromStop, toStop) pair of indices into the trip, saving probably hundreds of megabytes, if not several gigabytes, of memory. (The transit drivers would then be able to create their routed itinerary only when they start, and consume it as they drive.)
- We only need one router actor. Since routing is stateless, it can return immediately, starting a Future which routes and sends the answer. (We **could** run the router in its own dispatcher, limiting the number of threads it can use, as before. But only if we see a reason for that.)
- Removed lots of initialization protocol.
- Removed the need for giving routing requests an id. Akka can keep track of which response belongs to which request by itself, with the "?"-pattern. (It's safe to use that within Actors, only ``Await``ing them is bad, but that is not necessary because Futures can be combined, without losing the order.) Removed the custom Aggregator -- I'm almost sure it was only used in this case to re-implement "?".
- Remove a global map of Ids to ActorRefs for transit drivers. Ideally, of course, ActorRefs are passed around, and everyone only knows the people they need to know. But keeping them in a global map is superfluous -- An ActorRef _is_ already only a name or id or URL, it can easily be created on the fly from a String when it is needed, without going through an Id and a Map. The same can be done for the other global ActorRef maps as well.

What do you think?


colinsheppard:
Hi @michaz I support these improvements but would like to discuss a couple of them before we move ahead with the merge. 

If Id's are directly translatable to actor-refs and we can make lightweight converters (or better yet implicit conversions) from and Id to the appropriate actor-ref, then that would be beautiful and excise a lot of needless clutter from the codebase. To me, the following feels a bit verbose despite the fact that it does solve the need for global maps:

`TransitDriverAgent.lookupActorFromVehicleId(context, inferredVehicle.outermostVehicle())`

Can you look into something a bit more concise?

Second, about using the transit legs as keys... I have some concerns about the simplified protocol you suggest:

* This may require treating transit differently from the perspective of PersonAgents as they execute a trip. Right now in order to consume a trip, a PersonAgent is assumed to have all atomic BeamLegs in their "currentRoute" and then check the BeamLeg sent to them by NotifyLegStartTrigger against their route to confirm they are executing each step in proper order. If the PersonAgents don't know about the individual legs of their transit rides, they will not be able to do this check and we may therefore be treating transit differently than riding as a passenger in other modes.
* We would have to combine the trip Id and stops with other data like travel time, cost, and mode so the agents can properly evaluate and choose between the transit itinerary and other alternatives.
* We wouldn't have a convenient mechanism for within-trip replanning if we ever want to implement such a feature. I'm not overly concerned with this at the moment, since we're in the same predicament with our CAR trips anyway (i.e. we will have to break those up into smaller segments to enable replanning). But since we are  leaning toward enabling within-trip replanning, we might want to be careful about reorganizing transit in this way.

michaz:
Thanks!

> Can you look into something a bit more concise?

Yes, sounds good and doable.

> Second, about using the transit legs as keys... I have some concerns about the simplified protocol you suggest:

I think you're right. (This PR leaves this as it is. I think I just wanted to point out the "price tag".)

colinsheppard:
Based on our discussion yesterday, let's merge this after you have refactored the Id to ActorRef pattern.

PR #153, opened 2017-11-15
Logging

michaz:
- Fix upstream MATSim so it doesn't use its bronze-age log4j1 in an interfering way.
- Replace (from the outside) MATSim's log4j1 by an slf4j proxy.
- Attach slf4j to Akka by configuration, so Akka senses what log level the logger *implementation* wants.
- Re-introduce the log4j2 (that is what we are actually using) configuration files, whose settings Akka now knows about. Contrary what I said earlier, I don't think we can fully configure logging programmatically. Reason: The configuration should hold for tests, too, and we can't / shouldn't enforce every unit test explicitly loading some logging config. There are now two easily-findable logging configurations, one for production, and one for tests. On their respective class-paths, i.e. in the resource folders. If we want to customize logging for a run, the place to look for how that works is now the docs of the logger implementation, i.e. log4j2, and it should work as advertised.
- Remove all traces of logger configuration from the code.
- In case we want to, we should now in fact be able to drop-in-replace the logger implementation! Just replace log4j2 by, say, logback in build.gradle, and give it the configuration files it wants, and delete the log4j2 config files.


michaz:
Finished.

michaz:
Not finished. Got to provoke and fix this presumed event race condition thing.

michaz:
Finished. Integrated MATSim and Akka Events handling the other way round: We implements ``EventsManager``, and our implementation (hopefully) waits on the Akka event bus on ``finishProcessing``. It does so by posting an event indicating the end of the iteration _to the same Akka event bus on which the MATSim events go_, in the hope that there is _some_ kind of FIFO guarantee even across Actors in this case. It then waits for that same event to come back via the event bus.

Ugh.

colinsheppard:
@michaz so apologies for not getting to this before it was merged. Can you confirm a couple of features that we want?

* The ability to independently control the logging level between BEAM logging (i.e. akka) and dependencies (MATSim and R5 in particular). If this is easily configurable through slf4j, can you post a couple examples in our docs (i.e. docs/users.rst) so that each individual contributor doesn't have to conduct their own research on how to control the logger.
* The ability to save console output into the outputDirectory. It is a hassle when running the model on EC2 to manually manage the location of the console log and remember to archive it along with the run outputs. This was a key motivation to programmatically controlling the logging because we don't know where that file will be before the run. So how will we accomplish this now?
* Finally, so are the directives currently in the BEAM config file (beam.outputs.logging.beam.logLevel, beam.outputs.logging.dependencies.logLevel) now deprecated? If so, we should note this and/or remove them from the config...

michaz:
Thanks for checking. 

a) is easy (will document) 
b) okay, will get that back quickly
c) yes

Issue #105, opened 2017-09-19
Experiment Manager


colinsheppard:
We need a tool that makes it easy to specify multi-factorial experiments and deploy them to the cloud. We can re-use a utility I developed in the past as a starting place:

https://github.com/colinsheppard/abmutils/blob/master/src/main/java/abmutils/experiment/

This works by creating an experiment file (in yaml format). Here's an example:

```
Author: Andy
Title: Verify-Nest-Initiation
InputsDirectory: modeling-notebook/model-inputs/
OutputsDirectory: ternpop_verification/
Replicates: 2
NonDefaults:
- Logfiles: Tern,Site,Event,Nest
- NumThreads: 1
Factors:
- Code: MaxBreedingAttempts
  Levels:
  - Code: Normal
    Params:
    - MaxNumBreedingAttempts: 2
  - Code: Extended
    Params:
    - MaxNumBreedingAttempts: 3
- Code: MortalityProbability
  Levels:
  - Code: Low
    Params:
    - MortalityProbability: 0.1
  - Code: High
    Params:
    - MortalityProbability: 0.2
```

The key component to this is the "Factors". In this example there are 2 factors, MaxBreedingAttempts and MortalityProbability, each with 2 levels. The experiment manager's job is to configure and run the model for every combination of factors (in this case 4 times). In addition there is a Replicates field that means the model will be re-run multiple times with different random seeds as the starting place. While actually using Replicates may be overkill for BEAM, we should keep that functionality. 

The NonDefaults field is a way to quickly override a config parameter's default value for the entirety of an experiment (without needing to modify the base conf file).

**How BEAM Experiment Manager will be different**

The experiment manager I created was designed to run serially within one session. Our manager needs to enable cloud deployment. More to come on this...

dserdyuk:
@colinsheppard  I assume Params for BEAM are going to be the one from beam.conf ?

colinsheppard:
Yeah that's the idea. so in the yaml we could basically specify the fully qualified path to the param in beam.conf, e.g. "beam.agentsim.numAgents" as the param name which we then access via reflection. Also, feel free to propose a re-organization of the yaml file if you think it can be improved. Also keep in mind that we will definitely want the ability to do parameter sweeps where we specify from, to, by, or monte carlo where specify a distribution. But most of the time, discrete levels will work.

PR #158, opened 2017-11-21
 #105 Experiment Manager

dserdyuk:
* added template generator + template for beamville


colinsheppard:
Thanks @dserdyuk. Can you please add some documentation under docs/users.rst describing how to use the manager. Feel free to do so in short bullet point form, but please add examples of how to configure the tpl files, how to run the generator, and then how to run the experiment locally. We can add how to deploy on the cloud once that capability is ready.

dserdyuk:
@colinsheppard Done

zishanbilal:
@colinsheppard i have integrated cloud deployment part and also updated the user guide as well.

src/main/scala/beam/experiment/ExperimentRunSandbox.scala
```diff
@@ -13,7 +13,6 @@ case class ExperimentRunSandbox(projectRoot: Path, experimentBaseDir: Path, expe
   lazy val runConfig: Config = buildRunConfig
 
   def runDirectory = Paths.get(experimentBaseDir.toString,
-    experimentDef.title.replace("\\s", "_"),
```
dserdyuk: I'd better leave replace in case user want to put whitespaces in title, otherwise it may cause file not found error while resolving runExperiment.sh, etc.

test/input/sf-light/calibration/transport-cost/experiments.yml
```diff
@@ -3,8 +3,6 @@
 title: Calibration-SFLight-v6
 author: Colin
 beamTemplateConfPath: test/input/sf-light/sf-light-100k.conf
-runExperimentScript: test/input/sf-light/calibration/runExperiment.sh.tpl
```
dserdyuk: I thought runExperimentScript and batchRunScript are mandatory  params ?

Issue #160, opened 2017-11-22
Transit not used


colinsheppard:
When production of sf-light run, no one choose transit.

This is due to some recent changes in mode choice which now distinguish between WALK-TRANSIT and DRIVE-TRANSIT.

colinsheppard:
Getting agents to choose these modes was easy... but now there are a couple of transit related bugs causing model to stall (skipOverBadActors is FALSE). So this issue has now morphed into debugging those.

colinsheppard:
The transit bug was due to the GTFS data having unrealistic stop times... specifically, subsequent stops on a trip would have the exact same timestamp... when loaded and run in BEAM, this caused agents to get "confused" about StartLeg trigger should happen first... for transit trips with a large number of stops that all have the same timestamp, the agent get's stuck.

My solution to this was to amend the GTFS data and repair these stop sequences by replacing the stop times with ones that interpolated between successive stops for which there is a change in time. E.g. if previously the stops looks like A:0, B:0, C:0, D:15.... then the repaired sequence would become A:0, B:5, C:10, D:15.

colinsheppard:
See PR #164 for the new script, under src/main/R/bin/repairStopTimes.R

michaz:
I'm working on branch ``islandpruning``, where I removed your time window fix for the end-of-day problem again. It prevents many (all?) transit trips, since for transit trips, the actual departure time is _normally_ outside the time window.

(The time window specifies the _earliest_ departure time. R5 responds with the actual departure time.)

michaz:
((We have a transit router test now, can't happen again.))

colinsheppard:
Yeah, I saw that... problem is that yesterday I definitely found a case where a trip was 24 hours after the time window. So we need a filter to prevent that. My working solution right now is to filter out anything more than 30 minutes after the start of the window.

michaz:
Okay.. and why does a trip 24 hours after the time window lead to errors on our side? Are we making yet another day-wrapping error on top of that, or is it something else?

Otherwise I'd think the agent would just go on a very late trip..

colinsheppard:
Yeah, I think we're making a day wrapping error on top of the issue... the start times for the walk legs end up a day ahead of the start times for the transit legs.

PR #164, opened 2017-11-22
Cjrs/issue160 notransit

colinsheppard:
Issues #160 #150 are addressed here.


colinsheppard:
@michaz can you please review and merge at your convenience?

Thanks!

colinsheppard:
@michaz 

I mistakenly pushed some partial updates to master... I think once you review and merge this PR all will be well again:

https://travis-ci.org/LBNL-UCB-STI/beam/builds/311377071

michaz:
Hmm, were the visualization commits developed on this branch, or did they get in here by accident? I'm just asking as a precaution, I don't want to produce merge conflicts later... Also the data file changes.

michaz:
But since this is the de-facto master at the moment anyway, I guess I'll just merge. :-)

colinsheppard:
That stuff made it's way in via application-sfbay which I've merged with my feature branch in order to test my bug fixes at scale. I'm not worried about them though because for now all viz stuff happens in post-processing and shouldn't impact the model itself.

michaz:
What concerns me a bit is that this PR introduces a fix which patches data and messages with dummy values, but when I remove those changes (a2ad8c5), it still works.

Those lines look very much like we should have a test which shows what they are fixing, otherwise we won't understand after two weeks what we did there.

Since we have sfbay-light on master now, would it be possible to produce a test which fails now, in this state?

michaz:
That would probably add "people use transit" to the sfbay-light spec.

I recently commented out the one spec on the result of that run, because Travis would choke on the event file. (Which is reasonable. We can't expect to load all events into _memory_, they are meant for stream processing.)

I'll try to put an example for testing event streams here.

Issue #165, opened 2017-11-30
Refactor Driving/Riding Protocols


colinsheppard:
We will refactor driving in order to accomplish the following:

* Make driving more dynamic, particularly for transit drivers, where passenger manifest is unknown until the time to drive has arrived
* Decompose management of the manifest for each Leg out of DrivesVehicle and into a DriverManager, that is responsible for managing PassengerSchedules and dispatching the driver
* Management of becoming and unbecoming a driver will happen separately by a DriverDeterminer, who is responsible for telling the Driver their vehicle and their DriverManager

PR #166, opened 2017-11-30
Refactor driving/riding Protocols

colinsheppard:
See issue #165.


michaz:
Why would an agent get a CompletionNotice (and have to forward it)?

michaz:
Ah, because when an agent thinks it's done with something which was originally prompted by a trigger, it still "replies" with the completion notice to whomever sent the last message which was required for that, which may not be the scheduler. Hmpf.

colinsheppard:
Yes... adding this redirect avoids wasting time tracking down bugs caused simply by forgetting to send the complete to the scheduler versus "sender"

michaz:
Hmhmm, I see.. still don't think it would be wasted time to track this down. I mean, sending messages to the receipient required by the protocol is something we should do without "soft error handling".

Issue #122, opened 2017-10-06
Save data to outputs folder


colinsheppard:
In order of priority:

- config file
- console output (all of it)
- small files used as input (e.g. ModeChoiceParams)
- git commit hash + path to config file used from project root (to be used later to recreate the state of all inputs)

colinsheppard:
To elaborate, we want to copy the config file and other small files from the inputdirectory to the outputDirectory at the end of the run.

We also want to make sure all console output is getting logged to the console.out file in the outputDirectory.

Finally, we want to save the commit hash and path to config file for reproducibility.

PR #191, opened 2018-01-04
Integration enrich output #122

zishanbilal:
Save data to outputs folder #122 
- config file
- FileLogger for console output
- small files like ModeChoiceParams
- git commit hash + path to config file


colinsheppard:
@zishanbilal so when the model runs, I do get console logging going to a file in the output directory. But the output no longer appears on the console. Can we have it go to both places?

PR #216, opened 2018-01-30
Fdariasm/integration revert

fdariasm:
Reverting missing changes for EventsFileSpec, un-ignoring two passing tests. Please check and confirm to merge back to master.


michaz:
Thanks -- actually, I was happy to get rid of the ``getListOfSubDirectories`` thing in ``EventsFileHandlingCommon``, but now it's coming back.. The real output directory is contained in  ``matsimConfig.controler().getOutputDirectory``, so it doesn't need to be guessed but can be taken from there, like in ``getEventsFilePath``, just above ``getRouteFile``. Can it be done the same way in your case?

fdariasm:
Yes for sure, I missed that change.

fdariasm:
@michaz done, sorry for the delay, I was having lunch

michaz:
Okay, thanks!

PR #219, opened 2018-02-06
Refactor Logit into Case Calss

colinsheppard:
@michaz I've changed the MNL class into a case class and moved the mode choice parameters into the config file instead of an external XML. 

I still want to do a comparable refactor of the LCCM, but would like your input before proceeding.

Previously, we were cloning the mode choice object for each agent because it used mutable state to do the internal calculations and sampling, etc. But now (for MNL at least) this is no longer the case. I haven't changed that behavior yet.


michaz:
Oh, thank you! Have to review yet, but that sounds good.

sfwatergit:
Very important! I was just hacking into this myself by adding an extractor (unapply) method so that pattern matching would be possible. It was confusing to have that rogue apply method there.

michaz:
Now if we also remove the ``var expectedMaximumUtility`` in ``ModeChoiceMultinomialLogit``, it is a pure function. We can then move the switch-block in ``object ModeChoiceCalculator`` to the Guice module, dependency-inject the thing, and done. No factories, no ``HasServices``, no nothing.

michaz:
(I'll gladly do that last step.)

PR #281, opened 2018-05-22
Refactor - RideHailingManager

carloscaldas:
* General
  - Removed unused fields: log, RIDE_HAIL_MANAGER
  - Removed commented code: just noise. people never read it. Changes should be introduced with tests
  - Unnecessary mutable: rideHailResourceAllocationManager, passengerScheduleOverrideRequests
  - Removed warnings: return; shadowing names, type annotation for public and others
  - Removed variables with values never queried: `todo_allocationsDuringReservation`
  - extract variables

* Reliability
  - Encapsulation issues
    public var <-- very dangerous
    public val <-- not big problem
  - Removed possibility of `NoSuchElementException`: `resources.get(agentsim.vehicleId2BeamVehicleId(vehId)).get`

* performance
  - test iterator emptyness checking size -> isEmpty(): 
    some data structures need O(N) to identify the size and O(1) to check if is empty
  - Moved constant to Companion object: save memory

* readability
  - Avoid negating codition (`timesToCustomer.nonEmpty`):
    eg. if (!A) B else C -> if (A) C else B
  - Formatting: removing FQN-Full Qualified Name from code
  - identation and small things
  - Fixed typo
  - Improved variable names -> more meaningful
    Good practice: always use the unit of time, distance, etc. 
    e.g.: radius -> radiusInMeters; fare -> farePerHour
  - breaked long lines
  
Suggested nex steps:
- Generate tests
- Extract methods
- Extract class
- Achieve Single Responsibility principle


carloscaldas:
should be merged with the branch `tnc-api-dev` instead of `master`

PR #288, opened 2018-06-02
Immutable collection with var are replaced by appropriate mutable

REASY:
- Immutable collection with var are replaced by appropriate mutable
- Added type for public methods and fields

### DIFF
--- a/src/main/java/beam/agentsim/agents/choice/logit/LatentClassChoiceModel.scala
+++ b/src/main/java/beam/agentsim/agents/choice/logit/LatentClassChoiceModel.scala
@@ -17,7 +17,7 @@ import scala.collection.mutable
   * BEAM
   */
 class LatentClassChoiceModel(override val beamServices: BeamServices) extends HasServices {
-  val lccmData: IndexedSeq[LccmData] = parseModeChoiceParams(beamServices.beamConfig.beam.agentsim.agents.modalBehaviors.lccm.paramFile)
+  private val lccmData: IndexedSeq[LccmData] = parseModeChoiceParams(beamServices.beamConfig.beam.agentsim.agents.modalBehaviors.lccm.paramFile)
 
   val classMembershipModels: Map[TourType, MultinomialLogit] = extractClassMembershipModels(lccmData)
   val modeChoiceModels: Map[TourType, Map[String, MultinomialLogit]] = {
--- a/src/main/java/beam/agentsim/agents/choice/logit/LatentClassChoiceModel.scala
+++ b/src/main/java/beam/agentsim/agents/choice/logit/LatentClassChoiceModel.scala
@@ -11,12 +11,13 @@ import org.supercsv.io.CsvBeanReader
 import org.supercsv.prefs.CsvPreference
 
 import scala.beans.BeanProperty
+import scala.collection.mutable
 
 /**
   * BEAM
   */
 class LatentClassChoiceModel(override val beamServices: BeamServices) extends HasServices {
-  val lccmData: Vector[LccmData] = parseModeChoiceParams(beamServices.beamConfig.beam.agentsim.agents.modalBehaviors.lccm.paramFile)
+  val lccmData: IndexedSeq[LccmData] = parseModeChoiceParams(beamServices.beamConfig.beam.agentsim.agents.modalBehaviors.lccm.paramFile)
 
   val classMembershipModels: Map[TourType, MultinomialLogit] = extractClassMembershipModels(lccmData)
   val modeChoiceModels: Map[TourType, Map[String, MultinomialLogit]] = {
@@ -24,21 +25,21 @@ class LatentClassChoiceModel(override val beamServices: BeamServices) extends Ha
     mods
   }
 
-  def parseModeChoiceParams(lccmParamsFileFile: String): Vector[LccmData] = {
+  def parseModeChoiceParams(lccmParamsFileFile: String): IndexedSeq[LccmData] = {
     val beanReader = new CsvBeanReader(IOUtils.getBufferedReader(lccmParamsFileFile), CsvPreference.STANDARD_PREFERENCE)
     val header = beanReader.getHeader(true)
     val processors: Array[CellProcessor] = LatentClassChoiceModel.getProcessors
 
     var row: LccmData = new LccmData()
-    var data: Vector[LccmData] = Vector()
+    val data = mutable.ArrayBuffer[LccmData]()
     while (beanReader.read[LccmData](row, header, processors: _*) != null) {
-      if (Option(row.value).isDefined && !row.value.isNaN) data = data :+ row.clone().asInstanceOf[LccmData]
+      if (Option(row.value).isDefined && !row.value.isNaN) data += row.clone().asInstanceOf[LccmData]
       row = new LccmData() // Turns out that if beanReader encounters a missing field, it just doesn't do the setField() rather than set it to null.... so we need to mannualy reset after every read
     }
     data
   }
 
-  def extractClassMembershipModels(lccmData: Vector[LccmData]): Map[TourType, MultinomialLogit] = {
+  def extractClassMembershipModels(lccmData: IndexedSeq[LccmData]): Map[TourType, MultinomialLogit] = {
     val classMemData = lccmData.filter(_.model == "classMembership")
     Vector[TourType](Mandatory, Nonmandatory).map { theTourType =>
       val theData = classMemData.filter(_.tourType.equalsIgnoreCase(theTourType.toString))
@@ -55,7 +56,7 @@ class LatentClassChoiceModel(override val beamServices: BeamServices) extends Ha
    * We use presence of ASC to indicate whether an alternative should be added to the MNL model. So even if an alternative is a base alterantive,
    * it should be given an ASC with value of 0.0 in order to be added to the choice set.
    */
-  def extractModeChoiceModels(lccmData: Vector[LccmData]): Map[TourType, Map[String, MultinomialLogit]] = {
+  def extractModeChoiceModels(lccmData: IndexedSeq[LccmData]): Map[TourType, Map[String, MultinomialLogit]] = {
     val uniqueClasses = lccmData.map(_.latentClass).distinct
     val modeChoiceData = lccmData.filter(_.model == "modeChoice")
     Vector[TourType](Mandatory, Nonmandatory).map { theTourType =>
--- a/src/main/scala/beam/agentsim/agents/Population.scala
+++ b/src/main/scala/beam/agentsim/agents/Population.scala
@@ -20,7 +20,7 @@ import org.matsim.core.api.experimental.events.EventsManager
 import org.matsim.households.Household
 import org.matsim.vehicles.Vehicles
 
-import scala.collection.JavaConverters
+import scala.collection.{JavaConverters, mutable}
 import scala.collection.JavaConverters._
 import scala.concurrent.{Await, Future}
 
@@ -33,21 +33,20 @@ class Population(val scenario: Scenario, val beamServices: BeamServices, val sch
     case _: Exception => Stop
     case _: AssertionError => Stop
   }
-  private implicit val timeout = Timeout(50000, TimeUnit.SECONDS)
+  private implicit val timeout: Timeout = Timeout(50000, TimeUnit.SECONDS)
 
   import context.dispatcher
 
-  private var personToHouseholdId: Map[Id[Person], Id[Household]] = Map()
+  private val personToHouseholdId: mutable.Map[Id[Person], Id[Household]] = mutable.Map[Id[Person], Id[Household]]()
   scenario.getHouseholds.getHouseholds.forEach { (householdId, matSimHousehold) =>
-    personToHouseholdId = personToHouseholdId ++ matSimHousehold.getMemberIds.asScala.map(personId => personId -> householdId)
+    personToHouseholdId ++= matSimHousehold.getMemberIds.asScala.map(personId => personId -> householdId)
   }
-
 
   // Init households before RHA.... RHA vehicles will initially be managed by households
   initHouseholds()
 
 
-  override def receive = {
+  override def receive: PartialFunction[Any, Unit] = {
     case Terminated(_) =>
     // Do nothing
     case Finish =>
--- a/src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceDriveIfAvailable.scala
+++ b/src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceDriveIfAvailable.scala
@@ -14,11 +14,8 @@ import scala.collection.mutable.ArrayBuffer
 class ModeChoiceDriveIfAvailable(val beamServices: BeamServices) extends ModeChoiceCalculator {
 
   def apply(alternatives: Seq[EmbodiedBeamTrip]): Option[EmbodiedBeamTrip] = {
-    val containsDriveAlt: ArrayBuffer[Int] = ArrayBuffer[Int]()
-    alternatives.zipWithIndex.foreach { alt =>
-      if (alt._1.tripClassifier == CAR) {
-        containsDriveAlt += alt._2
-      }
+    val containsDriveAlt = alternatives.zipWithIndex.collect {
+      case(trip, idx) if trip.tripClassifier == CAR  => idx
     }
     if (containsDriveAlt.nonEmpty) {
       Some(alternatives(containsDriveAlt.head))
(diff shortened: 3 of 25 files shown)


src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceDriveIfAvailable.scala
```diff
@@ -6,16 +6,18 @@ import beam.router.Modes.BeamMode.CAR
...
+
 /**
   * BEAM
   */
 class ModeChoiceDriveIfAvailable(val beamServices: BeamServices) extends ModeChoiceCalculator {
 
   def apply(alternatives: Seq[EmbodiedBeamTrip]): Option[EmbodiedBeamTrip] = {
-    var containsDriveAlt: Vector[Int] = Vector[Int]()
+    val containsDriveAlt: ArrayBuffer[Int] = ArrayBuffer[Int]()
     alternatives.zipWithIndex.foreach { alt =>
```
REASY: Fix this with:
```
alternatives.zipWithIndex.filter { case (beamTrip, idx) => beamTrip. tripClassifier == CAR }.map { case (beamTrip, idx) => idx}

```

carloscaldas:
Since you are replacing immutable with mutable I recommend:
- Tight the visibility to private 
- Pay special attention to not share the references to other classes

JustinPihony:
I agree with @carloscaldas As a rule of thumb you can use mutable internally, but expose immutable - otherwise it starts to get messy

src/main/java/beam/agentsim/agents/choice/logit/LatentClassChoiceModel.scala
```diff
@@ -11,34 +11,35 @@ import org.supercsv.io.CsvBeanReader
...
 
     var row: LccmData = new LccmData()
-    var data: Vector[LccmData] = Vector()
+    val data = mutable.ArrayBuffer[LccmData]()
     while (beanReader.read[LccmData](row, header, processors: _*) != null) {
-      if (Option(row.value).isDefined && !row.value.isNaN) data = data :+ row.clone().asInstanceOf[LccmData]
+      if (Option(row.value).isDefined && !row.value.isNaN) data += row.clone().asInstanceOf[LccmData]
       row = new LccmData() // Turns out that if beanReader encounters a missing field, it just doesn't do the setField() rather than set it to null.... so we need to mannualy reset after every read
     }
     data
```
JustinPihony: data is still an ArrayBuffer under the covers, and technically mutable. I'd propose `data.toIndexedSeq`

src/main/scala/beam/agentsim/agents/Population.scala
```diff
@@ -33,21 +33,20 @@ class Population(val scenario: Scenario, val beamServices: BeamServices, val sch
     case _: Exception => Stop
     case _: AssertionError => Stop
   }
-  private implicit val timeout = Timeout(50000, TimeUnit.SECONDS)
+  private implicit val timeout: Timeout = Timeout(50000, TimeUnit.SECONDS)
 
   import context.dispatcher
 
-  private var personToHouseholdId: Map[Id[Person], Id[Household]] = Map()
+  private val personToHouseholdId: mutable.Map[Id[Person], Id[Household]] = mutable.Map[Id[Person], Id[Household]]()
```
JustinPihony: As this is only done at creation then it seems the only reason mutability was used was because of the initialization. So, I'd argue for something more like this:

```
  private val personToHouseholdId: scala.collection.Map[Id[Person], Id[Household]] = 
    scenario.getHouseholds.getHouseholds.asScala
      .flatMap{ case (householdId, matSimHousehold) =>
        matSimHousehold.getMemberIds.asScala.map(_ -> householdId)
      }
```

Of course if you feel the performance of mutation to be better, then you could do something more like this and end with the same result:

```
  def initPersonToHouseholdId(houseHolds: java.util.Map[Id[Household], Household]): scala.collection.Map[Id[Person], Id[Household]] = {
    val personToHouseholdId: mutable.Map[Id[Person], Id[Household]] = mutable.Map[Id[Person], Id[Household]]()
    scenario.getHouseholds.getHouseholds.forEach { (householdId, matSimHousehold) =>
      personToHouseholdId ++= matSimHousehold.getMemberIds.asScala.map(personId => personId -> householdId)
    }
    personToHouseholdId.toMap
  }
  
  private val personToHouseholdId: scala.collection.Map[Id[Person], Id[Household]] = 
    initPersonToHouseholdId(scenario.getHouseholds.getHouseholds)
```

src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceDriveIfAvailable.scala
JustinPihony: Even better would be:

`val containsDriveAlt = alternatives.zipWithIndex.collect{ case(trip, idx) if(trip.tripClassifier == CAR) => idx}`

src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceLCCM.scala
```diff
@@ -136,7 +136,7 @@ class ModeChoiceLCCM(val beamServices: BeamServices, val lccm: LatentClassChoice
       ModeChoiceData(altAndIdx._1.tripClassifier, tourType, vehicleTime, walkTime, waitTime, bikeTime, totalCost.toDouble, altAndIdx._2)
     }
 
-    val groupedByMode = modeChoiceAlternatives.sortBy(_.mode.value).groupBy(_.mode)
+    val groupedByMode: Map[BeamMode, Seq[ModeChoiceData]] = modeChoiceAlternatives.groupBy(_.mode)
```
JustinPihony: I believe that the sort will indeed be lost in the long term, but was this meant to be an attempt to sort the values within the grouping? I don't see any problem with the optimization, and am just curious if the original intent was lost and should be fixed instead of optimized?

src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceRideHailIfAvailable.scala
```diff
@@ -6,16 +6,18 @@ import beam.router.Modes.BeamMode.RIDE_HAIL
...
+import scala.collection.mutable.ArrayBuffer
+
 /**
   * BEAM
   */
 class ModeChoiceRideHailIfAvailable(val beamServices: BeamServices) extends ModeChoiceCalculator {
 
   override def apply(alternatives: Seq[EmbodiedBeamTrip]): Option[EmbodiedBeamTrip] = {
-    var containsRideHailAlt: Vector[Int] = Vector[Int]()
+    val containsRideHailAlt: ArrayBuffer[Int] = ArrayBuffer[Int]()
```
JustinPihony: Same as above:

`val containsRideHailAlt = alternatives.zipWithIndex.collect{ case(trip, idx) if(trip.tripClassifier == RIDE_HAIL) => idx}`

src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceTransitIfAvailable.scala
```diff
@@ -5,18 +5,20 @@ import beam.router.Modes
...
 /**
   * BEAM
   */
 class ModeChoiceTransitIfAvailable(val beamServices: BeamServices) extends ModeChoiceCalculator {
 
   override def clone(): ModeChoiceCalculator = new ModeChoiceTransitIfAvailable(beamServices)
 
-  override def apply(alternatives: Seq[EmbodiedBeamTrip]) = {
-    var containsTransitAlt: Vector[Int] = Vector[Int]()
+  override def apply(alternatives: Seq[EmbodiedBeamTrip]): Option[EmbodiedBeamTrip] = {
```
JustinPihony: Same as other alt comments. I'll stop here and this should just be repeated for any future instances like this.

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -140,32 +140,32 @@ object HouseholdActor {
...
       * Available [[Vehicle]]s in [[Household]].
       */
-    var _vehicles: Vector[Id[Vehicle]] = vehicles.keys.toVector.map(x => Id.createVehicleId(x))
+    val _vehicles: Vector[Id[Vehicle]] = vehicles.keys.toVector.map(x => Id.createVehicleId(x))
 
     /**
       * Concurrent [[MobilityStatusInquiry]]s that must receive responses before completing vehicle assignment.
       */
-    var _pendingInquiries: Map[Id[MobilityStatusInquiry], Id[Vehicle]] = Map[Id[MobilityStatusInquiry], Id[Vehicle]]()
+    val _pendingInquiries: Map[Id[MobilityStatusInquiry], Id[Vehicle]] = Map[Id[MobilityStatusInquiry], Id[Vehicle]]()
```
JustinPihony: This one can be dropped. It isn't used.

src/main/scala/beam/agentsim/agents/rideHail/RideHailingManager.scala
```diff
@@ -90,9 +90,9 @@ class RideHailingManager(
   }
 
   // TODO: remove following (replace by RideHailModifyPassengerScheduleManager)
-  val repositioningPassengerSchedule = mutable.Map[Id[Vehicle], (Id[Interrupt],Option[PassengerSchedule])]()
-  val reservationPassengerSchedule=mutable.Map[Id[Vehicle], (Id[Interrupt],ModifyPassengerSchedule)]()
-  val repositioningVehicles = mutable.Set[Id[Vehicle]]() // TODO: move to RideHailModifyPassengerScheduleManager?
+  val repositioningPassengerSchedule: mutable.Map[Id[Vehicle], (Id[Interrupt], Option[PassengerSchedule])] = mutable.Map[Id[Vehicle], (Id[Interrupt],Option[PassengerSchedule])]()
```
JustinPihony: Is there actually a bug here? It appears that the code which `put` values is commented out. So all of the contains that WERE here are going to be empty. Same for `reservationPassengerSchedule`

src/main/scala/beam/agentsim/agents/rideHail/RideHailingManager.scala
```diff
@@ -859,14 +859,12 @@ class RideHailingManager(
...
-  def lockVehicle(vehicleId: Id[Vehicle]) = {
+  def lockVehicle(vehicleId: Id[Vehicle]): Unit = {
     lockedVehicles += vehicleId
   }
 
 
-  def getVehicleFuelLevel(vehicleId: Id[Vehicle]): Double = {
-    vehicleFuelLevel.get(vehicleId).get
-  }
+  def getVehicleFuelLevel(vehicleId: Id[Vehicle]): Double = vehicleFuelLevel(vehicleId)
```
JustinPihony: Should this return an Option? Granted this is not really used afaict. Just as an API example in `StanfordRideHailAllocationManagerV1`, so probably not a big deal...maybe even remove?

src/main/scala/beam/agentsim/agents/rideHail/allocationManagers/DefaultRideHailResourceAllocationManager.scala
```diff
@@ -14,7 +14,7 @@ class DefaultRideHailResourceAllocationManager extends RideHailResourceAllocatio
   }
 
 // TODO RW/Asif: how to make sure no one ever can call this?
-def allocateVehicles(allocationsDuringReservation: Vector[(VehicleAllocationRequest, Option[VehicleAllocation])]): Vector[(VehicleAllocationRequest, Option[VehicleAllocation])] = {
+def allocateVehicles(allocationsDuringReservation: Vector[(VehicleAllocationRequest, Option[VehicleAllocation])]): IndexedSeq[(VehicleAllocationRequest, Option[VehicleAllocation])] = {
   log.error("batch processing is not implemented for DefaultRideHailResourceAllocationManager")
     //???
   return allocationsDuringReservation
```
JustinPihony: What's the purpose of the signature change here? Underlying, the data is the same - same for other areas like this

src/main/scala/beam/agentsim/agents/rideHail/allocationManagers/StanfordRideHailAllocationManagerV1.scala
```diff
@@ -59,7 +61,7 @@ class StanfordRideHailAllocationManagerV1(val rideHailingManager: RideHailingMan
         }
       }
 
-      result = result :+ (vehicleAllocationRequest, vehicleAllocation)
+      result += ((vehicleAllocationRequest, vehicleAllocation))
     }
     result
```
JustinPihony: return `result.toIndexedSeq` to truly enforce the type

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -280,8 +281,8 @@ class R5RoutingWorker(val beamServices: BeamServices, val transportNetwork: Tran
           //TODO make a more sensible window not just 30 minutes
           startTime >= time.fromTime && startTime <= time.fromTime + 1800
         }.map(itinerary => {
-          var legsWithFares = Vector[(BeamLeg, Double)]()
-          maybeWalkToVehicle.foreach(legsWithFares +:= (_, 0.0))
+          val legsWithFares = ArrayBuffer[(BeamLeg, Double)]()
```
JustinPihony: How about just move it into a one line initializer?

`val legsWithFares = maybeWalkToVehicle.map(x=> ArrayBuffer((x, 0.0))).getOrElse(ArrayBuffer[(BeamLeg, Double)]())`

REASY:
@carloscaldas, @JustinPihony , thanks for the raised questions and I think all of them should be solved. But I'm not sure that it should be in this PR. This PR is about changing immutable types to mutable. I did not want to create extra complexity to figure out what I broke by doing a lot of changes in on PR, if I would break anything.

colinsheppard:
I concur with @JustinPihony in that we should treat immutability as a means to an end, not an end in itself. 

Internal to a method, there's no harm in using mutability for convenience (particularly convenient for devs who are still in the habit of procedural programming).  Not to say that immutability is worse, it's just not necessarily worth a lot of time refactoring or fussing over.

Internal to an Actor, immutability is also acceptable in my opinion. Everyone knows (or should know) never to attempt to directly access an Actor's internal state, so the Actor should feel free to mutate state in whatever way is most concise, understandable, and maintainable. All that said, I have been convinced by @michaz that using the Akka FMS state data as the way to manage state does have benefit because that state is passed around, put into logging, etc, which greatly facilitates debugging. It also has the benefit of making it easier to mock Actors for testing.

So, in summary, I'm happy to see this work going on, but I think the key emphasis should be put into places where we were managing global state but using mutability for convenience (BeamServices in particular), a regrettable choice made early on (by me) when I was new to Scala. We should correct this and find a way to make BeamServices immutable or make all mutable fields private and then disallow changes other than during initialization.

carloscaldas:
Immutability can be expensive: even if we we don't have to recreate the
whole object still have have to allocate memory again (even for a small
object is expensive)

I would sumarize in the following way:

-object does not leave method scope and don't make parallel
computation: *mutable
object*
-object leaves the class scope: *immutable object*
-object leaves the method scope: *immutability sometimes (should analyse
case by case)*
*-makes parallel computation over the object: immutable *(this is not a
silver bullet but will make better code in 90% of the times)

src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceLCCM.scala
REASY: It's hard to say about original intent, maybe @colinsheppard can help here.

src/main/scala/beam/agentsim/agents/rideHail/allocationManagers/DefaultRideHailResourceAllocationManager.scala
REASY: Using `IndexedSeq` allows to fill `ArrayBuffer` and return it as result, instead of converting it to vector by calling `toVector`

src/main/java/beam/agentsim/agents/choice/logit/LatentClassChoiceModel.scala
REASY: If I get this right [TraversableOnce.scala](https://github.com/scala/scala/blob/6ac6da8b61a3a427089a166c7802a940eac71064/src/library/scala/collection/TraversableOnce.scala#L308), `toIndexedSeq` will allocate new buffer and copy elements there - more GC pressure.

src/main/scala/beam/agentsim/agents/rideHail/allocationManagers/StanfordRideHailAllocationManagerV1.scala
REASY: If I get this right [TraversableOnce.scala](https://github.com/scala/scala/blob/6ac6da8b61a3a427089a166c7802a940eac71064/src/library/scala/collection/TraversableOnce.scala#L308), toIndexedSeq will allocate new buffer and copy elements there - more GC pressure.

PR #289, opened 2018-06-03
RideHailingManager: Optimized get closest idle ride hailing agent

REASY:
In `getClosestIdleRideHailingAgent` method we use only one element - so we don't really need to sort `distances2RideHailingAgents` - it's faster and less GC pressure(`sortBy` allocates new collection)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/rideHail/RideHailingManager.scala
+++ b/src/main/scala/beam/agentsim/agents/rideHail/RideHailingManager.scala
@@ -808,6 +808,11 @@ class RideHailingManager(
 
   def getClosestIdleVehiclesWithinRadius(pickupLocation: Coord, radius: Double): Vector[(RideHailingAgentLocation,
     Double)] = {
+    getIdleVehiclesWithinRadius(pickupLocation, radius).sortBy(_._2)
+  }
+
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/agentsim/agents/rideHail/RideHailingManager.scala
```diff
@@ -816,12 +821,16 @@ class RideHailingManager(
...
+    distances2RideHailingAgents.filterNot(x => lockedVehicles(x._1.vehicleId))
   }
 
   def getClosestIdleRideHailingAgent(pickupLocation: Coord,
                                      radius: Double): Option[(RideHailingAgentLocation, Double)] = {
-    getClosestIdleVehiclesWithinRadius(pickupLocation, radius).headOption
+    val idleVehicles = getIdleVehiclesWithinRadius(pickupLocation, radius)
+    if (idleVehicles.isEmpty) None
+    else {
+      Some(idleVehicles.minBy { case (location, radius) => radius })
```
JustinPihony: I assume you went with this for readability over `Some(idleVehicles.minBy(_._2))`?

REASY: Yes, right. I think this `Some(idleVehicles.minBy(_._2))` is pretty hard to read and understand, if you're not an author (even if you're author, will forget it fast :D)

PR #294, opened 2018-06-12
Optimization: Extract Event.getAttributes for memory optimization

REASY:
Cherry-picked extremely useful commit https://github.com/LBNL-UCB-STI/beam/commit/b2f031c44c29a967b92122feb9a047dc054dd9bf which reduce GC Pressure a lot (there were a lot of unnecessary allocations)


REASY:
Opened the same with proper name for CI: https://github.com/LBNL-UCB-STI/beam/pull/295

PR #297, opened 2018-06-18
Project: Added GC logs which can be helpful to investigate performance issue

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/297)


JustinPihony:
LGTM, however should this be more of an opt in, versus always on option? I see no imminent problem, but at least want to voice the question.

REASY:
if it would be be default, It could help me a lot during investigation. I'm ok to make it as optional, but not sure how to do it in gradle. Will search about it.

carloscaldas:
Please don´t merge. I am reviewing it

build.gradle
```diff
@@ -464,9 +466,17 @@ mainClassName = "beam.sim.RunBeam"
 
 def myAvailableRam = System.getenv("MAXRAM") ?: (project.findProperty('maxRAM') ?: "16g")
 
+def getCurrentTimestamp = {
+    Date today = new Date()
+    SimpleDateFormat df = new SimpleDateFormat("MM-dd-yyyy_HH-mm-ss")
```
carloscaldas: ⚠️ Consider using the appropriate `Locale` when formatting a `DateTime`, otherwise, the code will rely on the default `Locale`.

⚠️ Consider using the appropriate `ZoneOffset` otherwise, the code will rely on the default ZoneOffset

⚠️ Consider adding the type of public methods.

❗️ Every time this method is called it creates an instance of a date formatter. This can be expensive. Consider creating a constant using `DateTimeFormatter`(Java 8) because it immutable/`thread-safe` and avoid creating multiple instances.

# Suggestion

```Scala
private val formatter = DateTimeFormatter.ofPattern("MM-dd-yyyy_HH-mm-ss")
    .withLocale(Locale.US)
    .withZone(ZoneOffset.UTC)

def getCurrentTimestamp : String = formatter.format(Instant.now())

```

REASY: @carloscaldas, thanks for the mentioned things. This function is part of build process. I'm not sure that performance related things make sense there.

PR #313, opened 2018-07-11
Reduce memory allocation by using static compare

JustinPihony:
Also, remove the Tuple2 creation, used only for comparison purposes

This is pending a full test, but otherwise preliminary findings found a 33% reduction in memory footprint when used against sf-light.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/313)


colinsheppard:
Looks good, once CI comes back clean, please merge. Thanks!

PR #315, opened 2018-07-12
Rw/tnc2transit merge master 4ci (WIP) 

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/315)


src/main/java/beam/agentsim/events/ReplanningEvent.java
```diff
@@ -0,0 +1,50 @@
...
+import org.matsim.api.core.v01.Id;
+import org.matsim.api.core.v01.events.Event;
+import org.matsim.api.core.v01.population.Person;
+import org.matsim.core.api.internal.HasPersonId;
+
+import java.util.Map;
+
+public class ReplanningEvent extends Event implements HasPersonId {
+
+    public final static String EVENT_TYPE = "Replanning";
```
carloscaldas: ⚠️ This class already have a method exposing this value (`getEventType`). Consider making the constant private and avoiding redundance

src/main/java/beam/agentsim/events/ReserveRideHailEvent.java
```diff
@@ -0,0 +1,96 @@
...
+        this.originY = originY;
+        this.destinationX = destinationX;
+        this.destinationY = destinationY;
+    }
+
+    private Map<String, String> attr;
+
+    @Override
+    public Map<String, String> getAttributes() {
+        if(attr != null) return attr;
```
carloscaldas: ⚠️ Consider keeping consistent in the formatting:
- Using correct spaces
- Always using curly braces

❗️ Don´t return a mutable reference from a public method. It breaks encapsulation.

## Suggestion

Before:
```Java
public Map<String, String> getAttributes() { 
    if(attr != null) return attr; 
    attr = super.getAttributes(); 
  
    attr.put(ATTRIBUTE_PERSON, personId.toString()); 
    return attr; 
}
```

After:
```Java
public Map<String, String> getAttributes() { 
    if (attr == null) {
        attr = super.getAttributes(); 
        attr.put(ATTRIBUTE_PERSON, personId.toString()); 
    }
    return Collections.unmodifiableMap(attr);
}
```

src/main/java/beam/analysis/plots/DeadHeadingStats.java
```diff
@@ -38,79 +47,157 @@ public void processStats(Event event) {
...
-        Integer _num_passengers = getPathTraversalEventNumOfPassengers(event, eventAttributes);
-        boolean validCase = isValidCase(graphName, _num_passengers);
-        if (validCase) {
-            updateNumPassengerInDeadHeadingsMap(hour,graphName,_num_passengers);
-        }
+        Map<String, String> attributes = event.getAttributes();
+        String mode = attributes.get(PathTraversalEvent.ATTRIBUTE_MODE);
+        String vehicle_id = attributes.get(PathTraversalEvent.ATTRIBUTE_VEHICLE_ID);
+        String graphName = getGraphNameAgainstModeAndVehicleId(mode, vehicle_id);
+        Integer _num_passengers = getPathTraversalEventNumOfPassengers(event);
```
carloscaldas: ⚠️ This variable does not follow Java naming conventions. Consider using `numPassengers`

src/main/java/beam/analysis/plots/DeadHeadingStats.java
```diff
@@ -127,108 +214,73 @@ private boolean updateDeadHeadingTNCMap(double length,int hour,Integer _num_pass
...
     }
-    private List<String> getLegendItemList(String graphName,int dataSetRowCount,int bucketSize){
-        List<String> legendItemList = new ArrayList<>();
-        for (int i = 0; i < dataSetRowCount; i++) {
-            legendItemList.add(getLegendText(graphName, i,bucketSize));
+
+    private double[][] buildDeadHeadingDataSetTnc0() {
+        List<Integer> hours = GraphsStatsAgentSimEventsListener.getSortedIntegerList(deadHeadingsTnc0Map.keySet());
+        int maxHour = 0;
+        if (hours.size() > 0) {
```
carloscaldas: ⚠️ Consider using `Collections.isEmpty()` instead of `Collections.size() > 0`. It increases the readability and also the complexity time can be reduced depending on the concrete class

## Suggestion

```Java
if (!hours.isEmpty()) {
```

src/main/java/beam/analysis/plots/DeadHeadingStats.java
```diff
@@ -127,108 +214,73 @@ private boolean updateDeadHeadingTNCMap(double length,int hour,Integer _num_pass
...
             if (hourData != null) {
-                modeOccurrencePerHour[index] = hourData.get(outerLoopIndex) == null ? 0 : hourData.get(outerLoopIndex) / METERS_IN_KM;
+
+                if (hourData.get(outerLoopIndex) == null) {
+                    modeOccurrencePerHour[index] = 0;
+                } else {
+
+                    double val = hourData.get(outerLoopIndex) / METERS_IN_KM;
+//                    double val = hourData.get(outerLoopIndex);
+                    //val = Math.round(val * 100) / 100;
```
carloscaldas: ⚠️ Avoid committing commented code

src/main/java/beam/analysis/plots/RealizedModeStats.java
```diff
@@ -0,0 +1,309 @@
...
+            countOccurrenceJava(mode, 1, ShortLevel(), tags);
+            if (personIdList.contains(personId)) {
+                personIdList.remove(personId);
+                return;
+            }
+
+            Integer frequency = 1;
+            if (hourData != null) {
+                frequency = hourData.get(mode);
+                if (frequency != null) {
```
carloscaldas: ⚠️ Simplify expression to make the code less verbose and increase readability/maintainability.

## Suggestion

Before:
```Java
frequency = hourData.get(mode);
if (frequency != null) {
    frequency++;
} else {
    frequency = 1;
}
```

After:
```Java
frequency = hourData.getOrDefault(mode, 0) + 1;
```

src/main/java/beam/analysis/plots/RealizedModeStats.java
```diff
@@ -0,0 +1,309 @@
...
+                out.write(builder.toString());
+                out.newLine();
+            }
+            out.flush();
+        } catch (IOException e) {
+            log.error("CSV generation failed.", e);
+        }
+    }
+
+    class ModePerson {
```
carloscaldas: ❗️Consider making this class static otherwise every instance will have a reference to the enclosing class. This increase the memory usage and can lead to memory leaks.

PR #326, opened 2018-07-18
One last memory optimization

JustinPihony:
Finally got back to working on this - so the last item I had noted previously. This actually makes a fairly decent dent :)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/326)


colinsheppard:
New conflict, but otherwise ready to merge.

PR #349, opened 2018-07-30
Optimization: reduced quadratic complexity (space and time) to linear

carloscaldas:
## Main optimisations:
- `getDistance` replaced by accumulator: the original approach was traversing the whole `Vector` for every iteration. The impact was quadratic time. Was replaced by an accumulator with similar behaviour
- `Vector` replaced by `ListBuffer`: a new instance object was created for every iteration. Since the variable was used locally (inside the method) there is no need of being immutable.

## Small optimisations:
- Removed redundant variables
- Removed duplicate calculation

## Suggestions for future
- Find a way to do the same calculation using functional approach. The breakable creates an `Exception` including the stacktrace and intercept it. It add a huge constant time because.
- Create a test case for this method

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/349)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/RideHailUtils.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/RideHailUtils.scala
@@ -1,5 +1,7 @@
 package beam.agentsim.agents.ridehail
 
+import scala.collection.mutable.ListBuffer
+
 import beam.agentsim.events.SpaceTime
 import beam.router.RoutingModel
 import beam.router.RoutingModel.BeamLeg
@@ -7,7 +9,6 @@ import beam.utils.GeoUtils
 import com.conveyal.r5.profile.{ProfileRequest, StreetMode}
 import com.conveyal.r5.transit.TransportNetwork
 import org.matsim.api.core.v01.{Coord, Id}
-
 import scala.util.control.Breaks._
 
 object RideHailUtils {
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/agentsim/agents/ridehail/RideHailUtils.scala
```diff
@@ -18,51 +19,48 @@ object RideHailUtils {
...
-      endPoint = updatedEndPoint,
-      distanceInM = updatedDistanceInMeters
-    )
-    val updatedDuration = (stopTime - originalBeamLeg.startTime).toLong
+      val updatedEndPoint = SpaceTime(endPointLocation, stopTime.toLong)
 
-    originalBeamLeg.copy(duration = updatedDuration, travelPath = updatedTravelPath)
-  }
+      val updatedTravelPath = originalBeamLeg.travelPath.copy(
+        linkIds = linkIds.toVector,
```
REASY: `linkIds.toVector` will allocate new buffer for `Vector`. Why just don't change `BeamPath. linkIds` type to `Seq[Int]` and do not need any conversion between collections . Also I would go with `ArrayBuffer` instead of `ListBuffer` due to data locality (more data will be in cache)

carloscaldas: Hi @REASY ,
According to the documentation `https://www.scala-lang.org/api/current/scala/collection/mutable/ListBuffer.html`
```Scala
def toList: immutable.List[A]
```
> Converts this buffer to a list. Takes constant time. The buffer is copied lazily, the first time it is mutated.
Constant time means doesn´t allocate new buffer

carloscaldas: For some reason, I have `toList` in mind :-)

PR #351, opened 2018-07-30
Improved stuck agent detection

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/351)


src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -309,8 +300,8 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
   }
 
   override def postStop(): Unit = {
-    monitorTask.foreach(_.cancel())
-    skipOverBadActorsTask.foreach(_.cancel())
+    scheduleMonitorTask.foreach(_.cancel())
```
JustinPihony: Shouldn't these use `monitorTask` and `stuckAgentChecker`?

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -105,43 +113,14 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
...
-          new FiniteDuration(
-            math.round(beamConfig.beam.debug.secondsToWaitForSkip / 4.0),
-            TimeUnit.SECONDS
-          ),
-          self,
-          SkipOverBadActors
-        )
-      )
-    else None
+  private val stToStuckTimes: mutable.HashMap[ScheduledTrigger, Int] = mutable.HashMap.empty
```
JustinPihony: This is understandable only when you see the type. Otherwise it isn't easily understood and should be `scheduledTriggerToStuckTimes` or even just `triggerToStuckTimes`

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -181,10 +169,12 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
       }
       val completionTickOpt = triggerIdToTick.get(triggerId)
       if (completionTickOpt.isEmpty || !triggerIdToTick.contains(triggerId) || !awaitingResponse
-            .containsKey(completionTickOpt.get)) {
+        .containsKey(completionTickOpt.get)) {
```
JustinPihony: A bit of an aside for this one, but seems like it wouldn't hurt here. Why not use pattern matching here...as well as the `contains` seems duplicative.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -181,10 +169,12 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
...
         log.error(s"Received bad completion notice $notice from ${sender().path}")
       } else {
-        awaitingResponse.remove(completionTickOpt.get, triggerIdToScheduledTrigger(triggerId))
+        val st = triggerIdToScheduledTrigger(triggerId)
+        awaitingResponse.remove(completionTickOpt.get, st)
```
JustinPihony: Using the stuckFinder makes awaitingResponse unnecessary unless I missed something. Shouldn't we JUST use that instead?

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -207,55 +197,54 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
...
-      var numReps = 0L
-      currentTotalAwaitingResponse = awaitingResponse.values().stream().count()
-      if (currentTotalAwaitingResponse == previousTotalAwaitingRespone && currentTotalAwaitingResponse != 0) {
-        numberRepeats += 1
-        numReps = numberRepeats
-        log.debug("DEBUG: {} repeats.", numReps)
-      } else {
-        numberRepeats = 0
-      }
+      val stuckAgents = stuckFinder.detectStuckAgents()
```
JustinPihony: Should most of this be encapsulated into the detector itself? Then it could just be more like stuckfinder.register and unregister - letting it handle the check itself. Then you would only need to add the values to be monitored.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -207,55 +197,54 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
...
+        log.warning("Cleaning {} non-special agents", canClean.size)
+        canClean.foreach { stuckInfo =>
+          val st = stuckInfo.value
+          st.agent ! IllegalTriggerGoToError("Stuck Agent")
+          self ! CompletionNotice(st.triggerWithId.triggerId)
+          log.warning("Cleaned {}", st)
+        }
+
+        val special = stuckAgents.diff(canClean)
+        log.warning("Processing {} special agents", special.size)
```
JustinPihony: Should this be renamed to `unexpectedStuckAgents` rather than special?

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -207,55 +197,54 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
...
+        special.foreach { stuckInfo =>
+          val st = stuckInfo.value
+          val times = stToStuckTimes.getOrElse(st, 0)
+          stToStuckTimes.put(st, times + 1)
+          // We have to add them back to `stuckFinder`
+          if (times < 50) {
```
JustinPihony: Should these values be configurable?

src/main/scala/beam/utils/StuckFinderHelper.scala
```diff
@@ -0,0 +1,54 @@
...
+
+class StuckFinderHelper[K] {
+
+  object Comparator extends Comparator[ValueWithTime[K]] {
+    def compare(o1: ValueWithTime[K], o2: ValueWithTime[K]): Int = {
+      o1.time.compare(o2.time)
+    }
+  }
+
+  private[this] val pq = new PriorityQueue[ValueWithTime[K]](Comparator)
```
JustinPihony: Any particular reason you chose Java's PQ over Scala's? Just curious.

src/main/scala/beam/utils/StuckFinder.scala
```diff
@@ -0,0 +1,59 @@
...
+  def removeByKey(st: ScheduledTrigger): Option[ValueWithTime[ScheduledTrigger]] = {
+    helper.removeByKey(st)
+  }
+
+  def detectStuckAgents(time: Long = System.currentTimeMillis()): Seq[ValueWithTime[ScheduledTrigger]] = {
+    @tailrec
+    def detectStuckAgents0(stuckAgents: mutable.ArrayBuffer[ValueWithTime[ScheduledTrigger]]): Seq[ValueWithTime[ScheduledTrigger]] = {
+      removeOldest match {
+        case Some(oldest) =>
+          val isStuck: Boolean = isStuckAgent(oldest.value, oldest.time, time)
```
JustinPihony: How about pushing the isStuckAgent into the removeOldest (And possibly rename) as a predicate so that you avoid thrashing too much. Just pull out the agents that are actually stuck.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
REASY: Yes, pattern matching would be better here, but I don't want to touch anything extra because there are no tests for `BeamAgentScheduler`

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
REASY: Good point. It seems to me that `StuckFunder` can cover the cases where `awaitingResponse` is used. @colinsheppard ?

src/main/scala/beam/utils/StuckFinderHelper.scala
REASY: [Scala Priority Queue](https://www.scala-lang.org/api/2.12.0/scala/collection/mutable/PriorityQueue.html) does not have an API to remove particular element from the queue

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
REASY: I'd like to leave `StuckFinder` as simple as it is now. I think `BeamAgentScheduler` is right place to make decision what to do with stuck agents.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -12,13 +12,21 @@ import beam.agentsim.events.EventsSubscriber._
...
+import beam.utils.StuckFinder
 import com.google.common.collect.TreeMultimap
 
 import scala.annotation.tailrec
 import scala.collection.mutable
 import scala.concurrent.ExecutionContext.Implicits.global
 import scala.concurrent.duration.FiniteDuration
 
+
+case class RideHailingManagerIsExtremelySlowException
```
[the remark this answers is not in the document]
REASY: I named is as is because it was especially related with `RideHailingManager`

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
REASY: Hm, I was wrong, the tests are in `BeamAgentSchedulerSpec`

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
JustinPihony: In the sake of keeping the execution centralized, how about using a function then? The user of the class can inject their specific logic for what to do with stuck agents. I'm only pushing this one because it seems like a bit more back and forth - especially on removing and re-putting objects.

src/main/scala/beam/utils/StuckFinderHelper.scala
JustinPihony: I guess that begs the question of why that is needed? It defeats the purpose of the PQ otherwise? Additionally, the only place I see you using it is with `removeByKey`, which isn't needed after the poll and the other usage I guess will depend on the open question in BeamAgentScheduler that you pinged Colin on. So it can be re-addressed once that is cleared up.

src/main/scala/beam/utils/StuckFinder.scala
```diff
@@ -0,0 +1,55 @@
...
+  def removeByKey(st: ScheduledTrigger): Option[ValueWithTime[ScheduledTrigger]] = {
+    helper.removeByKey(st)
+  }
+
+  def detectStuckAgents(time: Long = System.currentTimeMillis()): Seq[ValueWithTime[ScheduledTrigger]] = {
+    @tailrec
+    def detectStuckAgents0(stuckAgents: mutable.ArrayBuffer[ValueWithTime[ScheduledTrigger]]): Seq[ValueWithTime[ScheduledTrigger]] = {
+      helper.removeOldest match {
+        case Some(oldest) =>
+          val isStuck: Boolean = isStuckAgent(oldest.value, oldest.time, time)
```
JustinPihony: The changes applied still don't address my comment of pushing this into the helper itself. If you do it there then it seems that you could avoid thrashing (removing and re-adding).

colinsheppard:
@JustinPihony I believe your changes were all addressed.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -76,16 +84,17 @@ object BeamAgentScheduler {
   }
 
   def SchedulerProps(
-    beamConfig: BeamConfig,
-    stopTick: Double = 3600.0 * 24.0,
-    maxWindow: Double = 1.0
-  ): Props = {
-    Props(classOf[BeamAgentScheduler], beamConfig, stopTick, maxWindow)
+                      beamConfig: BeamConfig,
+                      stopTick: Double = 3600.0 * 24.0,
```
carloscaldas: Is it a unit of time like number of seconds in a day?
some questions:
- why double?
- put the unit of the time in the variable
- use something more meaningful like `TimeUnit.DAYS.toSeconds(1)`

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -207,55 +198,54 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
 
     case Monitor =>
       log.debug(
-        s"\n\tnowInSeconds=$nowInSeconds,\n\tawaitingResponse.size=${awaitingResponse
-          .size()},\n\ttriggerQueue.size=${triggerQueue.size},\n\ttriggerQueue.head=${triggerQueue.headOption}\n\tawaitingResponse.head=$awaitingToString"
+        s"\n\tnowInSeconds=$nowInSeconds,\n\tawaitingResponse.size=${
+          awaitingResponse
+            .size()
+        },\n\ttriggerQueue.size=${triggerQueue.size},\n\ttriggerQueue.head=${triggerQueue.headOption}\n\tawaitingResponse.head=$awaitingToString"
```
carloscaldas: This log is very hard to read. Consider:
- replacing `\n` with a constant(or local variable) like: `val EOL = System.lineSeparator()`
- use scala multiline string to improve readability

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -207,55 +198,54 @@ class BeamAgentScheduler(val beamConfig: BeamConfig, stopTick: Double, val maxWi
...
-      currentTotalAwaitingResponse = awaitingResponse.values().stream().count()
-      if (currentTotalAwaitingResponse == previousTotalAwaitingRespone && currentTotalAwaitingResponse != 0) {
-        numberRepeats += 1
-        numReps = numberRepeats
-        log.debug("DEBUG: {} repeats.", numReps)
-      } else {
-        numberRepeats = 0
-      }
+      val stuckAgents = stuckFinder.detectStuckAgents()
+      if (stuckAgents.size > 0) {
```
carloscaldas: Replace `.size` with `nonEmpty`.

colinsheppard:
Just to reiterate from a recent discussion, we should get some experience with this new feature by using on it on the production branch. I.e. please test it out and tune the thresholds on application-sfbay (or a clone of that branch merged with this branch) and make sure that it works (i.e. reduces detection time) with minimal false positives.

REASY:
1) Get measurements for 99.99-th percentile
2) Create different thresholds for different type of trigger message

REASY:
I'll choose thresholds for Iteration 0 from this report: https://docs.google.com/spreadsheets/d/1fGQxtsFqQGj0AhgQ7D1cic5SIf_7tZSO6xGN1O5D100/edit#gid=0

PR #388, opened 2018-08-14
Memory Optimizations for master branch

REASY:


src/main/java/beam/physsim/jdeqsim/AgentSimToPhysSimPlanConverter.java
```diff
@@ -253,8 +253,9 @@ private Leg createLeg(String mode, String links, double departureTime) {
         // hack: removing non-road links from route
         // TODO: debug problem properly, so that no that no events for physsim contain non-road links
         List<Id<Link>> removeLinks = new ArrayList<>();
+        Map<Id<Link>, ? extends Link> networkLinks = agentSimScenario.getNetwork().getLinks();
```
REASY: image

src/main/java/beam/agentsim/events/handling/BeamEventsWriterCSV.java
```diff
@@ -89,7 +89,9 @@ protected void writeEvent(Event event) {
                 String str = row[i];
                 if (str != null) {
                     if (str.contains(",")) {
-                        this.out.append("\"" + str + "\"");
+                        this.out.append('"');
```
REASY: image

src/main/scala/beam/router/RoutingModel.scala
```diff
@@ -141,17 +134,44 @@ object RoutingModel {
...
+          events += new LinkLeaveEvent(timeAtNode, vehicleId, Id.createLinkId(from))
+          events += new LinkEnterEvent(timeAtNode, vehicleId, Id.createLinkId(to))
+      }
+      events.toIterator
+    } else {
+      Iterator.empty
+    }
+  }
+
+  def traverseStreetLeg_opt(leg: BeamLeg, vehicleId: Id[Vehicle]): Iterator[Event] = {
```
REASY: image

src/main/java/beam/agentsim/events/handling/BeamEventsWriterCSV.java
carloscaldas: Comments: Good job here!

Recommended suggestions:
- Use fluent interface. It will become a little bit faster because will avoid some indirections and some garbage

Suggestion

```Java
out.append('"')
    .append(str)
    .append('"');
```

src/main/java/beam/analysis/physsim/PhyssimCalcLinkStats.java
```diff
@@ -52,7 +52,7 @@
     private BeamConfig beamConfig;
     private Network network;
     private OutputDirectoryHierarchy controllerIO;
-    private BeamCalcLinkStats linkStats;
+    public BeamCalcLinkStats linkStats;
```
carloscaldas: Recommended suggestion:
- Don´t expose fields as public. It breaks encapsulation. Let it private and if needed create accessor/mutator method.

src/main/scala/beam/agentsim/agents/BeamAgent.scala
```diff
@@ -83,23 +83,7 @@ trait BeamAgent[T] extends LoggingFSM[BeamAgentState, T] with Stash {
...
-    log.warning(s"${logWithFullPrefix(msg)}")
-  }
-
-  def logError(msg: String): Unit = {
-    log.error(s"${logWithFullPrefix(msg)}")
-  }
-
-  def logDebug(msg: String): Unit = {
-    log.debug(s"${logWithFullPrefix(msg)}")
+    s"$tickStr${triggerStr}State:$stateName ${logPrefix()}"
```
carloscaldas: Previously was logging something. Now creates a string and nothing happens. Looks like a dangling change here.

src/main/scala/beam/sim/FakeTravelTimeCalculator.scala
```diff
@@ -0,0 +1,38 @@
...
+  override def handleEvent(e: LinkLeaveEvent): Unit = {}
+
+  override def handleEvent(event: VehicleAbortsEvent): Unit = {}
+
+  override def handleEvent(event: VehicleArrivesAtFacilityEvent): Unit = {}
+
+  override def handleEvent(event: VehicleEntersTrafficEvent): Unit = {}
+
+  override def handleEvent(event: VehicleLeavesTrafficEvent): Unit = {}
+}
```
carloscaldas: Add posix end line here

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -195,14 +189,16 @@ trait ChoosesMode {
                 filterStreetVehiclesForQuery(streetVehicles, mode).headOption
               maybeVehicle match {
                 case Some(vehicle) =>
+                  val linkIds = new ArrayBuffer[Int](2 + r.getLinkIds.size())
+                  linkIds += r.getStartLinkId.toString.toInt
+                  r.getLinkIds.asScala.foreach { id => linkIds += id.toString.toInt }
+                  linkIds += r.getStartLinkId.toString.toInt
```
carloscaldas: I would rather follow scala approach. But not sure about performance in this case.
```Scala
val linkIdList = List(r.getStartLinkId) 
        ::: r.getLinkIds.asScala.toList
        ::: List(r.getStartLinkId)
val linkIds = linkIdList.map(_.toString.toInt)
```

src/main/java/beam/agentsim/events/PathTraversalEvent.java
```diff
@@ -76,6 +78,7 @@ public PathTraversalEvent(double time, Id<Vehicle> vehicleId, String vehicleType
         this.startY = startY;
         this.endX = endX;
         this.endY = endY;
+        this.attributes = new AtomicReference<Map<String, String>>(null);
```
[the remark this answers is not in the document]
REASY: I pass `null` there on purpose - to reduce memory consumption. Even empty HashMap is not free, around [128 bytes of memory](https://www.ibm.com/developerworks/library/j-codetoheap/index.html#HashMap). Keep in mind we have a lot of events.

src/main/java/beam/physsim/jdeqsim/AgentSimToPhysSimPlanConverter.java
JustinPihony: Nice catch

src/main/scala/beam/agentsim/agents/BeamAgent.scala
JustinPihony: No, he ripped it out in favor of explicit `log.debug`. I'm not sure that should be the path here, though. If it's the creation of `msg: String` that is the issue then how about making it a byval parameter:

`def logDebug(msg: => String): Unit = { ....`

Then you can keep the prefix portion encapsulated, and not something users have to remember to use when calling log. You could even make it a variable length argument if you want to avoid string building by the callers. The primary point here is that this change now forces users to have to know/remember to add the prefix to their logs.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
JustinPihony: Yah...the more functional approach comes at an allocation cost, which is what is being tackled here.

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -323,21 +322,15 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with HasServices {
       // Produce link events for this trip (the same ones as in PathTraversalEvent).
       // TODO: They don't contain correct timestamps yet, but they all happen at the end of the trip!!
       // So far, we only throw them for ExperiencedPlans, which don't need timestamps.
+      val head = data.passengerSchedule.schedule
+        .drop(data.currentLegPassengerScheduleIndex).head
```
JustinPihony: Unless I missed an area, you could even just get the first part of the tuple here and use it...as the second part is never used anyway.

src/main/scala/beam/agentsim/agents/planning/BeamPlan.scala
```diff
@@ -60,11 +60,11 @@ class BeamPlan extends Plan {
   //////////////////////////////////////////////////////////////////////
   // Beam-Specific methods
   //////////////////////////////////////////////////////////////////////
-  def trips: Vector[Trip] = tours.flatMap(_.trips)
+  lazy val trips: Vector[Trip] = tours.flatMap(_.trips)
```
JustinPihony: Are we sure that `tours` never changes? If so then maybe it should be implemented that way, otherwise somebody could go and change it...then `trips`, `activities`,... will not reflect accurately.

REASY: From my understanding yes, it will never change, it's initialized once. I could have rewritten `BeamPlan` in a bit better way without having `var`, but just did quick fix

JustinPihony: I'd argue that this is an important area - I guess if you create an issue so that it doesn't get forgot then that'd be an acceptable workaround.

src/main/scala/beam/agentsim/agents/BeamAgent.scala
```diff
@@ -83,23 +83,23 @@ trait BeamAgent[T] extends LoggingFSM[BeamAgentState, T] with Stash {
...
 
-  def logError(msg: String): Unit = {
-    log.error(s"${logWithFullPrefix(msg)}")
+  def logError(msg:  => String): Unit = {
+    log.error("{} {}", getPrefix, msg)
   }
 
   def logDebug(msg: String): Unit = {
-    log.debug(s"${logWithFullPrefix(msg)}")
+    log.debug("{} {}", getPrefix, msg)
```
JustinPihony: You missed turning this one lazy. Also, I noticed you opted to not make this take args itself. Just as an FYI I'm not sure how string interpolation works with lazy evaluation from an allocation timing perspective.

REASY: Generated code also proves that it's lazy:
image

JustinPihony:
@REASY I added a few areas that were optimized in the cluster code, but not here. Note that there are even more files changed than I touched due to scalafmt. Feel free to revert any of the changes. 

The only area I did NOT push in from the cluster is from src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala   where it uses `if (numReps > 2) { ` instead of 4. I'm not sure if there was a reason for that, but at least wanted to point it out.

REASY:
The reason for `if (numReps > 2) ` was to make stuck agent detection a bit faster (if I remember correct)

JustinPihony:
@REASY I believe this is good to go - but you should be the one to merge

REASY:
GC logs for on sf-light-25k: 
- [master with commit d47e98c4d2e5be92d46348d3f1985f4841a6c136](http://gceasy.io/my-gc-report.jsp?p=c2hhcmVkLzIwMTgvMDgvMjkvLS1nY18wOC0yOS0yMDE4XzE0LTM1LTQ2LmxvZy0tMTUtNi01OQ==)
- [this branch](http://gceasy.io/my-gc-report.jsp?p=c2hhcmVkLzIwMTgvMDgvMjkvLS1nY18wOC0yOS0yMDE4XzE1LTAwLTM2LmxvZy0tMTUtMTQtNDM=)
The difference is very small due to the size of model. Need to run on sf-bay

REASY:
GC logs on sf-bay with 100k agents:
- [master](http://gceasy.io/my-gc-report.jsp?p=c2hhcmVkLzIwMTgvMDgvMzAvLS1nY18wOC0yOS0yMDE4XzIwLTAwLTU2X25vX21lbV9vcHRpbWl6LmxvZy0tMi0zOC05)

PR #390, opened 2018-08-14
Ccaldas/331 buf ridehail request 4ci (WIP)

carloscaldas:


src/main/scala/beam/sim/BeamMobsim.scala
```diff
@@ -140,7 +141,7 @@ class BeamMobsim @Inject()(
             classOf[BeamAgentScheduler],
             beamServices.beamConfig,
             Time.parseTime(beamServices.beamConfig.matsim.modules.qsim.endTime),
-            300.0
+            30.0
```
wrashid: please change back to 300

PR #405, opened 2018-08-20
Cjrs/#403 notify resource idle 4ci

colinsheppard:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/405)


colinsheppard:
tests are passing:

https://beam-ci.tk/job/beam-4ci/852/

PR #407, opened 2018-08-21
Fdariasm/#333 sioux fall scenario 4ci

colinsheppard:
I went through the process and it worked for me. 

Would be good if one more person could go through the process, see docs/users.rst or see the docs:

https://beam.readthedocs.io/en/fdariasm--333-siouxfallscenario/users.html#converting-a-matsim-scenario-to-run-with-beam

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/407)


colinsheppard:
latest build, posting here before it completed:

https://beam-ci.tk/job/beam-4ci/855/console

colinsheppard:
The one failed test passes locally... is is a non-deterministic?

image

PR #425, opened 2018-08-23
Cs/#406 consider ev range 4ci

colinsheppard:
CI build hasn't started, but the review can begin. Thank you!

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/425)


src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
```diff
@@ -0,0 +1,73 @@
...
+        routeReqToDriverMap.getOrElse(response.requestId.get, dummyDriverId)
+      }
+
+    // If no route responses, then we have no agents in mind
+    var agentLocationOpt = if (routeResponsesByDriver.isEmpty) {
+      None
+    } else {
+      // Otherwise, we look through all routes returned and find the first one that
+      // corresponds to an idle driver and use this as our agentLocation
+      val maybeId = routeResponsesByDriver.keys.find(rideHailManager.getIdleVehicles.contains(_))
```
REASY: You could rewrite it without `_`, like this:
```
val maybeId = routeResponsesByDriver.keys.find(rideHailManager.getIdleVehicles.contains)
```

src/main/scala/beam/agentsim/agents/ridehail/allocation/ImmediateDispatchWithOverwrite.scala
```diff
@@ -22,14 +18,24 @@ class ImmediateDispatchWithOverwrite(val rideHailManager: RideHailManager)
...
     }
 
     // just go with closest request
-    None
+    rideHailManager
+      .getClosestIdleVehiclesWithinRadius(
+        vehicleAllocationRequest.request.pickUpLocation,
+        rideHailManager.radiusInMeters
+      )
+      .headOption match {
```
REASY: If you need only head element, it's better to use `getClosestIdleRideHailAgent`

src/main/scala/beam/agentsim/agents/ridehail/allocation/RideHailResourceAllocationManager.scala
```diff
@@ -20,7 +28,20 @@ abstract class RideHailResourceAllocationManager(private val rideHailManager: Ri
 
   def proposeVehicleAllocation(
     vehicleAllocationRequest: VehicleAllocationRequest
-  ): Option[VehicleAllocation]
+  ): VehicleAllocationResponse = {
+    // closest request
+    rideHailManager
+      .getClosestIdleVehiclesWithinRadius(
```
REASY: `getClosestIdleRideHailAgent` would be more efficient

PR #429, opened 2018-08-24
Fdariasm/#320 reserved parking spec

fdariasm:
Spec added for Reserved parking concept in Zonal Parking Manager:

Depot parking in ZonalParkingManager should return parking stalls according to reservedFor field
 - should return only rideHailManager stalls when all parking are reservedFor RideHailManager
 - should return some stalls when all parking are reservedFor Any

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/429)


colinsheppard:
Looks good, but please let me know when the CI build is complete with this test integrated.

Issue #434, opened 2018-08-25
debugging


colinsheppard:
fix VOT for beamville and StuckAgent issue

PR #435, opened 2018-08-25
Cs/#434 debugging 4ci

colinsheppard:
still pending CI, but these should fix a couple of bugs and get some ignored tests back into the suite.

issue #434  

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/435)


colinsheppard:
Tests passed, going to do the merge for the sake of expediency and b/c the changes were small and focused.

https://beam-ci.tk/job/beam-4ci/912/

PR #446, opened 2018-08-27
Added a new reference for osm-lib

carloscaldas:
In the the fork the file logback.xml was removed.


wrashid:
please disable logback class path abbreviations (e.g. "b.a.scheduler.BeamAgentScheduler").

also disable messages from ch.qos.logback.* in log

PR #466, opened 2018-08-29
Documentation: Scala tips

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/466)

### DIFF
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -329,3 +329,108 @@ This code marks the test with `com.beam.tags.Periodic` tag. You can also specify
 You can find details about scheduling a continuous integration build under DevOps section `Configure Periodic Jobs`_.
 
 .. _Configure Periodic Jobs: http://beam.readthedocs.io/en/latest/devops.html#configure-periodic-jobs
+
+Scala tips
+^^^^^^^^^^
+Scala Collection
+~~~~~~~~~~~~~~~~
+
+Use ``mutable`` buffer instead of ``immutable var``:
+****************************************************
+
+::
+
+   // Before
+   var buffer = scala.collection.immutable.Vector.empty[Int]
+   buffer = buffer :+ 1
+   buffer = buffer :+ 2
+
+   // After
+   val buffer = scala.collection.mutable.ArrayBuffer.empty[Int]
+   buffer += 1
+   buffer += 2
+
+Don’t create temporary collections, use ``view``:
+*************************************************
+
+::
+
+   val seq: Seq[Int] = Seq(1, 2, 3, 4, 5)
+
+   // Before
+   seq.map(x => x + 2).filter(x => x % 2 == 0).sum
+
+   // After
+   seq.view.map(x => x + 2).filter(x => x % 2 == 0).sum
+
+Don’t emulate ``collectFirst``:
+*******************************
+
+::
+
+   // Get first number >= 4
+   val seq: Seq[Int] = Seq(1, 2, 10, 20)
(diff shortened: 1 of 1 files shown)


docs/developers.rst
```diff
@@ -329,3 +329,108 @@ This code marks the test with `com.beam.tags.Periodic` tag. You can also specify
 You can find details about scheduling a continuous integration build under DevOps section `Configure Periodic Jobs`_.
 
 .. _Configure Periodic Jobs: http://beam.readthedocs.io/en/latest/devops.html#configure-periodic-jobs
+
+Scala tips
+^^^^^^^^^^
+Scala Collection
+~~~~~~~~~~~~~~~~
+
+Use ``mutable`` buffer instead of ``immutable var``:
```
JustinPihony: I was going to comment on this earlier, but waffled on it a bit due to our allocation issues - should we point out that var in general should be avoided, preferring val?

docs/developers.rst
```diff
@@ -329,3 +329,108 @@ This code marks the test with `com.beam.tags.Periodic` tag. You can also specify
...
+   var buffer = scala.collection.immutable.Vector.empty[Int]
+   buffer = buffer :+ 1
+   buffer = buffer :+ 2
+
+   // After
+   val buffer = scala.collection.mutable.ArrayBuffer.empty[Int]
+   buffer += 1
+   buffer += 2
+
+Don’t create temporary collections, use ``view``:
```
JustinPihony: I don't know if this needs to be something written. At minimum it should be described a bit more - maybe point to https://www.scala-lang.org/blog/2017/11/28/view-based-collections.html

docs/developers.rst
```diff
@@ -329,3 +329,108 @@ This code marks the test with `com.beam.tags.Periodic` tag. You can also specify
...
+
+   // Before
+   seq.filter(predicate).headOption
+
+   // After
+   seq.collectFirst { case num if predicate(num) => num }
+
+Don’t emulate ``collect``:
+**************************
+
```
JustinPihony: IMHO this can be combined with `collectFirst`. They are the same concept.

docs/developers.rst
```diff
@@ -329,3 +329,108 @@ This code marks the test with `com.beam.tags.Periodic` tag. You can also specify
...
+   val s: Seq[String] = Seq("C#", "C++", "C", "Scala", "Haskel")
+   val predicate: String => Boolean = (s: String)  => { s.size > 3 }
+
+   // Before
+   s.filter(predicate).map { s => s.head }
+
+   // After
+   s.collect { case curr if predicate(curr) => curr.head }
+
+Prefer not to use ``_1, _2,...`` for ``Tuple`` to improve readability:
```
JustinPihony: Should this be boiled down to prefer pattern matching? And should there be an example of `val` extraction? 
i.e `val (key, value) = tuple2`

docs/developers.rst
REASY: For me it's like `common sense` when you code in Scala, that's why I did not mention it here. Should I?

docs/developers.rst
REASY: Added an example with destructive binding for tuple

docs/developers.rst
JustinPihony: If you come from Java then it is not common sense. I can go either way, though.

REASY:
Last thing: should I mention other's companies guidelines, for example:
- [databricks](https://github.com/databricks/scala-style-guide)
- [Effective Scala by twitter](http://twitter.github.io/effectivescala/)

PR #472, opened 2018-08-30
RideHailManager: Use getClosestIdleRideHailAgent

REASY:
Use `getClosestIdleRideHailAgent` instead of `getClosestIdleVehiclesWithinRadius` because we need only head:

image

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/472)


colinsheppard:
getClosestAgent is a wrapper to getClosests agents.... so I don't think this will improve performance

PR #497, opened 2018-09-03
Project: Use getClosestIdleRideHailAgent instead of getClosestIdleVehiclesWithinRadius

REASY:
Use `getClosestIdleRideHailAgent` instead of `getClosestIdleVehiclesWithinRadius` because **only head element** is consumed
image

Use lazy logging instead of strict one:
image

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/497)


colinsheppard:
I'm ok with merging this... I will handle conflicts with EV branch.

PR #507, opened 2018-09-04
Cs/#426 ev range 4ci

colinsheppard:
It's very close on CI.... I will sent a reminder when it passes.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/507)


colinsheppard:
Closing in favor of using the combined EV and calibration branches.

PR #524, opened 2018-09-06
Jp/#342 temp cluster code 4ci

JustinPihony:
OK, here are some of the changes over what was already in rw/#342-router-perf-artavazd-do-not-batch-updated-matsim

* Initialization of BeamServices from config in R5RoutingWorker needed some updating to mesh with master
* Any conf may need updated. I especially have in my backlog whether we need to separate some of the config into a separate cluster one that gets blended in at runtime. Note that I didn't even push in beam_worker.conf as I have never seen it used.
* I changed BeamHelper to better handle command line arguments. You can even run --help and it'll provide some meaningful data. The primary changes was to enable local vs cluster vs clustertype at minimum
* RunBeamCluster was rolled into BeamHelper and DeadLetterReplayer
* sf-light-10k.conf/beam.conf -> did not keep beam.metrics.level = "verbose"
* The initialization that was done in BeamMobSim is now done in the router
* Removed the serialization work since Kryo resolves all of that
* BeamRouter  metricsPrinter ! Subscribe("histogram", "**") is only done in local/master - not sure if that's ok or not

Areas of note I'd like confirmed by @REASY  at minimum:

* In R5RoutingWorker my biggest area of questioning is whether/why getPlanUsingCache is no longer needed and only using cache.get(
* I also did not re-add  Kamon.counter("receiving-routing-requests") to R5RoutingWorker
* https://github.com/LBNL-UCB-STI/beam/commits/rw/router-perf-artavazd-do-not-batch-updated-matsim all items from Aug 7 and up were not included. They appeared to be ancillary items that can be re-added if need be later - but not critical to the initial push into master

Some items totally omitted from the original cluster branch as they were more focused on stats and Kamon update as far as I could tell:

===TRACKING DATA === *I believe Art removed most of these anyway*
RoutingResponse with createdAt (and other timing items) - affects tests 
src/main/scala/beam/agentsim/agents/modalBehaviors/ChoosesMode.scala
src/main/scala/beam/agentsim/agents/rideHail/RideHailingManager.scala
src/main/scala/beam/utils/Statistics.scala
src/main/scala/beam/utils/RoutingRequestSenderCounter.scala
REMOVED -> src/main/scala/beam/utils/RoutingRequestResponseStats.scala
===KRYO=== *I don't believe anything beyond what I already added is needed*
https://github.com/LBNL-UCB-STI/beam/commit/1dabbdc18f5d4bdd7dcc674aed2c8f1d2a93aa44 
===KAMON=== *These will be useful when we upgrade Kamon - but that isn't necessary here*
src/main/scala/beam/sim/metrics/MetricsSupport.scala
src/main/scala/beam/sim/metrics/MetricsPrinter.scala
src/main/scala/beam/sim/BeamHelper.scala

Ultimately, the core work of the cluster changes is in this PR - the rest was pushed in as perf changes or can be followed up with other changes.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/524)


JustinPihony:
Log output is here https://drive.google.com/file/d/1qnAh125wpr6DcGe334J9khaSXHgZUzCc/view?usp=sharing

PR #532, opened 2018-09-07
Fdariasm/#433 refactor vehicle types 4ci

fdariasm:
- I'm using vehicleCategory wherever description was required
- I'm using None wherever vehicleAttributes was required, the value seems not to be used anywhere.
- I'm using 0 as Default cost for DrivingCostDefaults.estimateDrivingCost when there is no vehicleType. Not sure if this is possible.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/532)


src/main/resources/beam-template.conf
```diff
@@ -26,6 +26,9 @@ beam.agentsim.agents.modalBehaviors.mulitnomialLogit.params.ride_hail_intercept
 beam.agentsim.agents.modalBehaviors.mulitnomialLogit.params.walk_intercept = "double | 0.0"
 beam.agentsim.agents.modalBehaviors.mulitnomialLogit.params.bike_intercept = "double | 0.0"
 beam.agentsim.agents.modalBehaviors.lccm.paramFile = ${beam.inputDirectory}"/lccm-long.csv"
+
+beam.agentsim.agents.modalBehaviors.lccm.paramFile = ${beam.inputDirectory}"/lccm-long.csv"
```
wrashid: is there some duplication here?

PR #575, opened 2018-09-13
Clean up code - Simplified expression

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/575)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/RideHailModifyPassengerScheduleManager.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/RideHailModifyPassengerScheduleManager.scala
@@ -148,19 +148,8 @@ class RideHailModifyPassengerScheduleManager(
         assert(vehicleId == modifyPassengerScheduleStatus.vehicleId)
         assert(tick == modifyPassengerScheduleStatus.tick)
 
-//        log.debug(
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/agentsim/agents/ridehail/RideHailModifyPassengerScheduleManager.scala
```diff
@@ -148,19 +148,8 @@ class RideHailModifyPassengerScheduleManager(
...
-          mutable.ListBuffer[RideHailModifyPassengerScheduleStatus]()
-
-        for (modifyPassengerScheduleStatus <- getWithVehicleIds(
-               modifyPassengerScheduleStatus.vehicleId
-             )) {
-          if (modifyPassengerScheduleStatus.interruptOrigin == InterruptOrigin.RESERVATION) {
-            reservationModifyPassengerScheduleStatus += modifyPassengerScheduleStatus
-          }
-        }
+        val reservationModifyPassengerScheduleStatus = getWithVehicleIds(modifyPassengerScheduleStatus.vehicleId)
```
JustinPihony: It's worth noting that we should be cautious of these changes. As this will result in additional allocations and GC hits depending on the size of the list. Maybe we should always specifically comment when we make use of mutable buffers for performance reasons? @REASY FYI

PR #576, opened 2018-09-13
Clean up - Removed 20 warnings

carloscaldas:
- Replaced map with foreach for side-effect expressions: 1 warning

- Added type annotation for Unit and public methods: 7 warnings

- Replaced null initialization with scala syntax underscore: 2 warnings

- Removed redundant get when getting a value from a map: 8 warnings

- Simplified complex filter expressions: 2 warnings

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/576)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/BufferedRideHailRequests.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/BufferedRideHailRequests.scala
@@ -16,7 +16,7 @@ class BufferedRideHailRequests(
   private var tick: Double = _
   private var triggerId: Long = _
 
-  def newTimeout(tick: Double, triggerId: Long) = {
+  def newTimeout(tick: Double, triggerId: Long): Unit = {
     this.tick = tick
     this.triggerId = triggerId
   }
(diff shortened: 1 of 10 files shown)


src/main/scala/beam/agentsim/agents/ridehail/BufferedRideHailRequests.scala
```diff
@@ -16,7 +16,7 @@ class BufferedRideHailRequests(
   private var tick: Double = _
   private var triggerId: Long = _
 
-  def newTimeout(tick: Double, triggerId: Long) = {
+  def newTimeout(tick: Double, triggerId: Long): Unit = {
```
JustinPihony: Note that these annotations aren't exactly required, and aren't deep so probably not huge - however this adds stress to the compiler. Every specific annotation results in faster and faster compiler speed.

src/main/scala/beam/agentsim/agents/ridehail/OutOfServiceVehicleManager.scala
```diff
@@ -61,7 +61,7 @@ class OutOfServiceVehicleManager(
       .rideHailAgent
 
     rideHailAgent.tell(
-      ModifyPassengerSchedule(passengerSchedules.get(vehicleId).get),
+      ModifyPassengerSchedule(passengerSchedules(vehicleId)),
```
JustinPihony: I'd almost rather have the explicit get as it should be a flag for refactoring - where an Option might be wanted here?

PR #581, opened 2018-09-14
improve logging fixed #580

### LINKED ISSUE #580: Improve logging to track run source (commit/config)
->  we should print config file path, which is started as log.info 
->  we see in log file on server which commit is used

zishanbilal:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/581)


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -242,7 +243,7 @@ trait BeamHelper extends LazyLogging {
       case _ =>
         val (_, outputDirectory) = runBeamWithConfig(config)
         val props = new Properties()
-        props.setProperty("commitHash", LoggingUtil.getCommitHash)
+        props.setProperty("commitHash", getCommitHash)
```
JustinPihony: This is a bit of personal preference, but I would rather have this be `BashUtil.getCommitHash` as it makes it easier to understand that this logic is being provided elsewhere.

src/main/java/beam/utils/BashUtil.java
```diff
@@ -0,0 +1,43 @@
+package beam.utils;
```
JustinPihony: Any reason to make this a java file, and not scala?

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -337,10 +338,16 @@ trait BeamHelper extends LazyLogging {
       beamConfig.beam.agentsim.simulationName,
       beamConfig.beam.outputs.addTimestampToOutputDirectory
     )
+
     LoggingUtil.createFileLogger(outputDirectory)
     matsimConfig.controler.setOutputDirectory(outputDirectory)
     matsimConfig.controler().setWritePlansInterval(beamConfig.beam.outputs.writePlansInterval)
 
+    logger.info(s"Starting beam on branch $getBranch at commit $getCommitHash.")
```
JustinPihony: While this is only done once, to remain consistent with the newer, more performant standard could you use `{}`:

`logger.info("Starting beam on branch {} at commit {}.", getBranch, getCommitHash)`

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -337,10 +338,16 @@ trait BeamHelper extends LazyLogging {
...
+    val outConf = Paths.get(outputDirectory, "beam.conf")
+    Files.write(outConf, config.root().render(ConfigRenderOptions.concise()).getBytes)
+    logger.info(s"Config [${beamConfig.beam.agentsim.simulationName}] copied to $outConf.")
```
JustinPihony: What is the point of this? Just makes it easier to debug what was actually configured for that run?

zishanbilal: to log them in output that is published to s3 for future analyses.

PR #584, opened 2018-09-14
Cleanup code: removed semicolon, removed return, negative condition a…

carloscaldas:
Improvements:
- Simplified map get: 2
- Simplified filter and size -> count: 1
- Replace size > 0 to nonEmpty: 1
- Removed redundant throws clause: 4
- Removed superfluous braces: 1
- Removed unecessary parenthesis: 3
- Removed superfluous return: 1
- Removed scala semicolon: 8

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/584)

### DIFF
--- a/src/test/scala/beam/agentsim/agents/choice/logit/MultinomialLogitSpec.scala
+++ b/src/test/scala/beam/agentsim/agents/choice/logit/MultinomialLogitSpec.scala
@@ -39,7 +39,7 @@ class MultinomialLogitSpec extends WordSpecLike with Matchers with BeamHelper {
 
       val samps = for (i <- 1 until 100) yield mnl.sampleAlternative(alts, rand).get
 
-      samps.filter(_.equals("walk")).length > 50 should be(true)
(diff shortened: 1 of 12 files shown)


src/test/scala/beam/agentsim/agents/choice/logit/MultinomialLogitSpec.scala
```diff
@@ -39,7 +39,7 @@ class MultinomialLogitSpec extends WordSpecLike with Matchers with BeamHelper {
 
       val samps = for (i <- 1 until 100) yield mnl.sampleAlternative(alts, rand).get
 
-      samps.filter(_.equals("walk")).length > 50 should be(true)
+      samps.count(_.equals("walk")) > 50 should be(true)
```
JustinPihony: For a future reference, you could say should be > 50 - http://www.scalatest.org/user_guide/using_matchers#greaterAndLessThan

PR #587, opened 2018-09-15
Simplification and removed warnings

carloscaldas:
Removed redundant conversion: 1
Removed redundant cast: 1
Removed redudant braces: 3
Removed redundant enclosing block: 5
Removed redundant val in case classes: 3
Simplification: filter and exists: 4
Simplification: nonEmpty: 2

### DIFF
--- a/src/main/scala/beam/calibration/utils/DeleteSuggestion.scala
+++ b/src/main/scala/beam/calibration/utils/DeleteSuggestion.scala
@@ -23,9 +23,8 @@ object DeleteSuggestion extends LazyLogging {
     BeamSigoptTuner.fetchExperiment(experimentId) match {
       case Some(_experiment) =>
         _experiment.suggestions().delete(suggestionId).call()
-      case None => {
+      case None =>
         logger.info(s"Experiment with id $experimentId not found")
-      }
     }
   }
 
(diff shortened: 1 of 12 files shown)


src/main/scala/beam/calibration/utils/DeleteSuggestion.scala
```diff
@@ -23,9 +23,8 @@ object DeleteSuggestion extends LazyLogging {
     BeamSigoptTuner.fetchExperiment(experimentId) match {
       case Some(_experiment) =>
         _experiment.suggestions().delete(suggestionId).call()
-      case None => {
+      case None =>
```
JustinPihony: This one is debatable - in fact it begs the question as to whether we should require brackets for anything that is not on a single line? I've seen subtle bugs where somebody adds something after the fact and their mental bracket expectations do not match the reality.

src/main/scala/beam/router/BeamRouter.scala
```diff
@@ -252,7 +252,7 @@ class BeamRouter(
   private def logExcessiveOutstandingWorkAndClearIfEnabledAndOver = Future {
     val currentTime = getCurrentTime
     outstandingWorkIdToTimeSent.collect {
-      case (workId: WorkId, timeSent: TimeSent) => {
+      case (workId: WorkId, timeSent: TimeSent) =>
```
JustinPihony: Same argument as above wrt brackets

PR #588, opened 2018-09-15
Cluster: Fix cluster init

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/588)

### DIFF
--- a/src/main/scala/beam/sim/BeamHelper.scala
+++ b/src/main/scala/beam/sim/BeamHelper.scala
@@ -2,6 +2,7 @@ package beam.sim
 
 import java.io.FileOutputStream
 import java.nio.file.{Files, Paths}
+import java.util
 import java.util.Properties
 
 import beam.agentsim.agents.ridehail.RideHailSurgePricingManager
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -140,7 +141,19 @@ trait BeamHelper extends LazyLogging {
             )
           ) ++ {
             if (parsedArgs.useCluster)
-              Map("beam.cluster.clusterType" -> parsedArgs.clusterType.get.toString)
+              Map(
+                "beam.cluster.clusterType" -> parsedArgs.clusterType.get.toString,
+                "akka.actor.provider"      -> "akka.cluster.ClusterActorRefProvider",
+                "akka.remote.artery.canonical.hostname" -> parsedArgs.nodeHost.getOrElse(
+                  throw new RuntimeException("`nodeHost` is not set")
```
JustinPihony: It probably doesn't hurt to have this as a backup. Note that it isn't necessary though - the args parser handles this already - Line 101. Same for nodePort and seedAddress

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -140,7 +141,19 @@ trait BeamHelper extends LazyLogging {
...
+                ),
+                "akka.remote.artery.canonical.port" -> parsedArgs.nodePort
+                  .map { _.toInt }
```
JustinPihony: Should we do this in the args parser itself and change the type to an int to start? Otherwise (or in conjunction) should we do something more like a flatMap{x=>Try{x.toInt}.toOption} to handle a non-integer?

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -140,7 +141,19 @@ trait BeamHelper extends LazyLogging {
...
+                  .getOrElse(throw new RuntimeException("`nodePort` is not set")),
+                "akka.cluster.seed-nodes" -> util.Arrays.asList(
```
JustinPihony: Thinking about this more in depth now - we need to handle multiple addresses. I don't believe this handles that currently. Again - probably something for the parser itself, though

src/main/scala/beam/sim/BeamHelper.scala
REASY: Changed type to the string and it works, no issue, going to fix it

src/main/scala/beam/sim/BeamHelper.scala
REASY: We leave is at is,but in the future, I think with Kubernetes we should not provide seed-nodes as args, but rather use discovery to get seed node.

JustinPihony: Agreed. You choose one or the other. Since we're moving to discovery, then no need to spend much time on this now anyway - especially since this is not long lived. Of course this would affect recovery if it can't find an appropriate node - but again not an issue for our current demands and future plans

PR #593, opened 2018-09-16
R5RoutingWorker in remote mode should initialize transits in constructor

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/593)


JustinPihony:
I have to look at this a bit deeper, but at first glance it does something I didn't want to do - introduce the concept of explicit, remote specific code paths. Why not just pass the map of vehicles in the constructor? It can be empty by default, and only if we read from config is it passed in currently. Again, I'll look a bit deeper later, but that's my initial thoughts.

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -1240,7 +1245,8 @@ object R5RoutingWorker {
           network,
           fareCalculator,
           tollCalculator,
-          transitVehicles
+          transitVehicles,
+          isRemote = false
```
JustinPihony: Instead of isRemote, have this be an empty Map.

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -170,13 +169,44 @@ class R5RoutingWorker(workerParams: WorkerParameters) extends Actor with ActorLo
 
   var workAssigner: ActorRef = context.parent
 
+  private var maybeTravelTime: Option[TravelTime] = None
+
+  private var transitSchedule: Map[Id[BeamVehicle], (RouteInfo, Seq[BeamLeg])] = initTransitMap(isRemote)
+
+  def initTransitMap(isRemote: Boolean): Map[Id[BeamVehicle], (RouteInfo, Seq[BeamLeg])] = {
```
JustinPihony: Move this inside the config constructor. Then pass the inited map into the WorkerParameters

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -144,7 +142,8 @@ class R5RoutingWorker(workerParams: WorkerParameters) extends Actor with ActorLo
     network,
     fareCalculator,
     tollCalculator,
-    transitVehicles
+    transitVehicles,
+    isRemote
```
JustinPihony: extract the provided map into the transitSchedule variable.

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -59,15 +53,15 @@ import scala.collection.mutable.ArrayBuffer
...
 
 case class WorkerParameters(
   beamServices: BeamServices,
   transportNetwork: TransportNetwork,
   network: Network,
   fareCalculator: FareCalculator,
   tollCalculator: TollCalculator,
-  transitVehicles: Vehicles
+  transitVehicles: Vehicles,
+  isRemote: Boolean
```
JustinPihony: Add the map instead of isRemote

src/main/scala/beam/router/BeamRouter.scala
```diff
@@ -176,6 +163,11 @@ class BeamRouter(
       }
     case GetMatSimNetwork =>
       sender ! MATSimNetwork(network)
+    case GetTravelTime =>
```
JustinPihony: Why move this from R5Routing to here?

REASY: I've moved it out because I don't want us to send this data across the network. BeamRouter sits clother to simulation (the same JVM), so if it will be sent there, it's just a reference.

PR #607, opened 2018-09-18
Project: Memory optimizations

REASY:
# Optimal way to write string to the buffer:
## Before:
image

## After:
image

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/607)

### DIFF
--- a/src/main/java/beam/utils/BeamCalcLinkStats.java
+++ b/src/main/java/beam/utils/BeamCalcLinkStats.java
@@ -160,43 +160,33 @@ public class BeamCalcLinkStats {
                         Link link = this.network.getLinks().get(linkId);
 
                         out.write(linkId.toString());
-                        // out.write("," + link.getFromNode().getId().toString());
                         writeCommaAndStr(out, link.getFromNode().getId().toString());
 
-                        // out.write("," + link.getToNode().getId().toString());
                         writeCommaAndStr(out, link.getToNode().getId().toString());
 
                         //WRITE HOUR
                         if (i < this.nofHours) {
-                            // out.write("," + Double.toString(i));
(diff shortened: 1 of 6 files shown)


REASY:
# [DeadHeadingStats.java](src/main/java/beam/analysis/plots/DeadHeadingStats.java) can be improved by passing attributes with event:
## Before
image
image

## After:
image

src/main/java/beam/utils/BeamCalcLinkStats.java
```diff
@@ -160,35 +160,52 @@ public void writeFile(final String filename) {
                         Link link = this.network.getLinks().get(linkId);
 
                         out.write(linkId.toString());
-                        out.write("," + link.getFromNode().getId().toString());
-                        out.write("," + link.getToNode().getId().toString());
+                        // out.write("," + link.getFromNode().getId().toString());
```
JustinPihony: Nit: I'd say remove all the comments

src/main/java/beam/utils/BeamCalcLinkStats.java
```diff
@@ -160,35 +160,52 @@ public void writeFile(final String filename) {
...
+                        writeCommaAndStr(out, link.getFromNode().getId().toString());
```
JustinPihony: What are your thoughts on removing the `out` via an implicit? Maybe hiding too much and would make it harder to debug for others?

REASY: @JustinPihony , as far as I know there is no `implicits` in Java. Or you want to extract this to Scala method with `implicit` parameter?

PR #608, opened 2018-09-18
Cleanup code - removed 20 warnings

carloscaldas:
Simplified expression and map operations: 4
Simplified option expression: 1
Replace size with length for array: 1
Replaced collections size with empty/nonEmpty: 2 warning removed
Removed unecessary parenthesis and redundant enclosing block: 12

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
@@ -749,7 +749,7 @@ class RideHailManager(
       }
 
     case DepotParkingInquiryResponse(None, requestId) =>
-      val vehId = parkingInquiryCache.get(requestId).get.vehicleId
+      val vehId = parkingInquiryCache(requestId).vehicleId
       log.debug(
         "No parking stall found, ride hail vehicle {} stranded",
         vehId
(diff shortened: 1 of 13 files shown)


src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -895,7 +895,7 @@ class RideHailManager(
     } else if (outOfServiceRideHailVehicles.contains(vehicleId)) {
       OutOfService
     } else {
-      log.error(s"Vehicle ${vehicleId} does not have a service status, assuming out of service")
+      log.error(s"Vehicle $vehicleId does not have a service status, assuming out of service")
```
JustinPihony: Might as well go to formatting vs interpolation if your improving things

src/main/scala/beam/utils/matsim_conversion/MatsimConversionTool.scala
```diff
@@ -17,7 +17,7 @@ import scala.collection.JavaConverters._
 
 object MatsimConversionTool extends App {
 
-  if (null != args && args.size > 0) {
+  if (null != args && args.length > 0) {
```
JustinPihony: Why not nonEmpty here as you did elsewhere? It'd be more efficient, too.

carloscaldas:
test!

PR #616, opened 2018-09-20
resolve why buses,walk,subway etc. not present in energy_use graph

rajnikantsh:
However the graph is generated for buses, subway but car, ridehail are not generating because code was manipulated by some one . walk data is coming into csv but it is not showing in graph because its value is very small in comparison to other modes.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/616)


JustinPihony:
Can this branch be deleted @rajnikantsh ?

PR #625, opened 2018-09-22
agents: warnings, code cleanup and logger fixed #624

### LINKED ISSUE #624: fix: agents warnings and code cleanup

zishanbilal:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/625)


JustinPihony:
Is there a reason for moving code blocks around? These are non-functional changes which make reviewing more difficult. I'm just curious if there is a reason behind it?

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -323,20 +248,22 @@ object HouseholdActor {
     override def receive: Receive = {
 
       case NotifyVehicleResourceIdle(
-          vehId: Id[BeamVehicle],
```
JustinPihony: Why the change here? An exhaustive warning? If so then my question is that this just hides the warning. Is there a scenario where the cast would fail?

JustinPihony: Same for all the other areas of the same type below

PR #644, opened 2018-09-25
TravelTime: BeamPath contains links' travel time

REASY:
- Add `linkTravelTime` to `BeamPath`
- Added `processLinkEvents` to `PersonAgent` which creates `LinkLeaveEvent` and `LinkEnterEvent` events

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/644)


src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -203,6 +203,43 @@ class PersonAgent(
     else Right(_experiencedBeamPlan.activities(ind))
   }
 
+  def getAverageTravelTime(linkTravelTime: IndexedSeq[Int]):IndexedSeq[Int] = {
+    if (linkTravelTime.isEmpty) Vector()
+    else {
+      // FIXME There are the way to make it faster and even lazy
+      // We don't need the travel time for the last link, so we drop it (dropRight(1))
```
colinsheppard: If I looked at a method "getAverageTravelTime" I would assume that I'd get the travel times for all links, not with the last one dropped. If the caller of getAverageTravelTime doesn't need the last time, then the *caller* should do the dropping, not the method.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -285,6 +286,7 @@ trait ChoosesMode {
         theRouterResult @ RoutingResponse(_, _, Some(requestId)),
         choosesModeData: ChoosesModeData
         ) if choosesModeData.rideHail2TransitRoutingRequestId.contains(requestId) =>
+      processLinkEvents(theRouterResult.itineraries)
```
colinsheppard: Why are we processing link events inside ChoosesMode? ChooseMode is only about selecting a trip plan, it doesn't mean that the travel has occurred. E.g. things can go wrong (like missing the bus) such that certain legs on a trip never happen. So throwing LinkEnter/Exit events will be wrong.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -341,6 +343,7 @@ trait ChoosesMode {
         }
       stay() using newPersonData
     case Event(theRouterResult: RoutingResponse, choosesModeData: ChoosesModeData) =>
+      processLinkEvents(theRouterResult.itineraries)
```
colinsheppard: Same as above, this should not happen in ChoosesMode.

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -182,6 +176,10 @@ trait ChoosesParking extends {
         responses: (RoutingResponse, RoutingResponse),
         data @ BasePersonData(_, _, _, _, _, _, _, _, _)
         ) =>
+      val (r1, r2) = responses
```
colinsheppard: Similar to ChoosesMode, ChoosesParking is mostly a planning class and I don't think we want to throw movement events here.  It should all happen in DrivesVehicle at the same time PathTraversal is processed. This is the true point when we know a movement through links has been made.

REASY:
PNG stats when run on sf-light-5k:
[master branch with a28b96aa1aecd8f91aeb3b558365aa6d4e522dba](https://github.com/LBNL-UCB-STI/beam/commit/a28b96aa1aecd8f91aeb3b558365aa6d4e522dba) => https://drive.google.com/open?id=1hjBKn6Uzk2qnBIFnkZNDIh_s6E9VR_Ey
[this branch](https://github.com/LBNL-UCB-STI/beam/commit/28dad24407869bf8951599a0a7ebef6a2cbaf262) => https://drive.google.com/open?id=1JQRu3BZ3JWd10E5Nwmh8W6ZweZ9hjfPE

colinsheppard:
looks good, thank you!

src/main/scala/beam/agentsim/agents/PersonAgent.scala
michaz: Hmm, but why are we averaging in the first place?

PR #682, opened 2018-10-01
fallback case of plans fixed

### LINKED ISSUE #659: fallback to latest iteration plan in case of missing root level plan
update code to look for a plans file in iterations folder as backup in case ouput_plans missing at root level
 e.g. search in it20
 search in it19

zishanbilal:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/682)


src/main/scala/beam/sim/BeamMobsim.scala
```diff
@@ -205,13 +204,13 @@ class BeamMobsim @Inject()(
...
           beamServices.beamConfig.beam.agentsim.numAgents.toDouble * beamServices.beamConfig.beam.agentsim.agents.rideHail.numDriversAsFractionOfPopulation
         )
-        private val rideHailVehicleType =
-          BeamVehicleUtils
-            .getVehicleTypeById(
-              beamServices.beamConfig.beam.agentsim.agents.rideHail.vehicleTypeId,
-              scenario.getVehicles.getVehicleTypes
-            )
-            .getOrElse(scenario.getVehicles.getVehicleTypes.get(Id.create("1", classOf[VehicleType])))
+//        private val rideHailVehicleType =
```
JustinPihony: I'll have to check by pulling into an editor instead of using the github.com reviewer....but what is accomplished by commenting the val? The code will still execute and now just not be assigned to a variable. Should this be the whole code block?

PR #683, opened 2018-10-01
simplified expression

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/683)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
@@ -63,8 +63,7 @@ class EVFleetAllocationManager(val rideHailManager: RideHailManager)
               .remainingRangeInM < routingResponses.get
               .map(
                 _.itineraries
-                  .filter(_.tripClassifier == RIDE_HAIL)
-                  .headOption
+                  .find(_.tripClassifier == RIDE_HAIL)
                   .map(_.beamLegs().map(_.travelPath.distanceInM).sum)
                   .getOrElse(Double.MaxValue)
               )


carloscaldas:
test!

src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
```diff
@@ -63,8 +63,7 @@ class EVFleetAllocationManager(val rideHailManager: RideHailManager)
               .remainingRangeInM < routingResponses.get
               .map(
                 _.itineraries
-                  .filter(_.tripClassifier == RIDE_HAIL)
-                  .headOption
+                  .find(_.tripClassifier == RIDE_HAIL)
                   .map(_.beamLegs().map(_.travelPath.distanceInM).sum)
```
zishanbilal: use fold instead of map and gotOrElse

src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
```diff
@@ -63,8 +63,7 @@ class EVFleetAllocationManager(val rideHailManager: RideHailManager)
               .remainingRangeInM < routingResponses.get
```
zishanbilal: unchecked access from Option could potentially break. shouldn't use getOrElse.

JustinPihony: I believe you meant - should use getOrElse instead of should NOT use...

src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
carloscaldas: disagree is a good choice. Sum is clearer and the code is already too confuse.

src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
carloscaldas: nice catch.

JustinPihony: Should this be closed out then?

PR #738, opened 2018-10-06
Parking: Lazy logging for `ZonalParkingManager`

REASY:
We log very often and even when log level is not DEBUG, we still need [to calculate argument](https://github.com/akka/akka/blob/master/akka-actor/src/main/scala/akka/event/Logging.scala#L1243) (it's call by value, not call by name)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/738)


REASY:
Test!

REASY:
Test!

REASY:
Test!

PR #739, opened 2018-10-06
EventsManager: Fix high thread contention

REASY:
Fix high thread contention via introducing blocking queue which is performant for our case => many writers (agents which are actors) and one reader (EventsManager)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/739)


REASY:
Before fix:
image

REASY:
After:
image

src/main/scala/beam/sim/LoggingParallelEventsManager.scala
```diff
@@ -56,4 +72,53 @@ class LoggingParallelEventsManager @Inject()(config: Config) extends EventsManag
...
+      val event = nonBlockingQueue.poll()
+      if (null != event)
+        eventManager.processEvent(event)
+    }
+    // We have to consumer the whole queue
+    var isDone = false
+    val start = System.currentTimeMillis()
+    while (!isDone) {
+      val event = nonBlockingQueue.poll()
+      if (null != event)
```
carloscaldas: Consider using positive condition
```Scala
if (null = event) {
  isDone = true
} else {
  eventManager.processEvent(event)
}
```

src/main/scala/beam/sim/LoggingParallelEventsManager.scala
```diff
@@ -9,17 +10,23 @@ import org.matsim.core.config.Config
...
 import org.matsim.core.events.handler.EventHandler
 
+import scala.concurrent.{Await, ExecutionContext, Future}
+import scala.concurrent.duration._
+
 class LoggingParallelEventsManager @Inject()(config: Config) extends EventsManager with LazyLogging {
   private val eventManager = new ParallelEventsManagerImpl(config.parallelEventHandling().getNumberOfThreads())
   private val numOfEvents: AtomicInteger = new AtomicInteger(0)
-  val threshold: Int = 1000000
   logger.info(s"Created ParallelEventsManagerImpl with hashcode: ${eventManager.hashCode()}")
```
carloscaldas: Consider putting the log close to the event:

Option 1:
```Scala
private val eventManager = {
    val result = new ParallelEventsManagerImpl(config.parallelEventHandling().getNumberOfThreads())
    logger.info(s"Created ParallelEventsManagerImpl with hashcode: ${result.hashCode()}")
    result
}

private val numOfEvents: AtomicInteger = new AtomicInteger(0)
val threshold: Int = 1000000
```

Option2: 
```Scala
private val eventManager = new ParallelEventsManagerImpl(config.parallelEventHandling().getNumberOfThreads())
logger.info(s"Created ParallelEventsManagerImpl with hashcode: ${result.hashCode()}")

private val numOfEvents: AtomicInteger = new AtomicInteger(0)
val threshold: Int = 1000000
```

src/main/scala/beam/sim/LoggingParallelEventsManager.scala
```diff
@@ -56,4 +72,53 @@ class LoggingParallelEventsManager @Inject()(config: Config) extends EventsManag
...
+  private def handleBlocking(): Unit = {
+    while (!isFinished.get()) {
+      val event = blockingQueue.poll(1, TimeUnit.SECONDS)
+      if (null != event)
+        eventManager.processEvent(event)
+    }
+    // We have to consumer the whole queue
+    var isDone = false
+    val start = System.currentTimeMillis()
+    while (!isDone) {
```
carloscaldas: ⚠️ Duplicated code
The lines `95-101` and `115-121` are duplicated. Consider extracting a method.

src/main/scala/beam/sim/LoggingParallelEventsManager.scala
```diff
@@ -56,4 +72,53 @@ class LoggingParallelEventsManager @Inject()(config: Config) extends EventsManag
...
+    Future {
+      // Let's go for now with blocking approach because it's safe CPU and pretty good for
+      // the scenario when it's many writers but only one reader
+      handleBlocking()
+    }(ec)
+  }
+
+  private def handleNonBlocking(): Unit = {
+    // This is going to waste some CPU
+    while (!isFinished.get()) {
```
carloscaldas: Consider extract the block of code to a method called `wasteCpu` in order to make the comments unecessary

REASY:
In our case with a lot of writers and one reader, but with one constraint - we want to have fast write. 
I think it worth it to run prod master + 2 workers to see the time if we would use [ConcurrentLinkedQueue](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ConcurrentLinkedQueue.html). 

Currently I'm using [LinkedBlockingQueue](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/LinkedBlockingQueue.html#offer-E-). An offer methods uses ReentrantLock. I'm not sure how good is ReentrantLock in case of many writes and one read.

Issue #757, opened 2018-10-09
only log ever nth warning


wrashid:
From time to time some simulations generate a huge amount of warnings and this is not desired. Consider logging only at n(th) time the message happen following a exponential function. For example, when the same event happens 16 times it should be logged only 5 times: 1, 2, 4, 8, 16

Example of message: Missed transit pickup during a ride_hail_transit trip

wrashid:
-> also add exponential option.

PR #768, opened 2018-10-10
ExponentialLogger trait - Issue 757

carloscaldas:
Issue https://github.com/LBNL-UCB-STI/beam/issues/757

## Mixing with the trait ExponentialLazyLogging
Example of usages:
```Scala
class SomeClass extends ExponentialLazyLogging {
  logger.error("The value {} is even.", 2)

  logger.error("Some event happened")

  logger.error("Then number {} is even", 2, new IllegalArgumentException("some exception"))
}
```

## Using ExponentialLazyLogging as a helper class
```Scala
class SomeClassMixed extends LazyLogging {
  logger.error("here is the LazyLogging approach")

  // below is the exponential approach
  ExponentialLazyLogging.logger.error("Then number {} is even", 2, new IllegalArgumentException("cannot be negative"))
}
```

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/768)


wrashid:
should we rename it to LogarithmicLazyLogging?

PR #772, opened 2018-10-12
report error if config not found

### LINKED ISSUE #721: if config file not found report that as error, instead of ConfigException
e.g. Caused by: com.typesafe.config.ConfigException$Missing: No configuration setting found for key

zishanbilal:
fixed #721

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/772)

### DIFF
--- a/src/main/scala/beam/sim/BeamHelper.scala
+++ b/src/main/scala/beam/sim/BeamHelper.scala
@@ -237,7 +237,7 @@ trait BeamHelper extends LazyLogging {
     }
     assert(
       !isConfigArgRequired || (isConfigArgRequired && parsedArgs.config.isDefined),
-      "Beam config is a required, must provide a valid configuration file."
+      "Beam config is a required, Please provide a valid configuration file."
     )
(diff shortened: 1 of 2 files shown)


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -237,7 +237,7 @@ trait BeamHelper extends LazyLogging {
     }
     assert(
       !isConfigArgRequired || (isConfigArgRequired && parsedArgs.config.isDefined),
-      "config is a required value, and must yield a valid config."
+      "Beam config is a required, must provide a valid configuration file."
```
JustinPihony: Can you change to "..., please provide a valid..." or "..., you must provide..." or ", a valid configuration file must be provided.".

And in fact, while we are improving it, let's help a bit more and add an example:

> Beam config is required.... For example:
>  ./gradlew :run -PappArgs="['--config', 'test/input/beamville/beam.conf']"

PR #780, opened 2018-10-15
EV: Fixed allocation

REASY:
Run on EV-fleet.conf with 5 iteration:
Only this branch: https://drive.google.com/open?id=1Mm7OCJbdCCQgl6KZ978_aG2jQfFeYWe1
With zonal parking optimization: https://drive.google.com/open?id=1wjdqxzLon_3ADtMz3OYdqOkx5Xtk0Tz4

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/780)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
@@ -5,6 +5,7 @@ import beam.agentsim.agents.ridehail.{RideHailManager, RideHailRequest}
 import beam.router.BeamRouter.RoutingResponse
 import beam.router.Modes.BeamMode.RIDE_HAIL
 import org.matsim.api.core.v01.Id
+import org.matsim.core.utils.geometry.CoordUtils
 import org.matsim.vehicles.Vehicle
 
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
```diff
@@ -62,16 +60,18 @@ class EVFleetAllocationManager(val rideHailManager: RideHailManager)
         // from future consideration if it fails this check
         val routingResponses = routeResponsesByDriver.get(agentLocation.vehicleId)
 
+        val routedDisanceInM = routingResponses
+          .getOrElse(List[RoutingResponse]())
```
JustinPihony: Why not just make this a fold also?

REASY:
[QuadTree.getDisk](https://github.com/matsim-org/matsim/blob/0.10.x/matsim/src/main/java/org/matsim/core/utils/collections/QuadTree.java#L140) uses `ArrayList`. We can try to change it to LinkedList because basically it's used as append only collection. 
image

src/main/scala/beam/agentsim/agents/ridehail/allocation/EVFleetAllocationManager.scala
colinsheppard: cause I'm not a scala ninja :)

JustinPihony: Fair point! I should word my comments better :)

PR #784, opened 2018-10-15
Pass trave time across network

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/784)


REASY:
Originally we handled `GetTravelTime` message in `R5RoutingWorker`, but I had moved it to BeamRouter. The question here is shall we move it back to `R5RoutingWorker` and start serializing it and sending across the network (for cluster)?

colinsheppard:
@JustinPihony please take lead review on this and if you approve I'm ok with merging.

src/main/scala/beam/router/BeamRouter.scala
```diff
@@ -126,10 +131,29 @@ class BeamRouter(
...
 
+  val kryoSerializer = new KryoSerializer(context.system.asInstanceOf[ExtendedActorSystem])
+
+  private val updateTravelTimeTimeout: Timeout = Timeout(3, TimeUnit.MINUTES)
+
   override def receive: PartialFunction[Any, Unit] = {
     case `tick` =>
       if (isWorkAndNoAvailableWorkers) notifyWorkersOfAvailableWork()
       logExcessiveOutstandingWorkAndClearIfEnabledAndOver
+    case t: TryToSerialize =>
```
JustinPihony: This part seems like unnecessary overhead just for logging purposes. Maybe we could put it behind a flag which defaults to off?

src/main/scala/beam/router/BeamRouter.scala
```diff
@@ -126,10 +131,29 @@ class BeamRouter(
...
+      val byteArray = kryoSerializer.toBinary(t)
+      log.info("TryToSerialize size in bytes: {}, MBytes: {}", byteArray.size, byteArray.size.toDouble / 1024 / 1024)
+    case UpdateTravelTime_v2(map) =>
+      localNodes.foreach { actor =>
+        log.info("Sending UpdateTravelTime_v2 to  {}", actor)
+        actor.ask(UpdateTravelTime_v2(map))(updateTravelTimeTimeout)
```
JustinPihony: Should these be `ask`ed? Maybe they should be `tell`ed and have the error listener (or replayer*) handle making sure they are sent?

*Those two classes need merged really

src/main/java/beam/physsim/jdeqsim/AgentSimToPhysSimPlanConverter.java
```diff
@@ -171,6 +173,14 @@ private void setupActorsAndRunPhysSim(int iterationNumber) {
             }
         }
 
+        // Black magic happens here
```
JustinPihony: `router.tell(new BeamRouter.UpdateTravelTime(travelTimeCalculator.getLinkTravelTimes()), ActorRef.noSender());` is still happening below - do we want to have both ways - maybe configurable - or local gets the uncompressed version? Also, shouldn't this be moved down to that same end of method location?

src/main/scala/beam/router/BeamRouter.scala
REASY: I left this here to see what is the size of payload. In case if it's not able to serialize it due to the limitation of message size. I'll add some flag to do it only on debug level

src/main/java/beam/physsim/jdeqsim/AgentSimToPhysSimPlanConverter.java
REASY: It's a good idea to keep both ways (optimal for local router), I'll make changes. Regarding to moving `router.tell(new BeamRouter.UpdateTravelTime_v2(map), ActorRef.noSender());`, I left it there because `linkStatsGraph.notifyIterationEnds` will start changing `travelTimeCalculator`, if there were no links

REASY:
`warm start` uses link stats which is average per **hour**, but [TravelTimeCalculator](https://github.com/matsim-org/matsim/blob/master/matsim/src/main/java/org/matsim/core/config/groups/TravelTimeCalculatorConfigGroup.java#L53), by default, is using 15 minutes bins. Is it ok in our case? If yes, maybe we can use averaged per hour, when send travel time calculator across the network (it will safe some memory)

REASY:
@wrashid , I've found this code which basically is what we need: https://github.com/matsim-org/matsim/blob/master/contribs/eventsBasedPTRouter/src/main/java/org/matsim/contrib/eventsBasedPTRouter/SerializableLinkTravelTimes.java
I think we can use it in our case

colinsheppard:
> @wrashid
> `warm start` uses link stats which is average per **hour**, but [TravelTimeCalculator](https://github.com/matsim-org/matsim/blob/master/matsim/src/main/java/org/matsim/core/config/groups/TravelTimeCalculatorConfigGroup.java#L53), by default, is using 15 minutes bins. Is it ok in our case? If yes, maybe we can use averaged per hour, when send travel time calculator across the network (it will safe some memory)

@REASY we want a configurable bin size and yes we should track travel times at the same bin size we use for writing link stats and reading on warm starts. Please help ensure there's 1 configurable parameter that controls this and then make both TravelTimeCalculator and link stats writer use that bin size.

REASY:
Run in cluster mode on this branch: https://drive.google.com/open?id=1Sq_DnH-ktj-VdzN6RJqZCi3gkm78SGas
Run in single JVM on master branch: https://drive.google.com/open?id=1BEm1EF9vM1CBVNXl5ykEfo2l48K1L_nJ

REASY:
Created branch to benchmark route calculation with/without travel time: art/bechmark-for-route-calc-and-travel-time

REASY:
Run in cluster mode with sending travel time to workers becomes slow due to the [issue in Kryo 4.0.0](https://github.com/EsotericSoftware/kryo/issues/558). Once I updated to latest version of Kryo 4, [kryo 4.0.2](https://github.com/EsotericSoftware/kryo/releases/tag/kryo-parent-4.0.2), simulation time became to normal range.
Logs when don't pass travel time: https://drive.google.com/open?id=1hxQ4OsUpPyfuY3aTG46oduSHZbqrDL9v
Logs when pass travel time: https://drive.google.com/open?id=1aTdpceLO3TAIEAQUiOzik-0KgD5I4ix9

REASY:
Test!

REASY:
On my machine all test are passed:
image

PR #801, opened 2018-10-18
refactor lambda, for-loop, StringBuilder and cleanup

zishanbilal:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/801)


src/main/java/beam/utils/DebugLib.java
```diff
@@ -92,7 +92,6 @@ public static String gcAndGetMemoryLogMessage(String message) {
 
     public static void whileTrue() {
```
JustinPihony: AFAICT this isn't used - nor does it make much sense...Maybe just remove?

PR #813, opened 2018-10-21
LinkTravelTimeContainer: Change internal representation

REASY:
Use mutable collection and change format to be `Map[Id[Link], Array[Double]]`
I used https://s3.us-east-2.amazonaws.com/beam-outputs/physsim-0.2-0.2_iter5_warmstart.zip to measure memory and performance.
Before:
image
After:
image

With this optimization it takes around 42 seconds to initialize that map, before it was around 49 seconds.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/813)


src/main/scala/beam/router/LinkTravelTimeContainer.scala
```diff
@@ -10,64 +10,65 @@ import org.matsim.api.core.v01.population.Person
 import org.matsim.core.router.util.TravelTime
 import org.matsim.vehicles.Vehicle
 
-class LinkTravelTimeContainer(fileName: String, timeBinSizeInSeconds: Int) extends TravelTime with LazyLogging {
+import scala.collection.mutable
 
-  private var linkTravelTimeMap: Map[Id[Link], Map[Int, Double]] = Map()
+class LinkTravelTimeContainer(fileName: String, timeBinSizeInSeconds: Int, maxHour: Int = 24) extends TravelTime with LazyLogging {
```
REASY: @wrashid , @colinsheppard Is it ok to have this parameter `maxHour` to pre-allocate array? I can make it configurable.

colinsheppard: @REASY does maxHour imply that no travel time data will have a tick corresponding to any hour >24? Or do the travel time bins wrap around (i.e. get modded by 24)?

@wrashid please confirm desired behavior, but I would imagine that we want the bins to wrap around, and therefore this pre-allocation approach should work. It will, however, become a problem if users want to use BEAM in the context of multi-day simulations and they expect the network travel times to contain bins beyond 24 hours.

REASY: @colinsheppard , `maxHour` here is used as the size of array to be allocated. What I've seen in `0.linkstats.csv` is that it contains hours from 0 to 23 for per link and it seems to me that there is no skip hours. Basically this will be more efficient by memory to store in `Array` instead of `Map`.  If we have more than 24 hours in simulation, we can just make this parameter configurable

REASY:
[BeamCalcLinkStats.java](https://github.com/LBNL-UCB-STI/beam/blob/master/src/main/java/beam/utils/BeamCalcLinkStats.java#L56) uses 24 hours. It also calculates travel time [in the end of each hour](https://github.com/LBNL-UCB-STI/beam/blob/master/src/main/java/beam/utils/BeamCalcLinkStats.java#L83)

REASY:
Facing an issue with making changes in Beam to respect [maxTime from TravelTimeCalculatorConfigGroup](https://github.com/matsim-org/matsim/blob/master/matsim/src/main/java/org/matsim/core/config/groups/TravelTimeCalculatorConfigGroup.java#L54). During Phys Sim it throws an exception because in [VolumesAnalyzer]( https://github.com/matsim-org/matsim/blob/master/matsim/src/main/java/org/matsim/analysis/VolumesAnalyzer.java#L186) it uses 24 hours as constant:
```
Caused by: java.lang.ArrayIndexOutOfBoundsException: 24

	at beam.utils.BeamCalcLinkStats.addData(BeamCalcLinkStats.java:88)
	at beam.analysis.physsim.PhyssimCalcLinkStats.notifyIterationEnds(PhyssimCalcLinkStats.java:84)
	at beam.physsim.jdeqsim.AgentSimToPhysSimPlanConverter.lambda$setupActorsAndRunPhysSim$0(AgentSimToPhysSimPlanConverter.java:175)
	at java.util.concurrent.CompletableFuture$AsyncRun.run(CompletableFuture.java:1626)
	at java.util.concurrent.CompletableFuture$AsyncRun.exec(CompletableFuture.java:1618)
	... 4 common frames omitted
```

REASY:
Test!

REASY:
test!

REASY:
test!

PR #820, opened 2018-10-23
Rw/#790 integrate tolls 4ci

### LINKED ISSUE #790: integrate tolls

colinsheppard:
closes #790

### DIFF
--- a/src/main/scala/beam/agentsim/agents/choice/mode/BridgeTollDefaults.scala
+++ b/src/main/scala/beam/agentsim/agents/choice/mode/BridgeTollDefaults.scala
@@ -44,7 +44,7 @@ object BridgeTollDefaults {
     }
   }
 
-  private def readTollPrices(tollPricesFile: String): Map[Int, Double] = {
+  def readTollPrices(tollPricesFile: String): Map[Int, Double] = {
     if (Files.exists(Paths.get(tollPricesFile))) {
       Source
         .fromFile(tollPricesFile)
--- a/src/main/scala/beam/router/r5/R5RoutingWorker.scala
+++ b/src/main/scala/beam/router/r5/R5RoutingWorker.scala
@@ -33,7 +33,7 @@ import beam.sim.common.{GeoUtils, GeoUtilsImpl}
 import beam.sim.config.{BeamConfig, MatSimBeamConfigBuilder}
 import beam.sim.metrics.{Metrics, MetricsSupport}
 import beam.utils.reflection.ReflectionUtils
-import beam.utils.{DateUtils, FileUtils, LoggingUtil}
+import beam.utils.{DateUtils, DebugLib, FileUtils, LoggingUtil}
 import com.conveyal.r5.api.ProfileResponse
 import com.conveyal.r5.api.util._
 import com.conveyal.r5.profile._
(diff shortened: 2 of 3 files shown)


colinsheppard:
@wrashid I created this PR so we could add some discussion.

Can you confirm the following behavior?

If two subsidies overlap in time (e.g. from hour 1-6 and from hour 4-10) then during the interval of overlap (e.g. from 4-6) both subsidies are applied in an additive manner?

This is what we are expecting based on a recent discussion.

michaz:
I'm not sure adding tolls to the travel times passed to R5 can work. This assumes travel times are only used by R5 as a utility function input -- but they aren't, we have time-dependent routing.

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -1050,14 +1057,24 @@ class R5RoutingWorker(workerParams: WorkerParameters) extends Actor with ActorLo
...
     }
 
+    if (pureTravelTimeWithoutCosts) {
+      travelTime
+    } else {
+      val tollPriceFile = beamServices.beamConfig.beam.agentsim.toll.file
+      val tollPrices = BridgeTollDefaults.readTollPrices(tollPriceFile) // TODO: load tolls file first time only, as soon as logic works
+      val valueOfTravelTimesSavings = 10.0 // 10$/h
+
+      travelTime + tollPrices.getOrElse(linkId, 0.0).asInstanceOf[Double] / valueOfTravelTimesSavings * 3600
```
JustinPihony: This happens to work, but I'd feel better to guarantee ordering by parentheses

    travelTime + ((tollPrices.getOrElse(linkId, 0.0).asInstanceOf[Double] / valueOfTravelTimesSavings) * 3600)

src/main/scala/beam/agentsim/agents/choice/mode/BridgeTollDefaults.scala
```diff
@@ -44,7 +44,7 @@ object BridgeTollDefaults {
     }
   }
 
-  private def readTollPrices(tollPricesFile: String): Map[Int, Double] = {
+  def readTollPrices(tollPricesFile: String): Map[Int, Double] = {
```
JustinPihony: Actually I'm going to agree with @michaz and at least request this be cached in some fashion.

michaz:
What is the intended relation to the existing TollCalculator, by the way? Should that one go away? Should it remain and be another way of _specifying_ link tolls? Then we should probably treat things the TollCalculator says and things the new stuff says the same way, i.e. add them to the same things in the same places.

michaz:
@colinsheppard There is nothing about time-dependent tolls in these commits. They are just reading a table of link ids and amounts.

michaz:
We should rename BridgeTollDefaults. It reads a scenario-specific table, so it does not have anything to do with defaults, even though it comes right after TransitFareDefaults and DrivingCostDefaults, which are actually defaults.

michaz:
BridgeTollDefaults lets us specify tolls by link id. Our network input format is OSM, where we don't have link ids yet. I always thought our link ids were internal.

michaz:
RideHailDefaults, DrivingCostDefaults and TransitFareDefaults are all private to the scoring function: They are used when the router returns embodiedBeamLeg.costEstimate == 0.

With BridgeTollDefaults we are now doing something else, namely feeding it back to a part of the router, even though the router already has a different notion of toll (from TollCalculator), which goes into the costEstimate.

I'm not sure if this can be made consistent. I'll try and figure out how to get tolls into R5, but I'll do this with the old TollCalculator for now, until we can talk about this.

colinsheppard:
@michaz is this PR now moot in favor of #875 ?

PR #843, opened 2018-10-26
Clean out all scala return statements

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/843)


JustinPihony:
I put this as a WIP because I still have to clean https://github.com/LBNL-UCB-STI/beam/blob/master/src/main/scala/beam/router/r5/R5RoutingWorker.scala#L504

It may not be worth it - given how nested that method is currently, though. I'll come back to it tomorrow though.

JustinPihony:
This is ready for review now - The R5RoutingWorker code diff is butchered. The diff is really around `maybeUseVehicleOnEgressTry` and the subsequent match

PR #874, opened 2018-11-01
metrics printer 

### LINKED ISSUE #868: monitor number of routing calculations made
I'd like to know how many times a route is calculated (i.e. how many calls to R5RoutingWorker.getPlanFromR5). Can we collect this? 

Just need this once (for now) so ok to do a hack and then revert.

zishanbilal:
Fixed #868

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/874)


src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -90,6 +91,11 @@ class BeamSim @Inject()(
         }
     }
 
+    beamServices.metricsPrinter = actorSystem.actorOf(MetricsPrinter.props())
+    Await.result(beamServices.metricsPrinter ? Identify(0), timeout.duration)
```
JustinPihony: Can we put this into a Future, then map to subscribe? We could then put off the blocking until the very last second. Or can everything else not currently handle the printer not being set? We should create an issue if so.

src/main/scala/beam/sim/BeamServices.scala
```diff
@@ -75,6 +76,7 @@ class BeamServicesImpl @Inject()(val injector: Injector) extends BeamServices {
   )
 
   var modeChoiceCalculatorFactory: ModeChoiceCalculatorFactory = _
+  var metricsPrinter: ActorRef = _
```
JustinPihony: I realize this is done elsewhere - but it's a bit dangerous. Could we at least force implementers to acknowledge the possibility of emptiness by making this an `Option` at least?

zishanbilal: Moved to BeamSim, printer is not needed in services.

PR #875, opened 2018-11-01
Integrate tolls 4ci

michaz:
- Uses a newly-patched R5 version to make the router toll-aware. We can specify a value of travel time savings (reversed: time value of money) in the routing request. See TollRoutingSpec.
- To implement that, we can now replace the routing weight function (so far, for CAR only), just as we can replace the travel time function.
- Tolls are time dependent: The toll-prices.csv got a new column using the Range data type, previously used in subsidies.
- Because the router _uses_ tolls, it is now also _responsible_ for tolls: The calculated toll (including both the "old" OSM tolls and the "new" link tolls) is summed up in the costEstimate, so the scoring function uses that verbatim.
- Right now, the tuning factor in the config is used only for link tolls, not for OSM tolls (for compatibility).
- Moved the new link tolls into the TollCalculator.
- Removed some dead code from TollCalculator which seemed like provisions for the future, but where I found it obscure which parts of it were already doing something. We can always recover it from the history.


michaz:
@zishanbilal Something about the CI seems to be wrong: The big red X on my commit leads me to failing tests of a totally different commit.

colinsheppard:
@michaz I started a new CI since the last one has a mysteriously failed test:

https://beam-ci.tk/job/beam-4ci/2696/console

Once we're passing and you handle the merge conflicts, feel free to finalize and merge!

PR #886, opened 2018-11-05
Remove TollPriceSpec

michaz:
TollPriceSpec says share of car must go down with rising tolls. With a toll-aware router, that need not be the case; people will route around the toll. (Which is tested in another test.)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/886)


michaz:
For the record, I am merging this despite a failing test, since that seems non-deterministic an reasonably likely not to have anything to do with this.

PR #903, opened 2018-11-10
Fix pt scaling

### LINKED ISSUE #899: Transit Capacity Scaling
Before the refactor from MATSim Vehicle to BeamVehicle, we modified the matsim "transitVehicle" data during initialization to scale the seating capacity of vehicles according to the config parameter:

beam.agentsim.tuning.transitCapacity = 0.1

The scaling happened in BeamSim line 75.

Because the matsim transit vehicles are no longer being used, we need to do the analogous scaling before initializing the transit vehicle fleet in class TransitInitializer.

The reason for doing the scaling in BeamSim was because it should only happen once per model run. If we did it inside TransitInitializer it would happen every iteration. 

That said, we used to mutate the vehicle type data in place, but now the BeamVehicleType is immutable, so there might be a better way to do this inside TransitInitializer where it makes the most logical sense.

sfwatergit:
Closes #899

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/903)


src/main/scala/beam/sim/BeamServices.scala
```diff
@@ -111,6 +111,16 @@ class BeamServicesImpl @Inject()(val injector: Injector) extends BeamServices {
     iterationNumber += 1
     Metrics.iterationNumber = iterationNumber
   }
+
+  def maybeScaleTransit(vehicleTypes: TrieMap[Id[BeamVehicleType], BeamVehicleType]): TrieMap[Id[BeamVehicleType], BeamVehicleType] = {
+    Option(beamConfig.beam.agentsim.tuning.transitCapacity) match {
```
JustinPihony: This won't work as you think, as it's a Double, so even if empty it would be 0.0 - however it comes with a 1.0 default anyway:

    beam.agentsim.tuning.transitCapacity = "double | 1.0"

You're better option would be to change the type in the template to be optional from the start.

PR #915, opened 2018-11-12
Add a perf tip when it comes to immutable vs mutable

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/915)

### DIFF
--- a/docs/developers.rst
+++ b/docs/developers.rst
@@ -400,6 +400,8 @@ Use ``mutable`` buffer instead of ``immutable var``:
    val mutableList = scala.collection.mutable.MutableList(1,2)
    mutableList += 3
    mutableList.toList //returns scala.collection.immutable.List
+                      //or return mutableList but explicitly set the method return type to 
+                      //a common, assumed immutable one from scala.collection (more dangerous)
 
 Don’t create temporary collections, use `view`_:
 ************************************************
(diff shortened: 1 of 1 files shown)


docs/developers.rst
```diff
@@ -394,6 +394,12 @@ Use ``mutable`` buffer instead of ``immutable var``:
    val buffer = scala.collection.mutable.ArrayBuffer.empty[Int]
    buffer += 1
    buffer += 2
+   
+**Additionally note that, for the best performance, use mutable inside of methods, but return an immutable**
+
+   val mutableList = scala.collection.mutable.MutableList(1,2)
+   mutableList += 3
+   mutableList.toList //returns scala.collection.immutable.List
```
REASY: To avoid extra GC pressure due to creating the copy of mutable collection, one can introduce method which returns common type from `scala.collection`, for example:
```
def prep(n: Int): scala.collection.Seq[Int] = {
  val rnd = new scala.util.Random()
  val mutableList = scala.collection.mutable.MutableList[Int]()
  (1 to n).foreach { i =>
    val v = rnd.nextInt()
    mutableList += v
  }
  mutableList
}
```

JustinPihony: Addressed - Unless it becomes an explicit problem I'd rather leave the choice open - since this way is slightly more dangerous since it's still ultimately mutable.

PR #925, opened 2018-11-13
add config param beam output writeGraphs 

rajnikantsh:
add config param for enable and disable graphs

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/925)


test/input/beamville/beam.conf
```diff
@@ -136,7 +136,10 @@ beam.calibration.mode.benchmarkFileLoc = ${beam.inputDirectory}"/calibration/ben
 beam.outputs.baseOutputDirectory = ${PWD}"/output/beamville"
 beam.outputs.baseOutputDirectory = ${?BEAM_OUTPUT}
 beam.outputs.addTimestampToOutputDirectory = true
-
+beam.outputs.writeGraphs = true
```
JustinPihony: Is this even needed? These are the defaults, so there should be no need to add more here.

test/input/beamville/beam.conf
```diff
@@ -283,9 +287,10 @@ matsim.modules {
   }
   controler {
     firstIteration = 0
-    lastIteration = 10
+    lastIteration = 1
```
JustinPihony: Was this a remnant of testing? Should it be changed for pushing in?

rajnikantsh: Hi @JustinPihony, yes it was remnant of testing. I have reverted this. I will make sure this should not happen again

test/input/beamville/beam.conf
rajnikantsh: Hi @JustinPihony , Yes it is needed because we have to generate graph based on this property(default value of this property is true). We can discuss further on this with @wrashid or @colinsheppard

JustinPihony: Yes, but what I was pointing out is that these are the defaults, so don't need set in beam.conf - since they are set in the beamtemplate.

rajnikantsh: Hi @JustinPihony ,Now this property has been removed from beam.conf its exist only in beam-template.conf

PR #941, opened 2018-11-20
Don't stop transit drivers prematurely

### LINKED ISSUE #879: delay stopping of transit drivers?
First task is to determine if this is actually an issue, second task is to fix **if** it's the cause.

Observation: when schedulerParallelismWindow >300, many more people experience errors like  "Person attempted to reserve ride with agent that was not found, message sent to dead letters."

Hypothesis on cause: these drivers are ones that completed their route and have stopped, see this block in TransitDriverAgent:

  when(PassengerScheduleEmpty) {
    case Event(PassengerScheduleEmptyMessage(_), _) =>
      val (_, triggerId) = releaseTickAndTriggerId()
      scheduler ! CompletionNotice(triggerId)
      stop

If they are stopped, then person agents attempting to board the bus "after-the-fact" are getting dead letters and are being stopped as well.

Solution: when a transit drive receives PassengerScheduleEmptyMessage, then he should schedule a PoisonPillTrigger into the future by schedulerParallelismWindow + 1 seconds and at that time take the pill, rather than doing so immediately.

michaz:
Also, test a person with two transit drivers with maximal time window.

Closes #879

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/941)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/TransitDriverAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/TransitDriverAgent.scala
@@ -7,7 +7,11 @@ import beam.agentsim.agents.PersonAgent.{DrivingData, PassengerScheduleEmpty, Ve
 import beam.agentsim.agents.TransitDriverAgent.TransitDriverData
 import beam.agentsim.agents.modalbehaviors.DrivesVehicle
 import beam.agentsim.agents.modalbehaviors.DrivesVehicle.StartLegTrigger
-import beam.agentsim.agents.vehicles.VehicleProtocol.{BecomeDriverOfVehicleSuccess, DriverAlreadyAssigned, NewDriverAlreadyControllingVehicle}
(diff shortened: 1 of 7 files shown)


michaz:
test!

michaz:
When I set the window big enough,
 it could happen that the day would end (scheduler shuts down).
 Even though an agent just replied with new triggers.
 It's difficult to trace, but the previous condition wasn't strong enough in the corner case where there are pending replies (scheduler waits for them),
 and those replies contain new triggers, in particular for the same time step
 then there are enqueued triggers, but no _pending_ triggers,
 and the enqueued triggers are not in the future, but now,
 so we need another chance to send them.
 (Or something.)

michaz:
Perhaps we can rearrange this at some point so it becomes shorter and clearer, but I find it non-obvious how. ,-)

PR #946, opened 2018-11-22
Issue 936 - Gradle task generating documentation

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/946)


carloscaldas:
@colinsheppard the requested changes were already made. please review it

carloscaldas:
@colinsheppard The change you requested was made. Also fixed Malformed table in `outputs.rst`

colinsheppard:
@carloscaldas this is **almost** there. I think the markdown formatting in my previous comment hid what I was asking for. Please make the header above the tables should like like this (i.e. I want the filename to be used as the title (with the word "File: " and the class name to come underneath). This is documentation for users so the classname is almost meaningless to them, the file name is important because this is what they see when they run BEAM:
 
```

File: modeChoice.csv
---------------------

Class name: ModeChosenAnalysisObject


```

colinsheppard:
Also, please rearrange the column order so that field comes first!

carloscaldas:
Hi @colinsheppard. There is a problem with this approach (adding the file as part of the title). Some Objects like `ModeChosenAnalysisObject` have more than one file depending on the field. How should solve this situation?

carloscaldas:
@colinsheppard now is grouped by `file`. In case you want to remove the field `className` from the tables just need to change the line below
https://github.com/LBNL-UCB-STI/beam/blob/13b6db019af0530c69457e476c39eb1ea7eeccaf/src/main/scala/beam/docs/GenerateDocumentationTask.scala#L63

PR #948, opened 2018-11-22
R5RoutingWorker: Override travel time

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/948)


michaz:
Well, how would this be applied in practice, and what's the purpose? If I read it correctly, it tells the router to sometimes put travel times in the response that are different from the ones it optimized for?

colinsheppard:
@michaz the idea is to allow people who have real world travel time data to use that instead of the router's estimate. They aren't interested in modeling congestion / traffic flow, but rather are interested in other features like modeling ride hail operations, etc.

Issue #1002, opened 2018-12-04
HouseholdAttributes, bicycles don't work


michaz:
The code that looks like it is supposed to add bicycles to households can't work. It throws a NullPointerException when it runs, and moreover, it looks like it would add _all_ the bicycles to _all_ the people, but within a for-loop, so once per person.

However, it is never called, because it is called from super-defensive code which does nothing if the person-to-household index isn't there (instead of erroring), see ``PopulationAdjustment``, where the ``householdAttributes`` are created.

For the same reason, all the ``householdAttributes`` are 0.0, always.

All of this should either be made functional and tested, or deleted.

PR #1010, opened 2018-12-05
Shared vehicle fleet

michaz:
---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1010)


michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
Okay, so this can now be played with, but not merged -- Still some non-deterministic things to be investigated, and lots to be cleaned up, but:

- There are now three (kinds of) vehicle managers, all following the same simple protocol:
 - The Households
 - InexhaustibleReservingVehicleFleet
 - FixedNonReservingVehicleFleets

(Reserved doesn't mean "reserved" in the terminology I found in the code, I'm redefining it here:)

Vehicle fleets can be reserving or non-reserving:
- When you get a reserved vehicle through your MobilityStatusResponse, you can behave as before (just start driving with it, nobody will interfere, but if you're not going to use it after all, please return it immediately -- nothing changed here).
- When you get an unreserved vehicle, you have to write a message to the manager before you can start driving. That can fail. Then you have to replan. (You were walking to a car-sharing car but it turns out to be gone.)

Both of this works already, and when you comment in or out the two vehicle fleet constructors in BeamSim, you can watch things happen, even in EventsFileSpec.

Issues left: (partial list)
- Agents still think they have to keep their car until the end of the day and drive home with it. (Change agent behavior.)
- An attempted, but ultimately unsuccessful, shared-car trip which is then replanned apparently still shows up as a car trip in experienced plans, and EventsFileSpec complains that the car tour was not finished by car. I _think_ this is just a bookkeeping issue and the behavior is already correct.
- Clean up the agent. There are now three different state variables for cars, potential cars, later car reservation status and such. It's incomprehensible right now, I'll clean it up. :-)

michaz:
> I didn't notice this before. I appreciate the simplifications to this Actor, but we at a minimum need to track vehicles by "category" and return one vehicle per category if available. This allows the person to know that they have a car AND a bike available and they can then choose between them in ChoosesMode.

Agreed, I'll put this back in. I was just going by my "this is untested and not _currently_ doing anything" heuristic, which is often okay because removing something and putting it back in _can_ be easier than refactoring.

michaz:
Maybe lets [give people bikes](https://github.com/LBNL-UCB-STI/beam/issues/1002), put the vehicle categories back in, and test it?

michaz:
Household vehicles are returned home, other vehicles are not (they are "released" right after a trip).

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

PR #1013, opened 2018-12-05
colors of modes should be fixed and same between realized mode vs. chosen mode

rajnikantsh:
colors are fixed of all modes in all graphs

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1013)

### DIFF
--- a/src/main/java/beam/analysis/plots/GraphUtils.java
+++ b/src/main/java/beam/analysis/plots/GraphUtils.java
@@ -15,6 +15,7 @@ import java.util.*;
 import java.util.List;
 
 public class GraphUtils {
+    private static final Map<String, Color> colorsForModes = new HashMap<>();
     private static final List<Color> colors = new ArrayList<>();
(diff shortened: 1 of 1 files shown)


src/main/java/beam/analysis/plots/GraphUtils.java
```diff
@@ -60,8 +75,14 @@ public static void setColour(JFreeChart chart, int colorCode) {
     public static void plotLegendItems(CategoryPlot plot, List<String> legendItemName, int dataSetRowCount) {
         LegendItemCollection legendItems = new LegendItemCollection();
         for (int i = 0; i < dataSetRowCount; i++) {
-            Color color = getBarAndLegendColor(i);
-            legendItems.add(new LegendItem(legendItemName.get(i), color));
+            String legendName = legendItemName.get(i);
```
wrashid: this is not robost: unknown modes can endup with same color as known modes

PR #1014, opened 2018-12-06
Bl/#940 coupling with energy model task1 4ci

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1014)


colinsheppard:
@bhavyalatha26 please make the following change to the getDirection method (please note the condition for 'R' is a logical OR but the rest are still AND) :

case r if angle < 10  || angle >= 350       => "R" // Right
case sr if angle >=10  & angle < 80       => "SR" // Soft Right
case s if angle >=80  & angle < 100       => "S" // Straight
case sl if angle >=100  & angle < 170       => "SL" // Soft Left
case l if angle >= 170 & angle < 190    => "L" // Left
case hl if angle >= 190 & angle < 270       => "HL" // Hard Left
case hr if angle >= 270 & angle < 350       => "HR" // Hard Right

Please leave the comments that spell these out.

No more need for "NA" we'll just let straight take care of that and deal with stopping separately.

colinsheppard:
Also, the angle computations will happen a lot. So we might as well take one operation away. Can you please delete Math.toDegrees and instead convert all of the angles into radians?

E.g. 10 would become 0.1745329, 250 would become 6.108652, and so on.

Make sure to also change the code that is keeping the angle positive!

bhavyalatha26:
@colinsheppard made the changes , can you please review.

bhavyalatha26:
@colinsheppard reset the below conf to 0
beam.outputs.writePlansInterval
beam.outputs.writeEventsInterval
beam.outputs.writeLinkTraversalInterval

I did not remove them in order to not break any existing use case . Please review

src/main/scala/beam/analysis/LinkTraversalAnalysis.scala
```diff
@@ -0,0 +1,229 @@
...
+              networkLinks.find(x => x.getId.toString.equals(id)) map { currentLink =>
+                val freeFlowSpeed = currentLink.getFreespeed
+                val linkCapacity = currentLink.getCapacity
+                val averageSpeed = try {
+                  currentLink.getLength / travelTime.toInt
+                } catch {
+                  case _: Exception => 0.0
+                }
+                val turnAtLinkEnd =
+                  getDirection(currentLink.getCoord, nextLink.map(_.getCoord).getOrElse(new Coord(0.0, 0.0)))
```
colinsheppard: Oops, I didn't catch this before. This is not the correct math to calculate the turning angle. What this does is effectively calculate the cardinal direction. 

What we want is to pass 2 "coords" which actually represent direction vectors of each link.

So the first argument to "getDirection" should be vectorFromLink(currentLink) and second argument vectorFromLink(nextLink.map(_.getCoord).getOrElse(currentLink))

***note how I used currentLink as the default, this is means that if the nextLink is empty we end up with a Straight instead of some unanticipated turn when you pick an arbitrary 0 coordinate.***

where vectorFromLink(link: Link)

new Coord(link.getToNode.getCoord.getX - link.getFromNode.getCoord.getX, link.getToNode.getCoord.getY - link.getFromNode.getCoord.getY)

colinsheppard:
test!

bhavyalatha26:
all test passed @colinsheppard  , please do a final review and merge

PR #1020, opened 2018-12-07
Rw/#1004 avoid exceeding max plan memory 4ci

wrashid:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1020)


wrashid:
test!

wrashid:
confirmed: the exceeding plan memory size is not an issue, as plans get pruned after each iteration (however plans are written out to file before the pruning)

PR #1036, opened 2018-12-12
Hl/#992 cav hh coordination

htmlrb:
+ Ready to be reviewed.
+ Let me know in case some part code happens to be not in accordance with the coding conventions.
+ The current version of the algorithm can process 50 thousands household, with 5 plans each and 2 vehicles, in 99.73 seconds.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1036)


src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
```diff
@@ -0,0 +1,396 @@
...
+case object MSRInit extends MobilityServiceRequestType
+
+case class MobilityServiceRequest(
+  person: Option[Person],
+  activity: Activity,
+  time: Double,
+  deltaTime: Double,
+  tag: MobilityServiceRequestType
+) {
+  override def toString =
```
JustinPihony: Would it be worth wrapping the string with the type, so it can be easily tracked back?

htmlrb: Through a trait extending an AnyVal ?

JustinPihony: No, I had meant something simpler: 
  
    s"MobilityServiceRequest( $tag{ ${person match {
    ...
    )

But I also realized that the values in `$tag` should provided close enough details

colinsheppard:
test!

PR #1038, opened 2018-12-12
Cs/ParkingRelatedBugFixes

colinsheppard:
Several bug fixes and some other misc.:

* If no car route returned when planning the movement to parking, the driver continues the drive to the destination as if the parking stall is there (even if its not)
* Cut-off for doing parking behavior is not based on radius from destination instead of distance along the route from the end, this prevents persisted routes from degrading more than the behavior threshold
* A couple of wrong CRS issues were resolved
* attribOfIndivid is now sent to parking behavior
* driver Id added to PathTraversal event
* index now used in ZonalParkingManager for all queries
* max search radius now 10km instead of 10,000km
* exploding events options removed as no one has ever used it
* some new R analysis scripts

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1038)


colinsheppard:
test!

colinsheppard:
test!

colinsheppard:
test!

PR #1061, opened 2018-12-18
extract matsim properties from beam.conf to matsim.conf

### LINKED ISSUE #1008: Remove matsim items from regular config.
please extract matsim.conversion and matsim.modules from our standard config files and put them into a matsim.conf file under "../common/".

However, there are just a few params that we actually use regularly from matsim:

all of the ones under matsim.modules.strategy, can we either keep those in the regular beam config even though the rest of matsim.modules is somewhere else? this would be the best outcome.

rajnikantsh:
Fixed #1008

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1061)


colinsheppard:
We still need to create params for all of the typical and useful things we do in BEAM and then set them in the config object during runtime.

E.g. firstIter / lastIter
modules.strategy
physsim.network
etc.

colinsheppard:
@rajnikantsh Please look closely at the changes I made and then apply the same changes to all of the sf-light config files:

* add the include for matsim.conf
* add the beam parameters in the same place I put them in beam.conf

Please test and make sure that the model runs for all of the updated conf files. 

Finally, please confirm for all of these parameters that the values are successfully communicated through to the BeamConfig class at runtime and not just taking on default values.

colinsheppard:
@rajnikantsh 

I've renamed and repositioning many of the params in sf-light-1k.

You can copy these changes to all of the sf-light files but just duplicating from sf-light-1k and replacing 1k withe the appropriate value (and be sure to change beam.agentsim.numAgents to the appropriate value as well).

PR #1070, opened 2018-12-20
Bl/#1035 order fields in events 4ci

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1070)


JustinPihony:
Yes, this could be problematic. I think a simpler fix should be to change from:

````
Set<String> getKeysToWrite(Event event, Map<String, String> eventAttributes) {
        return new HashSet<>(eventAttributes.keySet());
    }
````

to 

````
Set<String> getKeysToWrite(Event event, Map<String, String> eventAttributes) {
        return new LinkedHashSet<>(eventAttributes.keySet());
    }
````

The `keySet` will maintain order of the underlying Map (which happens to be a `LinkedHasMap`), and so `LinkedHashSet` should keep the order as well.

I haven't tried this, but it SHOULD work with that simple change.

JustinPihony:
> I just changed it to iterate over the LinkedHashMap directly (as it already preserves the order) and print the key and value.

This is better than the original implementation actually. I had thought it was a perf thing but was just part of the original.

PR #1071, opened 2018-12-21
Fix memory leak in link analysis

sfwatergit:


JustinPihony:
Should a test be added to make sure we don't revert?

sfwatergit:
The test case here may be too broad. Essentially any child of event should clear any mutable maps upon iteration reset. This is more or less a Java best practice that will be hard to encode as a contract to be fulfilled. Best to just note in review and consider using a dev branch with master only reflecting relatively stable release tags. I think the project is sufficiently mature and public facing to adopt this more principled vcs paradigm

PR #1073, opened 2018-12-21
#1039 create summary plots 4ci

rajnikantsh:
Create Summary Stats plots

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1073)


rajnikantsh:
test!

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -248,6 +255,51 @@ class BeamSim @Inject()(
     out.close()
   }
 
+  def createSummaryStatsGraph(fileName: String, event: IterationEndsEvent): Unit = {
```
JustinPihony: To reduce on reprocessing the same logic, could you expect the parts pulled out of the `event` - since that part won't change even if the file does - controlerIO and iteration.

rajnikantsh: Hi @JustinPihony , controlerIO is not changing with `event` but the iteration will change with the event . I have changed the code as per your comment . Please review it.

JustinPihony: LGTM

PR #1106, opened 2019-01-04
repairing the fallback BeamPath used for transit trips that can't be …

wrashid:
…routed to avoid exceptions later on

(cherry picked from commit 2cd4bdc)


wrashid:
test!

wrashid:
test!

PR #1112, opened 2019-01-05
Hl/#1058 pooling alonso mora

htmlrb:
SkimUtils to be used by pooling and cav

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1112)


src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
```diff
@@ -0,0 +1,195 @@
...
+          Some(plan.getPerson),
+          activity,
+          activity.getEndTime,
+          0.0,
+          MSRPickup
+        )
+    }
+  }
+  requests = requests.sortWith(_.time < _.time)
+  def apply(): List[MobilityServiceRequest] = { requests }
```
JustinPihony: This is an odd use of `apply`, which is usually meant as a de-sugaring method. But here it seems as though it should be more of a property accessor. So maybe it should be renamed to `requests`, so that it can be use akin to `householdplans.requests` - in fact I would just turn the whole thing into one expression which could be a `lazy val requests = for-comprehension`

src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
```diff
@@ -0,0 +1,195 @@
...
+  override def toString = s"${requests}"
+}
+
+class HouseholdCAVScheduling(
+  plans: List[BeamPlan],
+  fleetSize: Int,
+  timeWindow: Double,
+  skim: Map[Coord, Map[Coord, Double]]
+) {
+  private var fleet = List[HouseholdCAV]()
```
JustinPihony: This will be the last comment as I'm now seeing this in multiple places. In general you can avoid var in scala, and in these cases use a mutable collection or a more native expression (mapping) to build your list/mutation

src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
```diff
@@ -0,0 +1,195 @@
...
+    override def toString = {
+      var output = s"Household Schedule - COST:${cost}.\n"
+      cavFleetSchedule.foreach { i =>
+        output += s"${i}\n"
+      }
+      output
+    }
+  }
+
+  def apply(): List[HouseholdSchedule] = {
```
JustinPihony: Possible misuse of apply again?

PR #1151, opened 2019-01-16
Bl/#1079 update ride hail waiting stats 4ci

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1151)


src/main/java/beam/analysis/plots/RideHailWaitingTazAnalysis.java
```diff
@@ -0,0 +1,123 @@
...
+        BufferedWriter out = IOUtils.getBufferedWriter(csvFileName);
+        try {
+            out.write(heading);
+            out.newLine();
+            dataMap.forEach((k,v) -> {
+                DoubleSummaryStatistics stats = v.stream()
+                        .mapToDouble(x -> x)
+                        .summaryStatistics();
+                String line = k.getFirst() + "," + k.getSecond().toString() + "," + stats.getAverage() + "," +
+                        (stats.getMax() + stats.getMin())/2 + "," + 0 + "," + 0 + "," + 0;
```
colinsheppard: This is not the median:

> (stats.getMax() + stats.getMin())/2 

see this article to see how to do it:

http://fruzenshtein.com/scala-median-funciton/

Also, instead of making "numberOfPickups" hard coded 0, can't we just use "stats.size"?

colinsheppard:
test!

bhavyalatha26:
test!

src/main/java/beam/analysis/plots/RideHailWaitingTazAnalysis.java
bhavyalatha26: There is no stats.size , do you the size of the dataMap i.e., the number of entries written into the csv @colinsheppard ?

PR #1169, opened 2019-01-18
Test car sharing; add modes explicitly

michaz:
Moved some modes added ad-hoc in a getter function out of there.
Added them to the initialization of the agent.

(Found this when I was giving people "car" as only available mode, and they would walk anyway.)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1169)


michaz:
Also, put back the parallel EventsManager that I accidentally disabled the other day. @REASY

michaz:
test!

michaz:
test!

PR #1176, opened 2019-01-19
Ma/#1150 msa4r5router 2 4ci

mr-asif-nor:
A fix is made to MSA implementation

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1176)


mr-asif-nor:
I think this error that has come during its build might get resolved when work from @rajnikantsh will get merged into master. He has done a fix in beam.conf regarding planning modules order.

PR #1180, opened 2019-01-20
Configs for stuck detection & automatic stuckAgentDetection config generation

REASY:
I used maximum result from the run for 5 interation `sf-light-25k` for all sf-lights.
##### Could not find easy way to create `beam.debug.stuckAgentDetection` as `HOCON`, not `JSON`.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1180)


colinsheppard:
test!

JustinPihony:
What do you mean by 

> Could not find easy way to create beam.debug.stuckAgentDetection as HOCON, not JSON

Can you provide an example of what you want, versus what is already there...it looks fine to me.

test/input/beamville/beam.conf
```diff
@@ -161,13 +161,87 @@ beam.outputs.stats.binSize = 3600
...
 
 beam.debug.stuckAgentDetection {
-  enabled = false
-  thresholds = []
+  "checkIntervalMs" : 200,
+  "checkMaxNumberOfMessagesEnabled" : false,
+  "defaultTimeoutMs" : 60000,
+  "enabled" : false,
+  "overallSimulationTimeoutMs" : 100000,
+  "thresholds" : [
```
REASY: @JustinPihony,
As you can see, it's `JSON`, not `HOCON`. As `HOCON` it should be:
```
beam.debug.stuckAgentDetection {
  enabled = true
  checkIntervalMs = "1000ms"
  checkMaxNumberOfMessagesEnabled = true
  ....
```
So no `:`, no `"` and a bit more things.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -338,6 +338,14 @@ class BeamAgentScheduler(
         } else {
           log.debug(s"Statistics about trigger: ${System.lineSeparator()} ${triggerMeasurer.getStat}")
         }
+        if (beamConfig.beam.debug.triggerMeasurer.enabled && beamConfig.beam.debug.triggerMeasurer.writeStuckAgentDetectionConfig) {
```
JustinPihony: Shouldn't this be run in the `TriggerMeasurer` itself? It seems like an unnecessary redirection

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -131,7 +131,7 @@ class BeamAgentScheduler(
   private var startSender: ActorRef = _
   private var nowInSeconds: Int = 0
 
-  private val triggerMeasurer: TriggerMeasurer = new TriggerMeasurer
+  private val triggerMeasurer: TriggerMeasurer = new TriggerMeasurer(beamConfig.beam.debug.triggerMeasurer)
```
JustinPihony: While I understand the intent of pushing this into the object itself, and maybe keep it there as a backup - but I'd say why even create the object if it's disabled?

```
    private val triggerMeasurer: Option[TriggerMeasurer] = 
      if(beamConfig.beam.debug.triggerMeasurer.enabled) Some(new TriggerMeasurer(beamConfig.beam.debug.triggerMeasurer)) 
      else None
```

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
JustinPihony: In fact, I'm unsure why this is even in the measurer itself? Shouldn't it be in something else entirely?

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
JustinPihony: In fact if my comment below is taken then the instantiation isn't even needed at all maybe?

test/input/beamville/beam.conf
JustinPihony: I was able to get it to work seemingly - testing it more fully now:

```
beam.debug.stuckAgentDetection{
  checkIntervalMs = 200
  checkMaxNumberOfMessagesEnabled = false
  defaultTimeoutMs = 60000
  enabled = true
  overallSimulationTimeoutMs = 100000
  thresholds = [
    {
      actorTypeToMaxNumberOfMessages {
        population = 1
        rideHailAgent = 1
        transitDriverAgent = 1
      }
      markAsStuckAfterMs = 20000
      triggerType = "beam.agentsim.agents.InitializeTrigger"
    }
    {
      actorTypeToMaxNumberOfMessages {
        population = 7
      }
...
```

src/main/scala/beam/agentsim/scheduler/TriggerMeasurer.scala
```diff
@@ -14,102 +22,178 @@ import scala.collection.mutable.ArrayBuffer
...
+    maxPerActorType
+  }
 
-    maxPerActorType.foreach {
-      case (actorType, triggetTypeToMaxMsgs) =>
-        val s = triggetTypeToMaxMsgs.map { case (key, v) => s"\t\t${key} => $v$nl" }.mkString
-        val str = s"""\t$actorType => ${triggetTypeToMaxMsgs.map { case (_, v) => v }.sum}
-        |$s""".stripMargin
-        sb.append(str)
+  def prepareJsonConfig(maxPerActorType: Map[String, Map[String, Int]]): String = {
```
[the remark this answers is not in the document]
REASY: It's printing config with the values which it gets when simulation is finished. Sometimes we introduce/rename `Trigger` type and it's hard to derive proper way of configuration just from the output of `getStat`. In such cases one can run simulation with turned-off `StuckFinder`, but turned-on `TriggerMeasurer` to get config which is almost ready to use.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
REASY: Originally I put it as a extra step in `getStat` method, but then I thought it's too much things for the method with the name `getStat`. Maybe I can call it `onIterationEnd` or something like that and leave all these logic there? @JustinPihony, what do you think?

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
REASY: Fixed

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
JustinPihony: I see the dilemma...hard choice. Since you already made a choice this way I'd say go with it then :)

REASY:
@JustinPihony , yes, please, if you know the way how to make that config look like HOCON, let's add that. Don't want config look half HOCON half JSON. Thanks.

JustinPihony:
@REASY I updated to HOCON, and also merged the master - LGTM, but I was the last to touch it so it might be better for another to give the final review? - even if it's you I guess :p

REASY:
Thanks, @JustinPihony. Seems like I have misunderstood you. I've thought you will fix the code to make it 
 to generate `HOCON` instead of `JSON`.

REASY:
Test!

PR #1186, opened 2019-01-21
Make shared vehicle fleets configurable with vehicleType

michaz:
Prepare for pricing by making vehicle fleets configurable with vehicleType.
Add more tests.

Next up: Add parameters for monetary cost per meter / per second to vehicleType.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1186)


michaz:
--> Then we have car/bike/scooter/...-sharing with "linear" pricing finished, except that it still needs an input file for the initial vehicle locations.

PR #1193, opened 2019-01-21
 Remove a serious case of dead code

michaz:
This doesn't change any behavior, it just removes unused code that would make the protocol seem more convoluted than it is.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1193)


JustinPihony:
I'll defer to @wrashid or @colinsheppard as this *seems* that it will actually change behavior - as some replies that were previously sent will no longer be there.

JustinPihony:
Should this be cleaned up and moved forward, or dropped? @colinsheppard

PR #1194, opened 2019-01-22
StuckFinder: Increased thresholds and removed warning logs

REASY:
- Increased max number of messages to reduce false-positives
- Do not log when message type is not found (it will happen one time during the creation of `StuckFinder`)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1194)


JustinPihony:
LGTM, but wasn't scalafmt'ed - that's why it failed in CI

PR #1198, opened 2019-01-23
remove default strategies/weights from beam template

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1198)


src/main/resources/beam-template.conf
```diff
@@ -438,44 +438,44 @@ matsim.modules {
     maxAgentPlanMemorySize = 5
     planSelectorForRemoval = "WorstPlanForRemovalSelector"
 
-    ModuleProbability_1 = 0.7
-    Module_1 = "SelectExpBeta"
+    ModuleProbability_1 = 0.0
```
JustinPihony: This does NOT remove the defaults. Should this return "empty" defaults or throw an exception if unset?  @wrashid I know you merged, but I wanted to make sure this is as intended in the issue #1187

PR #1199, opened 2019-01-23
Tests: Set root log level to INFO

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1199)


michaz:
Not permanently, right? But until the being-stuck issue is resolved?

Btw if you change the "pattern" (above) to the commented-out block, you can see the name of the actual actor that is logging something. (I personally like that better than adding the agent id myself, in code, at every place..)

PR #1211, opened 2019-01-25
Bl/#1189 excluded modes for population 4ci

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1211)


src/test/scala/beam/integration/CarSharingSpec.scala
```diff
@@ -105,8 +106,11 @@ class CarSharingSpec extends FlatSpec with Matchers with BeamHelper {
 
     // Only driving allowed
     val population = scenario.getPopulation
+    val nonCarModes = BeamMode.allBeamModes map { mode =>
+      if (mode == BeamMode.CAR) "" else mode.toString.toLowerCase
+    } mkString ","
```
wrashid: this could lead to nonCarModes=",ride_hail,..." => avoid leading/trailing ","

src/main/scala/beam/utils/ScenarioReaderCsv.scala
```diff
@@ -156,7 +159,12 @@ object ScenarioReaderCsv {
         population.getPersonAttributes.putAttribute(person.getId.toString, "householdId", household_id)
         population.getPersonAttributes.putAttribute(person.getId.toString, "rank", 0)
         population.getPersonAttributes.putAttribute(person.getId.toString, "age", age.toInt)
-        population.getPersonAttributes.putAttribute(person.getId.toString, "available-modes", modes)
+        val attributesOfIndividual = PopulationAdjustment
+          .adjustBeamAttributes(population, person.getId.toString, modes.split(","))
+        population.getPersons
```
wrashid: can this additional operation be hidden from caller of module?

bhavyalatha26: Did you mean to club both the steps (fetching attributesOfIndividual & setting it inside beam-attributes) into a single function @wrashid ?

src/test/scala/beam/integration/CarSharingSpec.scala
bhavyalatha26: there will be no trailing or leading coma in this case , as we are making a list as csv string

I choose this because i thought it would be good to fetch all modes from the Beam Mode definitions rather than harcoding it here in the spec as "ride_hail,bus,train......"
@wrashid

bhavyalatha26:
test!

bhavyalatha26:
test!

bhavyalatha26:
@wrashid all checks passed , please do a final review and merge

colinsheppard:
test!

PR #1216, opened 2019-01-26
TollCalculator: Change type for cache and fields type to java.util.Map

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1216)


michaz:
Is this about performance? Of querying, or serialization, or both? (Just curious..)

REASY:
@michaz, performance wise it's only micro-optimization (they say java HashMap is almost twice faster than Scala HashMap for the lookup). I removed `.view` because on the small collections the overhead will be high. 
To store serialized data types, IMHO, java types are more preferable because they don't break backward compatibility like Scala does :)

PR #1230, opened 2019-01-29
plan sub sampler 4ci

zishanbilal:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1230)


src/main/scala/beam/utils/plan/SubSamplerApp.scala
```diff
@@ -0,0 +1,83 @@
...
+import org.matsim.core.scenario.ScenarioUtils
+import org.matsim.core.utils.geometry.transformations.GeotoolsTransformation
+import org.matsim.core.utils.io.UnicodeInputStream
+
+import scala.util.Try
+
+
+object SubSamplerApp extends App {
+
+  def createSample(inputPopulationFile: File, networkFile: File, samplesize: Double, outputPopulationFile: File): Unit = {
```
wrashid: should we pass the whole config file or more files and output also more files, e.g. PopulationAttributes, HouseholdAttributes, populaiton, housholds, vehicles, etc.

zishanbilal: updated

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -696,12 +698,13 @@ class R5RoutingWorker(workerParams: WorkerParameters) extends Actor with ActorLo
               for ((beamLeg, index) <- tripWithFares.trip.legs.zipWithIndex) yield {
                 var cost = tripWithFares.legFares.getOrElse(index, 0.0)
                 val age = routingRequest.attributesOfIndividual.flatMap(_.age)
-                val ids = beamServices.agencyAndRouteByVehicleIds.get(
-                  beamLeg.travelPath.transitStops.fold(vehicle.id)(_.vehicleId)
-                )
-                cost =
-                  ids.fold(cost)(id => beamServices.ptFares.getPtFare(Some(id._1), Some(id._2), age).getOrElse(cost))
-
+                val vehicleId = beamLeg.travelPath.transitStops.fold(vehicle.id)(_.vehicleId)
```
JustinPihony: Just as verification, is the logic here that if it's the same vehicle then the cost will be the same - so save the cycles?

src/main/scala/beam/sim/BeamMobsim.scala
```diff
@@ -122,7 +122,7 @@ class BeamMobsim @Inject()(
           "RideHailManager"
         )
         context.watch(rideHailManager)
-        Await.result(rideHailManager ? Identify(0), timeout.duration)
+//        Await.result(rideHailManager ? Identify(0), timeout.duration)
```
JustinPihony: This is somewhat pertinent to https://github.com/LBNL-UCB-STI/beam/issues/1341 so I'm curious - are we sure that we do not need to wait for the RHM to be inited fully?

zishanbilal: @sfwatergit may comment on it.

JustinPihony:
@colinsheppard Is this still needed, or should it be dropped?

PR #1243, opened 2019-02-04
Cs/#1201 household cav

htmlrb:
+ better tests for coherence checking of the cav algo
+ rewriting the code to fit better with the scala style

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1243)


JustinPihony:
First, it needs a `scalafmtAll` run on it. That's why CI is failing right now

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -288,18 +290,23 @@ class PersonAgent(
...
               // use the mode of the next leg as the new tour mode.
-              currentTourMode = data.currentTourMode.orElse(
-                _experiencedBeamPlan.getPlanElements
-                  .get(_experiencedBeamPlan.getPlanElements.indexOf(nextAct) - 1) match {
-                  case leg: Leg =>
-                    BeamMode.fromString(leg.getMode)
-                  case _ => None
-                }
-              )
+              currentTourMode = modeOfNextLeg match {
```
JustinPihony: This would be better as an `orElse`

    currentTourMode = modeOfNextLeg.orElse(data.currentTourMode)

The final `orElse` shouldn't be necessary, because it would be `None` no matter what since it's already in the None path.

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -135,10 +147,67 @@ object HouseholdActor {
     implicit val pop: org.matsim.api.core.v01.population.Population = population
 
     private var availableVehicles: List[BeamVehicle] = Nil
+    private var cavPlans: List[CAVSchedule] = List()
```
JustinPihony: Can we use a `mutable.List` instead of a mutable reference to an immutable List. This is the worst case for performance. Same for all the other. In fact, I would refactor the existing `availableVehicles` to a `mutable.List` also

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -135,10 +147,67 @@ object HouseholdActor {
...
+                Seq(),
+                transportNetwork,
+                tollCalculator
+              )
+            )
+            context.watch(cavDriverRef)
+            schedulerRef ! ScheduleTrigger(InitializeTrigger(0), cavDriverRef)
+            cav.manager = Some(self)
+          }
+          val householdBeamPlans = household.members.map(person => BeamPlan(person.getSelectedPlan)).toList
```
JustinPihony: You shouldn't need the `toList`?

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -158,68 +227,106 @@ object HouseholdActor {
...
         val vehId = vId.asInstanceOf[Id[BeamVehicle]]
         vehicles(vehId).spaceTime = whenWhere
         log.debug("updated vehicle {} with location {}", vehId, whenWhere)
 
       case ReleaseVehicle(vehicle) =>
-        vehicle.unsetDriver()
-        availableVehicles = vehicle :: availableVehicles
-        log.debug("Vehicle {} is now available for anyone in household {}", vehicle.id, household.getId)
+        if(vehicle.beamVehicleType.automationLevel > 3){
+          //TODO update cav schedule and re-dispatch this vehicle if necessary
```
JustinPihony: Per a recent issue by Colin, I think the preference is to open a github issue instead of adding a TODO in the code. So, can we get rid of this nop, and open an issue to track work on this?

src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
```diff
@@ -1,195 +1,398 @@
...
-              new MobilityServiceRequest(None, request.activity, request.time - 1, newDeltaTime, MSRRelocation)
-            val newRequest = new MobilityServiceRequest(request.person, request.activity, request.time, 0, MSRPickup)
-            newCavSchedule = Some(
-              new CAVSchedule(schedule :+ relocationRequest :+ newRequest, cav, newCost, occupancy + 1)
+    // ***
+    def check(
+      request: MobilityServiceRequest,
+      skim: Map[BeamMode, Map[Coord, Map[Coord, Double]]]
+    ): (List[CAVFleetSchedule], Boolean) = {
+      import scala.collection.mutable.{ListBuffer => MList}
```
JustinPihony: Can you change this to MutableList so that it is more explicit

src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
```diff
@@ -1,195 +1,398 @@
...
+            cav,
+            0
+          )
+        )
+        (req)
+    }
+    // compute all the feasible schedules through
+    val feasibleSchedules = MList[CAVFleetSchedule](new CAVFleetSchedule(emptyFleetSchedule.toList, householdRequests))
+    for (request  <- householdRequests.requests;
+         schedule <- feasibleSchedules) yield {
```
JustinPihony: There is no need to yield here because the end result is not used. It will translate into a nested foreach without the yield, which seems to be the intent

src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
```diff
@@ -1,195 +1,398 @@
...
-    for (request <- householdRequests()) {
-      var householdSchedulesToAdd = List[HouseholdSchedule]()
-      var householdSchedulesToDelete = List[HouseholdSchedule]()
-      for (schedule <- feasibleSchedules) {
-        householdSchedulesToAdd ++= schedule.check(request)
-        if (!schedule.feasible) {
-          householdSchedulesToDelete = householdSchedulesToDelete :+ schedule
-        }
+    for (src <- activitySet;
+         dst <- activitySet) yield {
```
JustinPihony: Same as above - if you are going for mutation, then you don't want a yield (map/flatMap), but instead a foreach (no yield)

src/main/scala/beam/agentsim/agents/planning/BeamPlan.scala
```diff
@@ -41,6 +45,36 @@ object BeamPlan {
     beamPlan.createToursFromMatsimPlan()
     beamPlan
   }
+
+  def addOrReplaceLegBetweenActivities(plan: Plan, leg: Leg, originActivity: Activity, destinationActivity: Activity): Plan = {
+    val newPlanElements = plan.getPlanElements.asScala.sliding(2).map{ elems =>
```
JustinPihony: This isn't dire to change, but pattern matching would resolve some of the casts and clean up the code here.

src/main/scala/beam/agentsim/agents/household/HouseholdCAVScheduling.scala
JustinPihony: Yes, but my point was to make the M more explicit, so the suggestion would change to MutableListBuffer

PR #1246, opened 2019-02-06
path traversal with length zero, but linkIds mentioned have length non zero

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1246)


JustinPihony:
test!

PR #1251, opened 2019-02-07
Cs/new network

colinsheppard:
minor tweaks associated with creating a new network

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1251)

### DIFF
--- a/src/main/scala/beam/router/osm/TollCalculator.scala
+++ b/src/main/scala/beam/router/osm/TollCalculator.scala
@@ -148,7 +148,7 @@ class TollCalculator @Inject()(val config: BeamConfig) extends LazyLogging {
       .flatMap(c => {
         c.split(" ")
           .headOption
-          .flatMap(token => try { Some(token.toDouble) } catch { case _ : Throwable => None })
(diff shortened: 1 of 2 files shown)


src/main/scala/beam/router/osm/TollCalculator.scala
```diff
@@ -148,7 +148,8 @@ class TollCalculator @Inject()(val config: BeamConfig) extends LazyLogging {
       .flatMap(c => {
         c.split(" ")
           .headOption
-          .flatMap(token => Some(Toll(token.toDouble, Range("[:]"))))
+          .flatMap(token => try { Some(token.toDouble) } catch { case _: Throwable => None })
```
JustinPihony: Most likely not worth changing, but interesting to be aware of:

    import scala.util.control.Exception._
    ...
    .flatMap(token => catching(classOf[Throwable]).opt(token.toDouble))
    .map(token => Toll(token, Range("[:]"))))

Note that the last statement can just be a `map` since you're flattening the Option anyway.

PR #1288, opened 2019-02-15
Bl/#1271 step by step guide

### LINKED ISSUE #1271: Update Step-by-Step Guide
I recently updated the "Getting Started" guide in our docs, but it's clear it needs to be updated more. 

In particular, we should move instructions on setting up git LFS to the getting started guide. 

Please begin from this branch (which will be merged to master soon):

cs/bsd-release

Then please use a virtual machine or some other computer you have access to, to actually walk through the process of installing all relevant software (including IDEA) in order to add as much detail as possible to the guide so that people totally new to using IDEA and running BEAM don't experience any issues. 

Then add to the guide under 

docs/users.rst

Note that the LFS config explanation is in:

docs/developers.rst

bhavyalatha26:
closes #1271 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1288)


docs/users.rst
```diff
@@ -55,6 +113,67 @@ The simplest, smallest, and fastest is the beamville scenario (described below).
...
+
+Since BEAM is built with java/scala . A scala sdk module needs to be configured to run BEAM. Check the below steps on how to add a scala module to IDEA
+* Go to File > Project Settings > Global Libraries
+* Click + and select Scala SDK
+* Select the required scala SDK from the list , if no SDK found click Create.
+* Click "Browse" and select the scala home path or click "Download" (choose 2.12.x version)
+
+**Running BEAM from IDE**
+
+BEAM requires some arguments to be specififed during run time like the scenario configuration.
```
JustinPihony: typo - specified

JustinPihony: Also, run-time in this case (note the dash)

docs/users.rst
```diff
@@ -17,11 +17,68 @@ System Requirements
...
+
+It will print out the installed version, and please make sure it is `git-lfs/2.3.4`.
+
+To update the text pointers with the actual contents of files, run the following command (if it requests credentials, use any username and leave the password empty)::
+
+   $ git lfs pull
+   Git LFS: (98 of 123 files) 343.22 MB / 542.18 MB
+
+
+**Installing git lfs on windows :**
```
JustinPihony: I'm not sure it's worth mentioning, but I was having problems with this and finally traced my issue down to that I needed to replace the git-lfs.exe in the C:\Program Files\Git\mingw64\bin. Just replacing it in C:\Program Files\Git\cmd wasn't enough if you use the git bash

bhavyalatha26: @colinsheppard wanted to mention this specifically for windows users.

bhavyalatha26: @JustinPihony can you please mention the steps we need to correct or add wrt to the windows installation here . I do not use windows so I am not sure where and what exactly should I change the guide

JustinPihony: Just a footnote after the steps. Something like:

* Another alternative to above is to get the old git-lfs.exe from the release archives and replace the executable found in `[INSTALL PATH]\mingw6\bin` and `[INSTALL PATH]\cmd`, where the default installation path is `C:\Program Files\Git`

bhavyalatha26: done

colinsheppard:
test!

PR #1292, opened 2019-02-16
Cs/fix drive transit

colinsheppard:
Fixes the "Trying to set a driver" issue which was caused by Drive_Transit within trip replanning.

In doing so, I have also limited the number of attempts to replan within a trip to 2, after which the person walks.

Still get a small number of "My vehicle is not parked" errors which can be debugged separately.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1292)


colinsheppard:
test!

colinsheppard:
test!

colinsheppard:
test!

PR #1296, opened 2019-02-18
Vehicle Energy rate from data file

JustinPihony:
This needs more tests, and to clean up some odd edge cases. But putting this out there for now.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1296)


build.gradle
```diff
@@ -110,6 +110,9 @@ dependencies {
     compile group: 'org.apache.commons', name: 'commons-math3', version: '3.5'
     compile group: 'org.apache.httpcomponents', name: 'fluent-hc', version: '4.5.2'
 
+    // Apache 2.0
+    compile group: 'com.univocity', name: 'univocity-parsers', version: '2.8.1'
```
JustinPihony: We use another csv library that is noted as being 1.5x slower than this one. Writing is slightly slower than Jackson, but still much faster than the supercsv we use today...might be worth switching out entirely. I can open an issue if there's no strong reason

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
```diff
@@ -0,0 +1,82 @@
...
+              }
+            }
+            case None => gradePercentFilter += gradePercentBin -> mutable.Map(numberOfLanesBin -> rate)
+          }
+        }
+        case None => currentRateFilter += speedBin -> mutable.Map(gradePercentBin -> mutable.Map(numberOfLanesBin -> rate))
+      }
+    })
+    println(currentRateFilter)
+    currentRateFilter.toMap.map{
```
JustinPihony: This might not be necessary if the return type is set to immutable maps. But since this should only be done once and cached, then I don't see a need to remove it for now.

build.gradle
colinsheppard: No reason not to use something faster, please feel free to do so!

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
```diff
@@ -0,0 +1,82 @@
...
+  def getRecordsFrom(csvFile: File): Iterable[Record] = {
+    val settings = new CsvParserSettings()
+    settings.detectFormatAutomatically()
+    val parser = new CsvParser(settings)
+    val csvLocation = config.beam.agentsim.populationAdjustment
+    parser.iterateRecords(new File(csvLocation)).asScala
+  }
+}
+
+case class TravelData(speed:Int, gradePercent: Int, numberOfLanes: Int)
```
colinsheppard: please clarify units in name of field, "speedInMPerS"

colinsheppard: or better: speedInMetersPerSecond

colinsheppard:
Looks good so far. 

So BeamVehicle.collectFuelConsumptionData is slow. Now that we have a first version of our energy model, please turn fields in FuelConsumptionData into options and comment out any code that is preparing data that is never used by the current model.

Let's keep the code around for now because the energy modeling work is evolving and we may want to re-enable these fields in the very near future.

JustinPihony:
@colinsheppard Sure I can make this lazy, however I had thought that `collectFuelConsumptionData` was still going to be used as the fallback anyway?

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -102,7 +102,11 @@ class BeamVehicle(
     val distanceInMeters = beamLeg.travelPath.distanceInM
     val network = beamServices.matsimServices.getScenario.getNetwork
     val fuelConsumption = BeamVehicle.collectFuelConsumptionData(beamLeg, beamVehicleType, beamServices.networkHelper)
-    val energyConsumed = powerTrain.estimateConsumptionInJoules(fuelConsumption)
+    val energyConsumed =
+      beamServices.vehicleEnergy.getFuelConsumptionEnergyInJoulesUsing(
+        fuelConsumption,
+        overallFallBack = powerTrain.estimateConsumptionInJoules,
+        singleFallBack = powerTrain.estimateConsumptionInJoules)
```
colinsheppard: Not clear what "overal" vs "single" means. Also, see below, these are return an energy estimate but you are using that as if it were a rate.

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
```diff
@@ -0,0 +1,209 @@
...
+                           fallBack: BeamVehicle.FuelConsumptionData => Double): Option[Double] = {
+    val BeamVehicle.FuelConsumptionData(linkId, vehicleType, numberOfLanesOption, _, _, _, speedInMilesPerHourOption, _, _, _) =
+      fuelConsumptionData
+    val numberOfLanes: Int = numberOfLanesOption.getOrElse(0)
+    val speedInMilesPerHour: Double = speedInMilesPerHourOption.getOrElse(0)
+    val gradePercent: Double = linkIdToGradePercentMap.getOrElse(linkId, 0)
+    consumptionRateFilterStore.getPrimaryConsumptionRateFilterFor(vehicleType).flatMap(
+      consumptionRateFilterFuture => getRateUsing(consumptionRateFilterFuture,
+        numberOfLanes, speedInMilesPerHour, gradePercent)
+    ).orElse(
```
colinsheppard: Secondary energy consumption is not a backup if there's no match on primary. Totally different meaning. 

It's intended for 2 fuel systems. I.e. a plug-in hybrid (like a Volt) you charge up the battery and cruise on all-electric for first 50 miles, then when the battery runs down, it starts using gasoline and runs like a regular hybrid.

So we only want to use secondary if the fuel tank in primary is out.

But let's not solve that problem here in this ticket, let's get things working and solve that in a separate issue.

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
```diff
@@ -0,0 +1,209 @@
...
+      consumptionRateFilterStore.getSecondaryConsumptionRateFilterFor(vehicleType).flatMap(
+        consumptionRateFilterFuture => getRateUsing(consumptionRateFilterFuture,
+          numberOfLanes, speedInMilesPerHour, gradePercent)
+      )
+    ).orElse(Some(fallBack(fuelConsumptionData)))
```
colinsheppard: Here, this is causing the problems. The fall back is returning a value in units of energy (J) instead of a rate (J/m). You are essentially multiplying by the length of the road twice, leading to huge amounts of energy being consumed. Make the fall back return just the rate and we should be good.

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
```diff
@@ -0,0 +1,209 @@
...
+    val consumptionRateFilter = Await.result(consumptionRateFilterFuture, 1.minute)
+    val filteredRates = consumptionRateFilter
+      .collect {
+        case (speedInMilesPerHourBin, restOfFilter) if speedInMilesPerHourBin.hasDouble(speedInMilesPerHour) =>
+          restOfFilter
+      }
+      .flatten
+      .collect { case (gradePercentBin, restOfFilter) if gradePercentBin.hasDouble(gradePercent) => restOfFilter }
+      .flatten
+      .collect { case (numberOfLanesBin, rate) if numberOfLanesBin.has(numberOfLanes) => rate }
```
colinsheppard: I'm worried about all of the processing required to just get a single number out of a table... this has to happen a LOT of times, for every link that every vehicle traverses. So 100's of millions of times in a single run. Performance is going to be critical, is this the fastest way to access the data?

colinsheppard: E.g. how would this approach with collect/flatten/etc compare to just putting all of the keys into a single n-tuple and make a single map lookup.

Map[(Int,Int,...),Double]

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
JustinPihony: Yah, overall vs single is if the entire thing results in a None - maybe that isn't a realistic case though. Whereas single is if a single rate lookup yields a None. Typing that out, I guess that doesn't make sense since the single fallback will always yield something. I'll remove the overall then.

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
JustinPihony: Thanks - can't believe I missed that - fresh eyes are always helpful!

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
JustinPihony: Ahhh, I didn't realize that - although it makes sense. Yah, I'll remove secondary for right now and create an issue

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
JustinPihony: Discussed online and resolved. There was a mental mismap on my part where I had expected possible overlaps of ranges which would resolve by the overall lookup. Note that this code will no longer be possible to catch bad data:

```
if (ratesSize > 1)
      log.warn(
        s"More than one ($ratesSize) rate was found using: Number of Lanes = $numberOfLanes; Speed in MPH = $speedInMilesPerHour; Grade Percent = $gradePercent." +
          "The first will be used, but the data should be reviewed for range overlap."
      )
```

Since the new return immediately as a hit via `find` will just ignore them. The same outcome, but no warning log. I'm sure that's fine - more of an FYI

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

src/test/scala/beam/router/TransitInitializerSpec.scala
```diff
@@ -18,7 +18,7 @@ class TransitInitializerSpec extends WordSpecLike with Matchers with MockitoSuga
     val transitInitializer: TransitInitializer = init
 
     "return SUV, based on agency[217] and route[1342] map" in {
```
JustinPihony: If this is temporary then that's fine - otherwise the test description should probably be changed also :)

JustinPihony:
The only thing that I haven't done or issue-ized is 
> //TODO: Fix so that vehicle type is mapped to new NREL
but I believe you and/or Cong are handling that.

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

PR #1299, opened 2019-02-18
Fix replanning bug/omission

sfwatergit:
Experienced plans not getting endtimes from selected plans... fixes observed end time drift

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1299)

PR #1300, opened 2019-02-18
if experienced plan end time doesnt matach original plan then times d…

colinsheppard:
…rift even without TimeMutator on, prevent this

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1300)


colinsheppard:
redundant with https://github.com/LBNL-UCB-STI/beam/pull/1299

JustinPihony:
Can this branch be deleted then?

PR #1313, opened 2019-02-19
change behavior of writing config outputs

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1313)


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -424,7 +426,9 @@ trait BeamHelper extends LazyLogging {
     logger.info("Starting beam on branch {} at commit {}.", BashUtils.getBranch, BashUtils.getCommitHash)
     new java.io.File(outputDirectory).mkdirs
     val outConf = Paths.get(outputDirectory, "beam.conf")
-    Files.write(outConf, config.root().render(ConfigRenderOptions.concise()).getBytes)
+    val location = config.getString("config")
+
+    Files.copy(Paths.get(location), outConf, StandardCopyOption.REPLACE_EXISTING)
```
JustinPihony: Should we make config an `Option`, and fallback to the old behavior? That way you don't have to worry about changing the test code. Or if we don't want to write out the config for tests, then it could just be an `option.foreach` so that it only runs when config is specified....although a warning log should be output if it can't be written maybe? To help debug possible future issues.

PR #1282, opened 2019-02-13
Hl/#1058 alonso mora for ridehail

colinsheppard:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1282)

PR #1318, opened 2019-02-21
Hl/#1311 alonso mora++

htmlrb:
This version of the algorithm parallelize an enhanced version of Alonso-Mora algorithm when building the RTVGraph.
The way I proceeded is by using 
`Future.sequence(Future{}) onComplete {case Success....}`
then each parallel thread is supposed to return a list of vertices and edges. Once onComplete is called the graph is constructed.
@JustinPihony I will need your feedback on this. What might be the "unexpected" consequences that can affect BEAM simulations giving the use of Future and onComplete !?

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1318)


src/main/scala/beam/agentsim/agents/ridehail/ParallelAlonsoMoraAlgForRideHail.scala
```diff
@@ -0,0 +1,135 @@
...
+            finalRequestsList.appendAll(kRequestsList)
+          }
+        }
+        (vertices, edges)
+      }
+    }) onComplete {
+      case Success(result) =>
+        result foreach {
+          case (vertices, edges) =>
+            vertices foreach (vertex => rTvG.addVertex(vertex))
```
JustinPihony: First, good job not letting mutability enter the Future itself. However this completion may not work the way you think because you are then mutating data afterwards*. Yes, it's on the return thread, however onComplete is not blocking, so when you return `rTvG` you are returning it immediately. Later the onComplete will kick in and in theory may update the edge for the reference you passed out, but it gets shady. I see two options - block until the future is truly complete (in which case you should use map/foreach instead of onComplete to make sure ALL work is done) or return the Future itself, which will return the rTvG. Then the code that needs the final value can be the one to wait - maybe it doesn't need it until way later anyway?

*Again good that it is all bunched onto the same thread via the packaging of Future.sequence, though.

src/main/scala/beam/agentsim/agents/ridehail/allocation/PoolingAlonsoMora.scala
```diff
@@ -0,0 +1,174 @@
...
+    }
+    if(allocResponses.size>0 ){
+      if(allocResponses.find{
+        case RoutingRequiredToAllocateVehicle(req,routes) =>
+          false
+        case VehicleMatchedToCustomers(_,_,pickdrops) =>
+          pickdrops.filter(_.leg.isDefined).size>2
+      }.isDefined){
+//      if(tick>=21900){
+        val i = 0
```
JustinPihony: This seems to do nothing also, and can be removed.

src/main/scala/beam/agentsim/agents/ridehail/allocation/PoolingAlonsoMora.scala
```diff
@@ -0,0 +1,174 @@
...
+          )
+        } else {
+          allocResponses = allocResponses :+ NoVehicleAllocated(request)
+          request.groupedWithOtherRequests.foreach { req =>
+            allocResponses = allocResponses :+ NoVehicleAllocated(req)
+          }
+        }
+      }
+    }
+    if(toPool.size > 0){
```
JustinPihony: Would it be better to just use toPool.map, then return a tuple2 of the customerReqs and IdToReqs, then map off of that? It might even be more efficient as it would loop once instead of twice

src/main/scala/beam/agentsim/agents/ridehail/ParallelAlonsoMoraAlgForRideHail.scala
```diff
@@ -0,0 +1,135 @@
...
+  implicit val skimmer: BeamSkimmer
+) {
+
+  private def parallelBuildOfRTVGraph(): AlonsoMoraPoolingAlgForRideHail.RTVGraph = {
+    import scala.concurrent.ExecutionContext.Implicits.global
+    import scala.collection.mutable.{ListBuffer => MListBuffer}
+    val rTvG = AlonsoMoraPoolingAlgForRideHail.RTVGraph(classOf[DefaultEdge])
+    Future.sequence(supply.withFilter(_.getFreeSeats > 0).map { v =>
+      Future {
+        val vertices = MListBuffer.empty[RTVGraphNode]
```
REASY: Is it possible to extract this to a function (the body of Future) to be able to test independently and improve readability?

src/main/scala/beam/agentsim/agents/ridehail/ParallelAlonsoMoraAlgForRideHail.scala
JustinPihony: I just pushed a fix for this - go ahead and take a look. The functionality shouldn't have changed, just the level where variables were manipulated in the future, and that the end result is a Future, which is just being awaited in the test.

colinsheppard:
@htmlrb this PR makes #1282 unnecessary, correct? If so please close that one. Thanks.

colinsheppard:
test!

colinsheppard:
test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -1049,14 +1052,34 @@ class RideHailManager(
...
+//    consistentPickDrops.groupBy(_._1).foreach { personPickDrop =>
+//      val firstLeg = personPickDrop._2.head._2
+//      val lastLeg = personPickDrop._2.last._2
+//      val subtrip = allLegs.dropWhile(_ != firstLeg).drop(1).takeWhile(_ != lastLeg) :+ lastLeg
+//      passSched = passSched.addPassenger(personPickDrop._1, subtrip)
+//    }
+//    passSched
+
+    var pickDropsForGrouping: Map[VehiclePersonId,List[BeamLeg]] = Map()
+    var passengersToAdd = Set[VehiclePersonId]()
```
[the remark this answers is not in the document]
colinsheppard: @JustinPihony can you explain why?

src/main/scala/beam/router/model/BeamLeg.scala
```diff
@@ -48,12 +48,12 @@ object BeamLeg {
     new BeamLeg(0, WALK, 0, BeamPath(Vector(), Vector(), None, SpaceTime.zero, SpaceTime.zero, 0))
       .updateStartTime(startTime)
 
-  def makeLegsConsistent(legs: List[BeamLeg]): List[BeamLeg] = {
-    if (legs.size > 0) {
-      var runningStartTime = legs.head.startTime
-      for (leg <- legs) yield {
-        val newLeg = leg.updateStartTime(runningStartTime)
-        runningStartTime = newLeg.endTime
+  def makeLegsConsistent(legs: List[Option[BeamLeg]]): List[Option[BeamLeg]] = {
```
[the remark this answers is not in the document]
colinsheppard: idiosyncratic need to preserve Nones in the list but still operate over the legs. The Nones serve as placeholders that are important to the calling method.

src/main/scala/beam/agentsim/agents/ridehail/ParallelAlonsoMoraAlgForRideHail.scala
```diff
@@ -0,0 +1,135 @@
...
+import scala.collection.JavaConverters._
+import scala.concurrent.Future
+import scala.util.{Failure, Success}
+
+class ParallelAlonsoMoraAlgForRideHail(
+  demand: List[CustomerRequest],
+  supply: List[VehicleAndSchedule],
+  timeWindow: Map[MobilityServiceRequestType, Int],
+  radius: Int,
+  implicit val skimmer: BeamSkimmer
```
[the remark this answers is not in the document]
colinsheppard: I agree. I'm not a big fan of implicit contexts unless there's substantial boilerplate to be removed. I.e. we might consider it from BeamServices which is passed around everywhere. But in this method, I think it's overkill and obfuscates the code.

PR #1325, opened 2019-02-22
Bl/#1294 implement anchored time mutator 4ci

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1325)


src/main/resources/beam-template.conf
```diff
@@ -150,6 +150,7 @@ beam.replanning.ModuleProbability_3=0.1
 beam.replanning.Module_4="TimeMutator"
 beam.replanning.ModuleProbability_4=0.0
 beam.replanning.fractionOfIterationsToDisableInnovation = "double | Double.PositiveInfinity"
+beam.replanning.anchorEndTimeMutator.mutation.range=1
```
wrashid: rangeInSeconds - default: 1800

src/main/scala/beam/replanning/AnchorEndTimeMutator.scala
```diff
@@ -0,0 +1,73 @@
+package beam.replanning
+
+import beam.sim.config.BeamConfig
```
wrashid: we need to enable usage of this replanning module as well

src/main/resources/beam-template.conf
bhavyalatha26: done

src/main/scala/beam/replanning/AnchorEndTimeMutator.scala
bhavyalatha26: done

bhavyalatha26:
test!

PR #1327, opened 2019-02-23
Rjf/#1127 generalized parking search

### LINKED ISSUE #1127: EV Charging
Add charging behavior for personally owned EVs.

robfitzgerald:
This PR provides mostly structural changes to the management of non-trivial parking options and placeholders for integration of our EV charging cost calculations. It also serves to limit conceptual drift between the PEV team and the master branch.

Beamville and Sf-light scenarios have been tested and agent behavior seems correct when searching for and using capacity-constrained parking options.

#### bigger changes

- charging and parking ADTs re-written
  - charging code not integrated but placeholders are in place
- replaced previous parking search 
  - was indexed by constructing all viable combinations of parking attributes 
  - now a nested lookup tree of Id[Taz] -> ParkingType -> ParkingZone index
  - price and charging attributes of a ParkingZone are optional and stored as ParkingZone fields
- removed "ReserveParkingType" concept
- Id[ParkingStall] dropped; we instead track the Id of the aggregate of parking stalls, namely ParkingZone
  - ParkingZone Id is the array index of a parking zone (a fixed id throughout BEAM iterations), which is found via the nested lookup tree
- pulled TAZ out of TAZTreeMap; put both in their own package
  - changed classpath touches many files but no change to functionality

#### smaller changes

- removed unused arguments from ZonalParkingManager class, but didn't change signature of the props method
- added helpful comment to build.gradle for users of matsimConversion task (asked to add this by Rashid)

#### caveat

- only two small tests provided for the search code; would like to add more but wanted to stop somewhere

Closes #1127 

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1327)


src/main/scala/beam/agentsim/infrastructure/charging/ChargingInquiryData.scala
```diff
@@ -0,0 +1,16 @@
+package beam.agentsim.infrastructure.charging
+
+case class ChargingInquiryData(data: Map[ChargingPoint, ChargingPreference]) {
+  val agentMustCharge: Boolean = data.values.exists(_ == ChargingPreference.MustCharge)
+}
+
+object ChargingInquiryData {
```
JustinPihony: Interesting - I had missed [that you can provide a custom apply that overrides the built in one directly](https://github.com/scala/scala/pull/5816). I really thought this wouldn't compile - which I guess it wouldn't before that fix :p

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZone.scala
```diff
@@ -0,0 +1,88 @@
...
+import scala.language.higherKinds
+
+import cats.Monad
+
+import beam.agentsim.infrastructure.charging.ChargingPoint
+
+/**
+  * stores the number of stalls in use for a zone of parking stalls with a common set of attributes
+  *
+  * @param stallsAvailable a (mutable) count of stalls free (a semiphore)
```
JustinPihony: There's nothing I see that guarantees this as a semaphore, unless I'm missing something.

JustinPihony: In fact, I'm not 100% that the stall changing is even what would be considered pure...

src/main/scala/beam/agentsim/infrastructure/charging/ChargingInquiryData.scala
robfitzgerald: ya! i agree, and i don't know if maybe the scala compiler should instead warn about this, since i don't think this behavior is defined. 

this code isn't referenced anywhere yet, and is just a placeholder for things we have been working on.

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZone.scala
robfitzgerald: you're right, it's absolutely not a semaphore. i think i meant "it's designed like a semaphore in that it releases 'parking stalls' as it counts down from the max value, and no longer releases parking stalls when the stallsAvailable reaches 0." sort of as a note to communicate to the developer what it's doing logically.

purity, you mean FP? the m.pure wrapper here would allow us to run this in some context like cats.effect.IO if we wanted to in the future to run things in a more pure way. if we don't provide an effect, it will run as is without any context, so no one needs to worry about pure FP if they don't want to. cats was already in the dependencies so i went for it.

robfitzgerald: also, since you mentioned concurrency: i've written this with the assumption that only one Actor has access to this data at a time, therefore the assumption that it doesn't need to consider synchronization issues.

use case: a ZonalParkingManager or RideHailManager will be the sole broker for parking stalls, and it will refer to an Array[ParkingZone] which tells it (on aggregate) the availability of parking resources (aka ParkingStalls).

JustinPihony: > purity, you mean FP? the m.pure wrapper here would allow us to run this in some context like cats.effect.IO if we wanted to in the future to run things in a more pure way. if we don't provide an effect, it will run as is without any context, so no one needs to worry about pure FP if they don't want to. cats was already in the dependencies so i went for it.

Admittedly my pure FP/cats is not strong. From reading about pure it seemed that it was meant to wrap pure code - and this clearly side effects.

> also, since you mentioned concurrency: i've written this with the assumption that only one Actor has access to this data at a time, therefore the assumption that it doesn't need to consider synchronization issues.

I kind of figured this might stay within 1 actor, but I guess I was figuring what makes sure nobody abuses that? When the mutation is encapsulated inside the actor then it's explicitly wrapped. Whereas this can escape should somebody not pay attention. I don't have a better solution to such a problem, though - and will remove the blocker. If interested I can ping the Akka team to see if there IS a solution for what I'm talking about?

robfitzgerald: > Admittedly my pure FP/cats is not strong. From reading about pure it seemed that it was meant to wrap pure code - and this clearly side effects.

Sure, that's a confusing semantic. 

The purpose of any Monad.pure{} block is to wrap/lift the enclosed expression into an execution context or container. In pure FP, we want to control when side effects happen, and we do this by turning expressions with side-effects (impure) into something that will produce the same result every time you evaluate it (pure). So, this mutation would be run in some effect monad such as [cats.effect.IO](https://github.com/typelevel/cats-effect/blob/master/core/shared/src/main/scala/cats/effect/IO.scala). 

The "how" is that the expression would be placed in essentially a memoized thunk, deferring its execution until later and preventing multiple executions. This thunk would be evaluated when ZonalParkingManager wants to evaluate it. So, one way then to prevent race conditions on a mutation within a computational context (Actor in this case) is that the actor has a schedule to finally evaluate these mutations (stored in a concurrent collection type), and can perform them at a known phase of a procedure where it knows no reads/writes are happening. A nice feature of this abstraction is that it would also makes it easier to see where updates are happening, as well as running optimizations on the queue of operations stored.

I don't think anyone on this team is into the pure FP thing so I intentionally didn't go all in with this. The ParkingZone operations which mutate state are wrapped in a monad, but the code calling it isn't running it in a context, so it ends up running like a OO increment() op.

> I kind of figured this might stay within 1 actor, but I guess I was figuring what makes sure nobody abuses that?

I think the only way to enforce this via code is for the operations to be Actor-scoped, either in their parameters or in their visibility. But I think that it's clearer to have these things separated. My feeling is that we can't enforce it, but we can provide comments, and make sure that these attributes are private attributes of an Actor. The latter is currently the case as the `stalls` argument is not exposed (not prefixed with `val` in the class constructor):

```Scala
class ZonalParkingManager(
  val beamServices: BeamServices,
  stalls: Array[ParkingZone],
  searchTree: ParkingZoneSearch.StallSearch,
  rand: Random
) extends Actor
    with HasServices
    with ActorLogging { ... }
```

robfitzgerald:
I caught a bug in the ParkingZone constructor which used "DefaultParkingZoneId" where there was no id field; adding parkingZoneId field. I will submit another revision along with addressing @JustinPihony 's observation about my incorrect use of cats.

robfitzgerald:
$10.

Just kidding. Maybe in our PEV meeting this afternoon?

robfitzgerald:
test!

robfitzgerald:
test!

robfitzgerald:
test!

robfitzgerald:
test!

robfitzgerald:
test!

robfitzgerald:
test!

robfitzgerald:
test!

robfitzgerald:
somehow changes i have made have impacted "PersonWithPersonalVehiclePlanSpec" with the following unexpected conditions:
- a RoutingRequest is expected but a VehicleLeavesTrafficEvent is found
- what appears to be an unexpected VehicleEntersLinkEvent
- an EmbodyWithCurrentTravelTime is expected but LinkLeaveEvent found

src/main/resources/logback.xml
```diff
@@ -87,7 +87,7 @@
 
     </Logger>
 
-    <Logger name="beam.router.BeamRouter" level="ERROR" additivity="false">
+    <Logger name="beam.router.BeamRouter" level="INFO" additivity="false">
```
JustinPihony: Please make sure to convert this back to ERROR before merging

src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
```diff
@@ -1,468 +1,259 @@
...
+            _search(thisOuterRadius, thisOuterRadius * SearchFactor)
+        }
+      }
+    }
+
+    _search(0, searchStartRadius) match {
+      case Some(result) =>
+        result
+      case None =>
+        val newStall = ParkingStall.emergencyStall(random)
```
JustinPihony: Should emergency stall be used if nothing is found? Or let the agent pass in whether it is an emergency, so that they can try again later if it's not an emergency?

src/main/scala/beam/agentsim/infrastructure/ParkingManager.scala
```diff
@@ -20,25 +21,25 @@ object ParkingManager {
     destinationUtm: Location,
     activityType: String,
     attributesOfIndividual: AttributesOfIndividual,
-    chargingPreference: ChargingPreference,
+    chargingInquiryData: Option[ChargingInquiryData[String, String]],
```
JustinPihony: Should `ParkingInquiry` be removed altogether in favor of the one at `beam.agentsim.infrastructure`? That way it avoids confusion later down the road. Or at least some renaming should be done.  In fact it seems that the whole Inquiry/Response setup here can be removed.

src/main/scala/beam/agentsim/infrastructure/charging/ChargingInquiryData.scala
```diff
@@ -0,0 +1,19 @@
+package beam.agentsim.infrastructure.charging
+import beam.agentsim.agents.choice.logit.MultinomialLogit
+
+case class ChargingInquiryData[A, T](utility: Option[MultinomialLogit[A, T]]) {
```
JustinPihony: Is this going to end up being used? Since @robfitzgerald said earlier this was a sketch, and it seems to still not be used. It is currently just passed around, but not utilized.

johanneshiry: As far as I know it is intended to be used later on in the ParkingSearch. We agreed to pass over the whole utility function to include it in the search as a constraint if more than one parking spot with charging capability is available and to ensure that we take into account personal preferences on charging while selecting the parking spot.

src/main/resources/logback.xml
rjf-nrel: ah. good catch.

src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
rjf-nrel: this is in fact the behavior we want, but the "emergency" is with respect to the supply, not the demand. the notion is, give them an expensive and far-away stall if no stalls are found. this will push agents to seek other modes in future iterations. 

perhaps the name "EmergencyStall" is misleading. it's "emergency" as in the simulation ran out of options to offer, and so we urgently need to fill one in. but if you can think of another name, i'm open to it.

src/main/scala/beam/agentsim/infrastructure/ParkingManager.scala
rjf-nrel: that's actually what i did - i made `beam.agentsim.infrastructure.ParkingInquiry`. i just forgot to delete this file. ;-)

src/main/scala/beam/agentsim/infrastructure/charging/ChargingInquiryData.scala
rjf-nrel: this will be replaced with the work that Johannes is up to. it's wired into `beam.agentsim.agents.ChoosesParking` and was written by Johannes, so i left it in as a placeholder. it will not interfere with runs of BEAM using this PR and nothing will end up instantiating a ChargingInquiryData object (it's always Option[ChargingInquiryData] = None).

src/main/scala/beam/agentsim/infrastructure/ParkingManager.scala
rjf-nrel: this is logged as a change request on ParkingManager.scala. could you remove it from your review? or am i supposed to do that somehow?

src/main/resources/logback.xml
robfitzgerald: ah, good catch.

src/main/scala/beam/agentsim/infrastructure/ParkingManager.scala
robfitzgerald: that's actually what i did - i made `beam.agentsim.infrastructure.ParkingInquiry`. i just forgot to delete this file. ;-)

src/main/scala/beam/agentsim/infrastructure/charging/ChargingInquiryData.scala
robfitzgerald: this will be replaced with the work that Johannes is up to. it's wired into `beam.agentsim.agents.ChoosesParking` and was written by Johannes, so i left it in as a placeholder. it will not interfere with runs of BEAM using this PR and nothing will end up instantiating a ChargingInquiryData object (it's always `Option[ChargingInquiryData] = None`).

src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
robfitzgerald: this is in fact the behavior we want, but the "emergency" is with respect to the supply, not the demand. the notion is, give them an expensive and far-away stall if no stalls are found. this will push agents to seek other modes in future iterations. 

perhaps the name "EmergencyStall" is misleading. it's "emergency" as in the simulation ran out of options to offer, and so we urgently need to fill one in. but if you can think of another name, i'm open to it.

wrashid: backup stall?

JustinPihony: I figured based on the code. Note that far away if it's an emergency seems counterproductive - I can understand costly, but far away defeats the purpose of meeting supply. 

As to the name...naming sucks, right! Emergency is good, and I was just verifying. Another alternative would be LastResortStall maybe?

wrashid: LastResortStall +1

src/main/scala/beam/agentsim/infrastructure/charging/ChargingInquiryData.scala
JustinPihony: ok, just wanted to make sure it didn't get lost in the shuffle :)

src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
robfitzgerald: naming is NP-Complete ;-)

I'll switch it to __LastResortStall__.

robfitzgerald:
test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

robfitzgerald:
Test!

PR #1329, opened 2019-02-24
Cs/bushwack crs

colinsheppard:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1329)


colinsheppard:
test!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -780,28 +780,24 @@ trait ChoosesMode {
...
-                      origLeg.copy(
-                        cost = origLeg.cost * poolingInfo.costFactor,
-                        isPooledTrip = origLeg.isRideHail,
-                        beamLeg = origLeg.beamLeg.scaleLegDuration(poolingInfo.timeFactor)
-                    )
-                  )
+              val pooledLegs = origLegs.map { origLeg =>
+                origLeg.copy(
+                  cost = origLeg.cost * poolingInfo.costFactor,
+                  isPooledTrip = origLeg.isRideHail,
```
wrashid: just a question: if it is ridehail vehicle, it is automatically pooling?

colinsheppard: No, that was just a way to differentiate between ride hail leg and any other leg which could be in the itinerary.

PR #1338, opened 2019-02-26
Scenario: Ability to load scenario from parquet

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1338)


src/main/scala/beam/utils/scenario/DataExchange.scala
```diff
@@ -0,0 +1,42 @@
...
+    case InputType.Parquet => "parquet"
+    case InputType.CSV     => "csv"
+  }
+}
+
+object InputType {
+  case object Parquet extends InputType
+  case object CSV extends InputType
+
+  def apply(`type`: String): InputType = {
```
JustinPihony: I'm not a fan of using a keyword for a parameter. Just because you _can_ doesn't mean you should.

src/main/scala/beam/utils/ParquetReader.scala
```diff
@@ -0,0 +1,70 @@
+package beam.utils
```
JustinPihony: There isn't a truly clean way to do it, so maybe this is the best option - but it might be worth documenting that you can also use [parquet-tools](https://github.com/apache/parquet-mr/tree/master/parquet-tools)

REASY: Is not it command line tool? I'm not sure how I can use this that library just by itself. I can see an example of [CatCommand.java](https://github.com/apache/parquet-mr/blob/master/parquet-tools/src/main/java/org/apache/parquet/tools/command/CatCommand.java#L82) and how they read parquet file.

REASY:
The runs on production-sfbay (https://github.com/LBNL-UCB-STI/beam/tree/production-sfbay/, commit fbef8bb4a4a50f505459f41cfe13a3c9fb36b529).

Master branch (80248689c1a11b64a735ad53f433f2d4cf9904aa) (Full archive: https://drive.google.com/open?id=1CHER0HSEFGFaJtEv57T_2NaBA_AX5gCE):





Run on urban sim data: https://github.com/LBNL-UCB-STI/beam/tree/ma/%23856-beamville-csv-setup-2-4ci/test/input/sfbay/test-data/urbansim in this brach (commit 7417dfebc8318961475388e664b4867edff9c21a) (Full archive: https://drive.google.com/open?id=1Sf-ABju3V2ZoNSWqOFZdWzfh2_GwZX0T):




@colinsheppard @wrashid What do you think? Doesn't it look as as we there is some issue either with UrbanSim data or the code?

colinsheppard:
I was able to take a closer look and I think the difference we see is due to your reader being right and the old reader having a mistake. 

In the urbansim data, all of the plans have "drive" for their mode. So this means everyone in the first iteration attempts to drive, EXCEPT for the people who don't have a car, these people therefore use the WALK which is the fallback. That explains by 75% drive / 25% walk in first iter.

Then in subsequent iterations, as replanning happens, some people begin to use other modes (slowly but it's happening).

So I'm fairly sure what you've implemented works! The only way to know 100% is to convert to XML as you already identified. But I think comparing to the old CSV reader is not worthwhile. My guess is that old reader wasn't properly passing the mode from plans.csv to the Leg in the persons' plans.

colinsheppard:
Here's what it looks like if I delete "drive" mode from the plans and let them select internally. Reasonable!

modechoice

src/main/scala/beam/utils/ParquetReader.scala
JustinPihony: Yah it is a command line tool - I thought that's what ParquetReader was too - I must have been mistaken

PR #1348, opened 2019-02-27
Bl/#1339 add steps to set maxram in user guide

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1348)


docs/users.rst
```diff
@@ -113,6 +113,27 @@ The simplest, smallest, and fastest is the beamville scenario (described below).
...
+
+    export MAXRAM=10g
+
+where 10g = 10GB
+
+Similarly on windows it can be set using the below command :
+
+.. code-block:: bash
+
+    setx MAXRAM="10g"
```
JustinPihony: Just an FYI that setx will set the variable permanently (different than export above). Should we say set for consistency? @colinsheppard @wrashid

PR #1349, opened 2019-02-27
avoid bad coordinate conversions when easy to detect, but warn

colinsheppard:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1349)


src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -459,7 +462,7 @@ object HouseholdActor {
       // Pipe my cars through the parking manager
       // and complete initialization only when I got them all.
       Future
-        .sequence(vehicles.values.map { veh =>
+        .sequence(vehicles.filter(_._2.beamVehicleType.automationLevel <= 3).values.map { veh =>
```
wrashid: only non fully automated cars go through this? what happens with higher automation level?

colinsheppard: after reversing the filter, it's there to only complete intialization for CAVs because ATM that is all the householdActor managers. the non-CAV vehicles are now managed by HouseholdFleetManager

PR #1359, opened 2019-02-28
DebugTool: Print the agent type and the line when coordinates are in wrong format

REASY:
It uses [sourcecode](https://github.com/lihaoyi/sourcecode) to get source file, line and enclosing name on compile-time using scala macros. Magic happens due to https://github.com/LBNL-UCB-STI/beam/compare/art/print-stacktrace-when-coordinates-are-wrong-4ci?expand=1#diff-a9cc8fe78944a2aa3bbe386d2d2c9629R10 trait `DebugUtil` with `implicit def getCallerInfo` and this implicit comes to the scope because of https://github.com/LBNL-UCB-STI/beam/compare/art/print-stacktrace-when-coordinates-are-wrong-4ci?expand=1#diff-526ef9e78edd0f381ba50a20bdc1e8c1R28. It was (IMHO) the easiest way with making less code changes.
An example of output:
```
03:30:09.778 WARN  beam.sim.common.GeoUtilsImpl - [4] Coordinate does not appear to be in WGS. No conversion will happen: [x=552954.6304091329][y=4176210.316522328]. 
 class beam.agentsim.agents.PersonAgent beam.agentsim.agents.modalbehaviors.DrivesVehicle#$anonfun#applyOrElse(DrivesVehicle.scala:222) C:\repos\beam\src\main\scala\beam\agentsim\agents\modalbehaviors\DrivesVehicle.scala:222
```

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1359)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -247,6 +247,8 @@ dependencies {
     scoverage "org.scoverage:scalac-scoverage-plugin_${scalaBinaryVersion}:1.3.1", "org.scoverage:scalac-scoverage-runtime_${scalaBinaryVersion}:1.3.1"
 
     compile 'org.apache.commons:commons-compress:1.18'
+
+    compile group: 'com.lihaoyi', name: 'sourcecode_2.12', version: '0.1.5'
 }
 
 // Autoformatting using scalafmt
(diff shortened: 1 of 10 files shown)


michaz:
With #1771, I'm now just throwing an exception when encountering WGS coordinates that aren't. At least in one very prominent place. Would suggest we do the same everywhere. Then we get the agent number for free, if the logger is configured accordingly.

REASY:
This is solved on https://github.com/LBNL-UCB-STI/beam/pull/1771

PR #1365, opened 2019-03-01
Hl/#1336 hh cav improvements

htmlrb:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1365)


colinsheppard:
Was subsumed into #1380

PR #1369, opened 2019-03-01
Cs/#1345 skimmer

colinsheppard:
first impl of a skimmer that collects data and averages it.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1369)


src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -1,19 +1,41 @@
...
-  override def toString =
-    s"[time:${timeAndCost.time.getOrElse("NA")}][cost:${timeAndCost.cost.getOrElse("NA")}][distance:${distance.getOrElse("NA")}]"
-}
+import scala.collection.concurrent.TrieMap
 
 //TODO to be validated against google api
-class BeamSkimmer(services: Option[BeamServices] = None) {
+class BeamSkimmer @Inject()() extends IterationEndsListener {
+  // The OD/Mode/Time Matrix
+  var previousSkims: TrieMap[(Int, BeamMode, Id[TAZ], Id[TAZ]), Skim] = TrieMap()
```
wrashid: should a case class be used instead of 4-tuple key, e.g. as a refactor after merge? / same for var skims

src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -41,50 +63,175 @@ class BeamSkimmer(services: Option[BeamServices] = None) {
...
+    val destTaz = beamServices.tazTreeMap
+      .getTAZ(destCoord.getX, destCoord.getY)
+      .tazId
+    val timeBin = timeToBin(origLeg.startTime)
+    val key = (timeBin, mode, origTaz, destTaz)
+    val payload =
+      Skim(trip.totalTravelTimeInSecs.toDouble, trip.beamLegs().map(_.travelPath.distanceInM).sum, trip.costEstimate, 1)
+    skims.get(key) match {
+      case Some(existingSkim) =>
+        val newPayload = Skim(
```
wrashid: the current implementation seems to keep average over all iterations - no memory limit to reflect current state faster, e.g. only last 2-5 iterations? due to the way that averaging works and how the simulation system evolves, the skim data might be representing "iteration-i/2" state instead of more recent one. just to make sure, this is what is needed for the applications using the skim. e.g. the ridehail repositioning history collector just keeps last 2 iteration history (and not average over all seen iterations till now)

src/main/scala/beam/router/BeamSkimmer.scala
colinsheppard: That would make things more clear. Let's do that when we expand what is stored in the skimmer.

colinsheppard:
Was subsumed into #1380

PR #1390, opened 2019-03-04
range refactoring

### LINKED ISSUE #1358: Improve Census Range implementation
The current implementation of Ranges is not technically correct/inconsistent in several cases: (What are their types? Are ranges internally represented as open/closed (Apparently they become closed when computed, but this changes the mathematical sense of them)? The expression Range("[20,19)") is legal and has an upperBound of 18, but a lowerBound of 20?, Range("(10,10)") is legal, etc.)

We also want to be able to support some range comparisons, arithmetic, and operations.

Suggested Fix:
[] Use Guava implementation of [Range](https://github.com/google/guava/wiki/RangesExplained)s internally.

zishanbilal:
Fixed #1358

### DIFF
--- a/src/main/scala/beam/sim/common/Range.scala
+++ b/src/main/scala/beam/sim/common/Range.scala
@@ -26,19 +26,20 @@ case class Range(lowerEndpoint: Int, upperEndpoint: Int) {
 
 object Range {
 
-  def apply(pattern: String): Range =
-    if (pattern == null || pattern.isEmpty) Range.empty()
+  val pattern = """\[-?\d*\:-?\d*\]"""
+  def apply(exp: String): Range =
+    if (exp == null || exp.isEmpty) Range.empty()
+    else if(!exp.matches(pattern)) throw new RuntimeException(s"Invalid range expression $exp, it should be [<NUM>:<NUM>].")
     else {
-      val bounds = pattern.split(":")
-      val lowerBound = Try(
-        bounds(0).substring(1).toInt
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/sim/common/Range.scala
```diff
@@ -26,33 +26,58 @@ case class Range(lowerBound: Int, upperBound: Int) {
 
 object Range {
 
-  def apply(pattern: String): Range =
-    if (pattern == null || pattern.isEmpty) Range.empty()
+  val pattern = """\[-?\d*\:-?\d*\]"""
```
JustinPihony: I think this will break some scenarios. For example, I the data for VehicleEnergy is in the format (#,#].

src/test/scala/beam/agentsim/agents/choice/mode/RangeSpec.scala
```diff
@@ -46,4 +47,7 @@ class RangeSpec extends FlatSpec {
     assert(!Range("[1:10)").has(10))
   }
 
+  "[10:1]" should "throw RuntimeException" in {
```
JustinPihony: This took me a second to realize - could it be more generic to say something like "lower value that is greater than higher value" should ...

src/main/scala/beam/sim/common/Range.scala
zishanbilal: it uses in specific case.

PR #1426, opened 2019-03-09
Beam skimmer cleanup 4ci

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1426)

### DIFF
--- a/src/main/scala/beam/router/BeamSkimmer.scala
+++ b/src/main/scala/beam/router/BeamSkimmer.scala
@@ -34,7 +34,7 @@ class BeamSkimmer @Inject()() extends IterationEndsListener {
   // The OD/Mode/Time Matrix
   private var previousSkims: TrieMap[(Int, BeamMode, Id[TAZ], Id[TAZ]), SkimInternal] = TrieMap()
   private var skims: TrieMap[(Int, BeamMode, Id[TAZ], Id[TAZ]), SkimInternal] = TrieMap()
-  private var modalAverage: TrieMap[BeamMode, SkimInternal] = TrieMap()
+  private val modalAverage: TrieMap[BeamMode, SkimInternal] = TrieMap()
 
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -182,23 +179,22 @@ class BeamSkimmer @Inject()() extends IterationEndsListener {
...
-    ((existingAverage * existingCount + newValue) / (existingCount + 1))
+  def mergeAverage(existingAverage: Double, existingCount: Int, newValue: Double): Double = {
+    (existingAverage * existingCount + newValue) / (existingCount + 1)
+  }
 
   override def notifyIterationEnds(event: IterationEndsEvent): Unit = {
     val fileHeader = "hour,mode,origTaz,destTaz,travelTimeInS,cost,distanceInM,numObservations"
-    // Output file relative path
-    val filePath = event.getServices.getControlerIO.getIterationFilename(
+    val fileRelativePath = event.getServices.getControlerIO.getIterationFilename(
```
wrashid: don't rename

src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -132,20 +131,18 @@ class BeamSkimmer @Inject()() extends IterationEndsListener {
 
   private def getSkimValue(time: Int, mode: BeamMode, orig: Id[TAZ], dest: Id[TAZ]): Option[SkimInternal] = {
     skims.get((timeToBin(time), mode, orig, dest)) match {
-      case Some(skim) =>
-        Some(skim)
+      case value @ Some(_) =>
```
wrashid: call it skim instead of value?

carloscaldas: how about `someSkim`?

PR #1447, opened 2019-03-10
Warm start integrate RouteHistory 4ci

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1447)


src/main/scala/beam/router/RouteHistory.scala
```diff
@@ -73,4 +107,98 @@ class RouteHistory @Inject()() extends LazyLogging {
...
+            }
+        }
+    }
+    val csvContent = flattenedRouteHistory.view
+      .map { tuple =>
+        s"${tuple._1},${tuple._2},${tuple._3},${tuple._4}"
+      }
+      .mkString("\n")
+
+    FileUtils.writeToFile(
```
JustinPihony: That was some interesting magic that I had to trace down - I hadn't expected this would automatically gzip, but it looks like it's built into matsim to gzip if it ends in gz

src/main/scala/beam/router/RouteHistory.scala
```diff
@@ -73,4 +107,98 @@ class RouteHistory @Inject()() extends LazyLogging {
...
+          case (originTazId: OriginTazId, destinations: TrieMap[DestTazId, Routes]) =>
+            destinations.flatMap {
+              case (destTazId: DestTazId, path: Routes) =>
+                Some(timeBin, originTazId, destTazId, path.mkString(":"))
+            }
+        }
+    }
+    val csvContent = flattenedRouteHistory.view
+      .map { tuple =>
+        s"${tuple._1},${tuple._2},${tuple._3},${tuple._4}"
```
JustinPihony: It might be more readable if you pattern match as part of the map partial function and extract the four components instead of using the underscore notation.

src/main/scala/beam/router/RouteHistory.scala
```diff
@@ -73,4 +107,98 @@ class RouteHistory @Inject()() extends LazyLogging {
...
     }
   }
+  override def notifyIterationEnds(event: IterationEndsEvent): Unit = {
+    val fileHeader = "timeBin,originTAZId,destTAZId,route"
+    val filePath = event.getServices.getControlerIO.getIterationFilename(
+      event.getServices.getIterationNumber,
+      RouteHistory.outputFileBaseName + ".csv.gz"
+    )
+
+    val flattenedRouteHistory: Iterable[(TimeBin, OriginTazId, DestTazId, String)] = routeHistory.flatMap {
```
JustinPihony: It seems to me that this might be read more clearly if it was in a for-comprehension, but maybe that's just my own preference. Take it or leave it.

src/main/scala/beam/router/RouteHistory.scala
carloscaldas: Agree

carloscaldas:
test!

carloscaldas:
test!

carloscaldas:
@colinsheppard converted the method `toCsv` from `String` to `Iterator[String]`. This way the big string is not built and is also easier to handle in the code, testing, etc. Please check.

carloscaldas:
test!

PR #1459, opened 2019-03-11
Zn/smart scenarios

zneedell:
Changes for SMART scenarios:

-- randomly generate ride hail vehicle type, driver shift, and fleet size (as % of total vehicle fleet)
-- Scale up or down private vehicle fleet size with config parameter
-- Sample randomly from urbansim households by an input factor
-- Generate individual VOTT based on household income (if individual VOTT is not already specified)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1459)


wrashid:
Test!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -306,8 +293,16 @@ class RideHailManager(
   // Are we in the middle of processing a batch?
   var currentlyProcessingTimeoutTrigger: Option[TriggerWithId] = None
 
+  private val numHouseholdVehicles = scenario.getHouseholds.getHouseholds
+    .values()
+    .asScala
+    .map(_.getVehicleIds.size())
```
JustinPihony: This could be done with a `reduce` or `fold` instead of a `map/sum` - which should be more performant I believe.

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -323,16 +318,61 @@ class RideHailManager(
       val persons: Iterable[Person] = RandomUtils
         .shuffle(scenario.getPopulation.getPersons.values().asScala, rand)
         .take(numRideHailAgents.toInt)
+      val vehicleTypes = VehiclesAdjustment
+        .getVehicleAdjustment(beamServices)
+        .sampleRideHailVehicleTypes(
+          numVehicles = numRideHailAgents.toInt,
+          vehicleCategory = VehicleCategory.Car
+        )
+      var activityEndTimes: ArrayBuffer[Int] = new ArrayBuffer[Int]()
```
JustinPihony: Please do not use a var. It is the worst possible scenario for creating garbage. Use an immutable reference to a mutable buffer instead.

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -323,16 +318,61 @@ class RideHailManager(
...
+      scenario.getPopulation.getPersons.asScala.map(
+        _._2.getSelectedPlan.getPlanElements.asScala
```
JustinPihony: I would suggest using a `.collect` here:

```
_._2.getSelectedPlan.getPlanElements.asScala.collect{ 
  case activity: Activity if activity.getEndTime.toInt > 0 => activity.getEndTime.toInt 
}.foreach(activityEndTimes += _)
```

At minimum note that you should use a `foreach` over a `map` since you are not creating anything - and it makes it explicit that you are mutating.

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -323,16 +318,61 @@ class RideHailManager(
...
+              fleetData += createRideHailVehicleAndAgent(
+                person.getId.toString,
+                vehicleType,
+                rideInitialLocation,
+                shiftString,
+                None
+              )
+            }
+          } catch {
+            case ex: Throwable =>
```
JustinPihony: There is no point in catching and rethrowing. Let the top level handle the catching and logging if that's all you are doing.

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -1328,4 +1370,13 @@ class RideHailManager(
     rideInitialLocation
   }
 
+  private def convertToShiftString(startTimes: ArrayBuffer[Int], endTimes: ArrayBuffer[Int]): Option[String] = {
+    if (startTimes.length != endTimes.length) {
+      None
+    } else {
+      var outStr: String = new String
+      (startTimes zip endTimes).foreach(x => outStr += ("{" + x._1.toString + ":" + x._2.toString + "}"))
```
JustinPihony: This should be a stringbuilder at minimum to avoid garbage. But in this case you could probably turn it into a `mkString`

src/main/scala/beam/sim/RideHailFleetInitializer.scala
```diff
@@ -143,7 +143,15 @@ class RideHailFleetInitializer extends LazyLogging {
         )
       val fileHeader = classOf[RideHailAgentInputData].getDeclaredFields.map(_.getName).mkString(", ")
       val data = fleetData map { f =>
-        f.productIterator.map(f => if (f == None) "" else f) mkString ", "
+        f.productIterator.map(
```
JustinPihony: This can use `map`s partial function overload:

```
f.productIterator.map{
  case Some(x:String) => x
  case x: String       => x
  case None            => ""
  case _               => f
}
```

src/main/scala/beam/sim/population/PopulationAdjustment.scala
```diff
@@ -229,6 +224,14 @@ object PopulationAdjustment extends LazyLogging {
           agentsim.agents.Population.getVehiclesFromHousehold(household, beamServices)
         HouseholdAttributes(household, houseHoldVehicles)
     }
+    // Read person attribute "valueOfTime", use function of HH income if not, and default it to the respective config value if neither is found
+    val valueOfTime: Double =
+      Option(personAttributes.getAttribute(person.getId.toString, "valueOfTime"))
+        .map(_.asInstanceOf[Double])
```
JustinPihony: This should probably just have been a `toDouble`, but maybe there is a reason to use `asInstanceOf`?

PR #1478, opened 2019-03-12
throw exception if bike is missing in vehicle types and remove default bike

zishanbilal:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1478)


JustinPihony:
I used the UI to fix - but this still needs scalaFmtAll run on it

zishanbilal:
test!

PR #1481, opened 2019-03-12
Scenario: Fixed sampling

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1481)

### DIFF
--- a/src/main/scala/beam/sim/vehicles/UniformVehiclesAdjustment.scala
+++ b/src/main/scala/beam/sim/vehicles/UniformVehiclesAdjustment.scala
@@ -12,16 +12,19 @@ case class UniformVehiclesAdjustment(beamServices: BeamServices) extends Vehicle
   private val realDistribution: UniformRealDistribution = new UniformRealDistribution()
   realDistribution.reseedRandomGenerator(beamServices.beamConfig.matsim.modules.global.randomSeed)
 
-  private val vehicleTypesAndProbabilitiesByCategory = beamServices.vehicleTypes.values.groupBy(_.vehicleCategory).map {
-    catAndType =>
-      val probSum = catAndType._2.map(_.sampleProbabilityWithinCategory).sum
-      val cumulProbs = catAndType._2
-        .map(_.sampleProbabilityWithinCategory / probSum)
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/sim/vehicles/UniformVehiclesAdjustment.scala
```diff
@@ -12,16 +12,19 @@ case class UniformVehiclesAdjustment(beamServices: BeamServices) extends Vehicle
...
-    catAndType =>
-      val probSum = catAndType._2.map(_.sampleProbabilityWithinCategory).sum
-      val cumulProbs = catAndType._2
-        .map(_.sampleProbabilityWithinCategory / probSum)
-        .scan(0.0)(_ + _)
-        .drop(1)
-        .toList :+ 1.0
-      (catAndType._1, catAndType._2.zip(cumulProbs).map(pair => (pair._1, pair._2)))
-  }
+  private val vehicleTypesAndProbabilitiesByCategory: Map[VehicleCategory, Array[(BeamVehicleType, Double)]] =
```
REASY: @colinsheppard, it's just rewritten with local vals and removed the usage of `._1` and `._2` to access the tuple (instead use `case` and `unapply`)

src/main/scala/beam/sim/vehicles/UniformVehiclesAdjustment.scala
```diff
@@ -12,16 +12,19 @@ case class UniformVehiclesAdjustment(beamServices: BeamServices) extends Vehicle
...
+    beamServices.vehicleTypes.values.groupBy(_.vehicleCategory).map {
+      case (cat, vehTypes) =>
+        val probSum = vehTypes.map(_.sampleProbabilityWithinCategory).sum
+        val cumulativeProbabilities = vehTypes
+          .map(_.sampleProbabilityWithinCategory / probSum)
+          .scan(0.0)(_ + _)
+          .drop(1)
```
REASY: @colinsheppard I'm just curious why we drop the first probability (which is zero) ?

PR #1487, opened 2019-03-13
make writing of skims include all od mode hour combinations

colinsheppard:
Skims written now cover every combination of O D mode hour and use defaults when no observations available.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1487)


src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -64,13 +64,47 @@ class BeamSkimmer @Inject()(
...
+            departureTime,
+            mode,
+            travelTime,
+            new BeamPath(null, null, None, null, null, travelDistance)
+          ),
+          vehicleTypeId,
+          beamServices
+        )
+      case RIDE_HAIL =>
+        beamServices.beamConfig.beam.agentsim.agents.rideHail.defaultCostPerMile * travelDistance / 1609.0 + beamServices.beamConfig.beam.agentsim.agents.rideHail.defaultCostPerMinute * travelTime / 60.0
```
JustinPihony: Could you put 1609 in a variable so that it make sense what this value is supposed to represent?

src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -221,16 +239,68 @@ class BeamSkimmer @Inject()(
...
+    val uniqueModes = skims.map(keyVal => keyVal._1._2).toList.distinct
+    val uniqueTimeBins = (0 to 23)
+
+    beamServicesOpt match {
+      case Some(beamServices) =>
+        val dummyId = Id.create("NA", classOf[BeamVehicleType])
+        val writer = IOUtils.getBufferedWriter(filePath)
+        writer.write(fileHeader)
+        writer.write("\n")
+        beamServices.tazTreeMap.getTAZs
```
JustinPihony: I'm wondering if a for loop would be more readable here, instead of the heavily nested foreach?

src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -112,11 +126,10 @@ class BeamSkimmer @Inject()(
     origin: Location,
     destination: Location,
     departureTime: Int,
-    vehicleTypeId: org.matsim.api.core.v01.Id[BeamVehicleType],
-    beamServices: BeamServices
+    vehicleTypeId: org.matsim.api.core.v01.Id[BeamVehicleType]
   ): (Double, Double) = {
-    val origTaz = beamServices.tazTreeMap.getTAZ(origin.getX, origin.getY).tazId
-    val destTaz = beamServices.tazTreeMap.getTAZ(destination.getX, destination.getY).tazId
+    val origTaz = beamServicesOpt.get.tazTreeMap.getTAZ(origin.getX, origin.getY).tazId
```
JustinPihony: Could we more gracefully handle this and be explicit in the exception that will be thrown to help later debugging? Just calling get makes using options pointless as it results in the same problems as null pointers. A high level match could work here, just as it is elsewhere:

```
beamServicesOpt match {
  case None => throw MeaningfulException
  case Some(beamServices) => RestOfLogic
}
```

PR #1505, opened 2019-03-14
reset vehicleEvents and vehicleEventsCache inside resetStats

### LINKED ISSUE #1479: cleanup memory leak fixes
Here is two places where I left comment:
https://github.com/LBNL-UCB-STI/beam/blob/master/src/main/java/beam/analysis/plots/DeadHeadingAnalysis.java#L137
https://github.com/LBNL-UCB-STI/beam/blob/master/src/main/java/beam/analysis/plots/DeadHeadingAnalysis.java#L348

But there is method `resetStats`, https://github.com/LBNL-UCB-STI/beam/blob/master/src/main/java/beam/analysis/plots/DeadHeadingAnalysis.java#L112, maybe we just need to clean vehicleEvents and vehicleEventsCache in `resetStats`:
vehicleEvents.clear();
vehicleEventsCache.clear();

bhavyalatha26:
Closes #1479 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1505)


JustinPihony:
Can you remove the comments that talk about this being a dirty hack since you are now doing the actions in the reset. I'm going to defer to @REASY to review since he was the one who just patched this not long ago (https://github.com/LBNL-UCB-STI/beam/commit/a45c719d18e3e54189188db4fd1321329fefcb88).

src/main/java/beam/analysis/plots/DeadHeadingAnalysis.java
```diff
@@ -168,9 +170,10 @@ private void processDeadHeadingDistanceRemainingRepositionings() {
                 }
             }
         }
-        finally {
-            vehicleEvents.clear();
+        catch (Exception e) {
```
JustinPihony: Changing the finally's to catch's is a behavior change. Now it will swallow exceptions, whereas before it didn't. Should it rethrow, call reset, swallow (as here), or something else?

michaz:
Let's revisit this.

- This class uses `static` fields, probably by accident.
- It is really two classes in one: One that operates on `vehicleEvents` and one that operates on `vehicleEventsCache`, with two sets of methods.
- I'm pretty certain all the maps should just be cleared between iterations.
- If lots of events pile up in those maps, there may be a bug beyond the map not being cleared, because the _intention_ seems to be that they _don't_ pile up.
- The test uses methods that only the test uses, and where it's not obvious if the real methods use the same functionality. If I delete the test and remove unused methods, half the class disappears.

Since so many people have worked on this now, maybe the author could help? @bhavyalatha26 ?

michaz:
Let's revisit this.

- This class uses `static` fields, probably by accident.
- It is really two classes in one: One that operates on `vehicleEvents` and one that operates on `vehicleEventsCache`, with two sets of methods.
- I'm pretty certain all the maps should just be cleared between iterations.
- If lots of events pile up in those maps, there may be a bug beyond the map not being cleared, because the _intention_ seems to be that they _don't_ pile up.
- The test uses methods that only the test uses, and where it's not obvious if the real methods use the same functionality. If I delete the test and remove unused methods, half the class disappears.

Since so many people have worked on this now, maybe the author could help? @bhavyalatha26 ?

Issue #1488, opened 2019-03-13
ExponentialLoggerWrapperImpl might cause memory leak


REASY:
Currently no one call `ExponentialLoggerWrapperImpl.reset()` to clean-up the data which is used for https://github.com/LBNL-UCB-STI/beam/blob/master/src/main/scala/beam/utils/logging/ExponentialLoggerWrapperImpl.scala#L67
Also that field is private for companion object, but can shared across multiple classes. In that case `reset` will reset the message -> counter stats for all others.

Also, why would not we use some fast hash function instead of storing the whole string to reduce memory usage. Here is great comparison of fast non-secure hash functions: https://softwareengineering.stackexchange.com/questions/49550/which-hashing-algorithm-is-best-for-uniqueness-and-speed/145633#145633

carloscaldas:
> Also, why would not we use some fast hash function instead of storing the whole string to reduce memory usage.

Currently is storing the first 20 characters. Storing the hash made the message useless unless we have all messages registered in the same place.

> Also that field is private for companion object, but can shared across multiple classes. In that case reset will reset the message -> counter stats for all others.

ExponentialLoggerWrapperImpl should not be referenced by other classes. Clients should be coupled to the interface not to the implementation.

JustinPihony:
Did #1514 fix this, and it can be closed?

PR #1514, opened 2019-03-15
added reset to companion object

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1514)

Issue: https://github.com/LBNL-UCB-STI/beam/issues/1488


carloscaldas:
@colinsheppard I believe it cannot be done after end iteration with current design because we have many listeners and we should have one that executes as last one in order to `reset` the logging. currently we don´t have specific ordering as far as I know

colinsheppard:
@carloscaldas let's instead call reset during startup of each iteration. To do this, make BeamSim also implement IterationStartsListener and then override 

  override def notifyIterationStarts(event: IterationStartsEvent): Unit = ???

and then add the rest here.

PR #1519, opened 2019-03-16
Metrics: Refactored and fixed unit format

REASY:
- Changed unit format from `µs` to `ns`
- Changed scale to log based 2
- Changed order of graphs
- Added tables on the right side of the tables

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1519)


REASY:
Before it looked like:
image

Now it looks like:
image

REASY:
test!

PR #1527, opened 2019-03-18
only use cars for ridehail

wrashid:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1527)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
@@ -297,8 +297,15 @@ class RideHailManager(
   private val initialNumHouseholdVehicles = scenario.getHouseholds.getHouseholds
     .values()
     .asScala
-    .map(_.getVehicleIds.size())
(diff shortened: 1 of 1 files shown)


wrashid:
Test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -297,8 +297,15 @@ class RideHailManager(
   private val initialNumHouseholdVehicles = scenario.getHouseholds.getHouseholds
     .values()
     .asScala
-    .map(_.getVehicleIds.size())
-    .sum / beamServices.beamConfig.beam.agentsim.agents.vehicles.fractionOfInitialVehicleFleet // Undo sampling to estimate initial number
+    .flatMap { hh =>
+      hh.getVehicleIds.asScala.map { vehId =>
```
JustinPihony: Just for future note - this seems like an area where `collect` would have been more apt. Basically anytime you see a map-filter - then think collect instead.

PR #1529, opened 2019-03-18
Fix conversion and leave commented link to lane)

JustinPihony:
Pretty much all comments except the metric fix

I thought about adding this to a config instead of commenting, but let's see if it will be needed again or not. Maybe we'll just delete it in time.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1529)


JustinPihony:
test!

PR #1534, opened 2019-03-18
zn/more ridehail agents

zneedell:
Previously, ride hail agents were limited to one shift per day (with duration estimated based on observed shift duration in the rideAustin data). But, many of the Austin drivers took multiple shifts per day. This meant that the a given number of ride hail agents in Beam would be on-call for a smaller number of hours than the same number of Austin drivers. This fixes that by (in practice) estimating the total number of shifts a given number of real ride-hail agents would work and then generating that many simulated ride hail agents.

Here, the average number of hours worked per day for a ride hail driver is taken from the Austin data (3.52 hours per day, taken only from drivers who took at least 4 trips in a day). This gives the total amount of time a given number of ride hail drivers would be expected to work in a given day. Shifts are then added (with duration drawn from a lognormal distribution with parameters from the Austin dataset) until the total number of ride hail agent hours reaches this value.

For now, there remains a 1:1 correspondence between autonomous ride hail vehicles and input ride hail vehicle fleet size, even though autonomous RH vehicles work all 24 hours while human driven ones work ~3.5 hrs. This means that increased automation means a bigger effective RH fleet size. That is probably reasonable, but worth noting.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1534)


wrashid:
Test!

PR #1536, opened 2019-03-20
zn/generalized time skims

zneedell:
Add columns to skims.csv and skimsExcerpt.csv. Generalized time (in seconds) gives the experienced time due to factors such as mode, automation level, and level of sharing. Generalized cost converts this generalized time into a monetary value using a traveler's value of time and adds the actual cost of the trip.

Involves some refactoring of the VOTT parts in PopulationAttributes to make everything work. This probably could be done more completely by moving most of the time scaling calculations to ModeChoiceCalculator, but I left it out for now to reduce the number of changes.

This also fixes what I think is a bug in warm start when skims are read in--It looks like the hour ID of a skim row was being treated as the travel time.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1536)


zneedell:
Test!

zneedell:
> LGTM but -- if you haven't already -- can you please follow up with some independent validation, i.e. run a couple of beamville scenarios and pick out the resulting skims and ensure they make sense given the VOT params used in the scenario?

Sure--I've done it in a few cases and they made sense. I'll do a couple more and collect them.

PR #1538, opened 2019-03-20
replace append with prepend for lists

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1538)


REASY:
LGTM, but just to make sure it won't break any logic. Changing append to prepend will affect the order:
image
Am I right understand that it's fine for the caller/user of this change?

bhavyalatha26:
> LGTM, but just to make sure it won't break any logic. Changing append to prepend will affect the order:
> image
> Am I right understand that it's fine for the caller/user of this change?

Yes @REASY its fine , as we calculate the sum/average of the values at the end , the order does not affect the use case here , so I made that change.

REASY:
test!

REASY:
test!

PR #1541, opened 2019-03-20
zn/#1539 rh start at activity locations

zneedell:
Give option for ride hail agents to be generated at random activity locations, not just random homes

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1541)


zneedell:
Test!

PR #1547, opened 2019-03-21
zn/production sfbay new energy files

zneedell:
update vehicle energy files

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1547)


JustinPihony:
@zneedell The latest code is most accurate from a vehicle energy calculation perspective. It was only JUST validated - so you may want to pull in the latest from master.

PR #1556, opened 2019-03-25
zn/rh vehicle random seed

zneedell:
fix bug that causes rh manager and scenario loader to repeatedly create the same vehicle type due to random seed being reset

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1556)


zneedell:
Test!

PR #1565, opened 2019-03-26
Fix total cost/incentive summary stat

sfwatergit:
Ensure that semantics of total summary stat matches computation

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1565)


sfwatergit:
Request merge into master; however, may need to limit to competitions-pin_dev until that branch is even w/ origin/master.

PR #1568, opened 2019-03-27
sorting disk of requests before filtering by distance

htmlrb:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1568)


htmlrb:
test!

PR #1580, opened 2019-03-28
fix vehicleMilesTraveled graphs

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1580)

### DIFF
--- a/src/main/java/beam/analysis/summary/VehicleMilesTraveledAnalysis.java
+++ b/src/main/java/beam/analysis/summary/VehicleMilesTraveledAnalysis.java
@@ -15,6 +15,7 @@ public class VehicleMilesTraveledAnalysis implements IterationSummaryAnalysis {
     private Map<String, Double> milesTraveledByVehicleType = new HashMap<>();
     private Set<Id<BeamVehicleType>> vehicleTypes;
     private String humanBodyVehicleType = BeamVehicleType.defaultHumanBodyBeamVehicleType().toString();
+    private final double meterToMileConverterConst = 0.000621371192;  // unit conversion from meters to miles
 
     public VehicleMilesTraveledAnalysis(Set<Id<BeamVehicleType>> vehicleTypes) {
         this.vehicleTypes = vehicleTypes;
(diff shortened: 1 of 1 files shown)


rajnikantsh:
test!

rajnikantsh:
test!

src/main/java/beam/analysis/summary/VehicleMilesTraveledAnalysis.java
```diff
@@ -46,8 +46,7 @@ public void resetStats() {
                 e -> e.getValue() * 0.000621371192 // unit conversion from meters to miles
         ));
 
-        vehicleTypes.foreach(vt -> result.merge("vehicleMilesTraveled_" + vt.toString(),0D, (d1, d2) -> d1 + d2));
-
+        vehicleTypes.foreach(vt -> result.put("vehicleMilesTraveled_" + vt.toString(),milesTraveledByVehicleType.getOrDefault(vt.toString(), 0.0) * 0.000621371192));
```
JustinPihony: We should use a variable for the conversion rate. Especially since it's used more than once. Also, this is quite a change n that it doesn't just change the conversion factor - it also overwrites any old values instead of summing them. That seems incorrect, also?

rajnikantsh: Hi @JustinPihony , on line number https://github.com/LBNL-UCB-STI/beam/blob/e8717bd0687bbccab616e5ce9dba3d20b05cd967/src/main/java/beam/analysis/summary/VehicleMilesTraveledAnalysis.java#L30 we already merging length with vehicle type, here we are just preparing stats param,previously we are always merging 0 (look into merge value id 0D) in this stats params so these stats param value was 0

JustinPihony: Duh - sorry for missing that.

rajnikantsh:
Closing This PR. This issue is resolved in https://github.com/LBNL-UCB-STI/beam/pull/1891

Issue #1571, opened 2019-03-27
output private vehicles assigned to households


wrashid:
create csv in root: housholdVehicles: vehicleId,vehicleType,householdId

PR #1603, opened 2019-03-31
output private vehicles assigned to households #1571

twoface88:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1603)

#1571


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -566,6 +561,21 @@ trait BeamHelper extends LazyLogging {
         scenario.getPopulation.removePerson(personId)
       }
 
+      val csvWriter = new FileWriter("householdVehicles.csv", true)
+      csvWriter.write("vehicleId,vehicleType,householdId\n")
+
+      scenario.getHouseholds.getHouseholds.values.asScala.foreach { householdId =>
+        householdId.getVehicleIds.asScala.foreach { vehicle =>
+          csvWriter.write(
+            beamServices.privateVehicles.get(vehicle).get.id + "," +
```
JustinPihony: You can reduce the reads by storing the `get` in a variable. 

Also, you do a `map...getOrElse` in the second position, but `get` in the first. This should be consistent - You could even just `fold` and output both together?

twoface88:
@JustinPihony please take a look

twoface88:
Test!

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -383,6 +389,26 @@ trait BeamHelper extends LazyLogging {
...
+      csvWriter.write("vehicleId,vehicleType,householdId\n")
+      scenario.getHouseholds.getHouseholds.values.asScala.foreach { householdId =>
+        householdId.getVehicleIds.asScala.foreach { vehicle =>
+          beamServices
+            .privateVehicles
+            .get(vehicle)
+            .map(
+              v => v.id.toString + "," + v.beamVehicleType.id.toString + "," + householdId.getId.toString + "\n"
+            )
+            .map((csvWriter.write(_)))
```
JustinPihony: When the mapped variables match the parameter list then you can just pass the function reference:

    .map(csvWriter.write)

JustinPihony:
test!

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -388,6 +389,25 @@ trait BeamHelper extends LazyLogging {
...
+    try {
+      csvWriter.write("vehicleId,vehicleType,householdId\n")
+      scenario.getHouseholds.getHouseholds.values.asScala.foreach { householdId =>
+        householdId.getVehicleIds.asScala.foreach { vehicle =>
+          beamServices.privateVehicles
+            .get(vehicle)
+            .map(
+              v => v.id.toString + "," + v.beamVehicleType.id.toString + "," + householdId.getId.toString + "\n"
+            )
+            .map((csvWriter.write(_)))
```
REASY: Seems like you don't use return value and do side-effect, so you could consider to use `foreach` instead of `map`

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -388,6 +389,25 @@ trait BeamHelper extends LazyLogging {
     }), scala.concurrent.duration.Duration.Inf)
   }
 
+  def writeScenarioPrivateVehicles(scenario: MutableScenario, beamServices: BeamServices) = {
```
REASY: Please, add return type to be explicit here

REASY:
test!

twoface88:
@REASY please take a look.

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -388,6 +389,25 @@ trait BeamHelper extends LazyLogging {
     }), scala.concurrent.duration.Duration.Inf)
   }
 
+  def writeScenarioPrivateVehicles(scenario: MutableScenario, beamServices: BeamServices): Unit = {
+    val csvWriter: FileWriter = new FileWriter("householdVehicles.csv", true)
```
wrashid: where is this being written out? root folder?

twoface88: @wrashid yes, it writes to root folder. as per requirement here https://github.com/LBNL-UCB-STI/beam/issues/1571

wrashid: it should go to the output folder, not beam root

wrashid:
Test!

wrashid:
Test!

twoface88:
Test!

twoface88:
Test!

PR #1607, opened 2019-04-01
capture all replanning event count

### LINKED ISSUE #1593: debugging: capture replanning chains
print in a csv all replanning chains with count

rajnikantsh:
count replaning events

Closes #1593

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1607)

### DIFF
--- a/src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
+++ b/src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
@@ -133,7 +133,8 @@ public class RealizedModeAnalysis implements GraphAnalysis, MetricsSupport {
             createRootRealizedModeChoosenGraph(referenceDataset,referenceGraphTitle, "# mode choosen(Percent)" , fileName , cumulativeReferenceMode);
         }
 
-        writeToReplaningChainCSV(event);
+        Map<String, Integer> modeCount = calculateModeCount();
+        writeToReplaningChainCSV(event, modeCount);
         writeToRootCSV();
         writeToCSV(event);
         writeToReferenceCSV();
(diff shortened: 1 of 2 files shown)


src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
```diff
@@ -625,6 +630,56 @@ public void writeToReferenceCSV(){
         }
     }
 
+    private void writeToReplaningChainCSV(IterationEndsEvent event) {
+        String fileName = GraphsStatsAgentSimEventsListener.CONTROLLER_IO.getIterationFilename(event.getIteration(), "replanningEventChain.csv");
+        String REPLANNING_SEPARATOR = "-"+ReplanningEvent.EVENT_TYPE+"-";
+        try (BufferedWriter out = new BufferedWriter(new FileWriter(new File(fileName)))) {
```
JustinPihony: This will not close the `FileWriter` resource afaik. You'd have to double try or use an explicit finally I believe.

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
```diff
@@ -625,6 +630,56 @@ public void writeToReferenceCSV(){
...
+            String heading = "modeChoiceReplanningEventChain,count";
+            out.write(heading);
+            out.newLine();
+            Set<String> persons = personReplanningChain.keySet();
+            Map<String, Integer> modeCount = new HashedMap();
+            for(String person: persons){
+                List<String> modes = personReplanningChain.get(person);
+                if(modes.size() > 1){
+                    StringBuffer lastMode = new StringBuffer();
+                    boolean isLastReplaning = false;
```
JustinPihony: Nit: typo :)

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
```diff
@@ -625,6 +630,56 @@ public void writeToReferenceCSV(){
...
+                    for(String mode: modes){
+                        if(ReplanningEvent.EVENT_TYPE.equals(mode)){
```
JustinPihony: Is this supposed to be an if else, or just an if to set the variable? Additionally, is this the best way to tell if it is the last? I guess I don't know this data structuring well enough, but it seems open to a lot of bugs. Would more assertions be beneficial? How about a test or tests to make sure this works as expected?

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
rajnikantsh: used double try to close both streams individually

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
rajnikantsh: Hi @JustinPihony I have simplifid the code. Also last means all modes that come before "Replanning" event. Added couple of test cases to cover this scenerio.

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
rajnikantsh: fix this typo :)

PR #1647, opened 2019-04-09
check if request buffer time is less then scheduler parallelism

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1647)


src/main/scala/beam/utils/scripts/FailFast.scala
```diff
@@ -34,5 +34,11 @@ object FailFast extends LazyLogging {
       )
     }
 
+    if (beamServices.beamConfig.beam.agentsim.agents.rideHail.allocationManager.requestBufferTimeoutInSeconds != 0 &&
+        beamServices.beamConfig.beam.agentsim.schedulerParallelismWindow > beamServices.beamConfig.beam.agentsim.agents.rideHail.allocationManager.requestBufferTimeoutInSeconds) {
+      throw new RuntimeException(
+        "Scheduler Parallelism Window must be greater then Request Buffer Timeout"
```
wrashid: must be smaller than

rajnikantsh: change comparision from greater then to less then

rajnikantsh:
test!

PR #1671, opened 2019-04-13
Events: Added linkTravelTime to attributes of PathTraversalEvent

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1671)


REASY:
@JustinPihony I think it is ok to merge it. The output file size won't be changed so dramatically.

PR #1716, opened 2019-04-19
TravelTime: Use 2-d array (instead of hashmap) to store travel time

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1716)


REASY:
`Map<Id<Link>, double[]> _linkIdToTravelTimeArray` is map from `linkId` to travel time array which contains travel time for hours from 0 to 30, so I'm not sure how we could use just single value (we could use pair (Link, Int) as a key of map <linkId, hour>. 

The main purpose of this change is to get performance benefit (hashmap is fast for lookup, but still is not faster than random access in the array (we use linkId as an offset in array).  `getLinkTravelTime` is called very often during routing calculation (https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/scala/beam/router/r5/R5RoutingWorker.scala#L930 is passed to R5 https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/scala/beam/router/r5/R5RoutingWorker.scala#L998)

REASY:
Also it reduces the memory around 14%. `_map` is original map, its size is 30529216 Bytes, `_data` is 2D-array, its size is 26034776.

image

JustinPihony:
Sure, but could the memory difference simply be due to one using Id<Link> and the other using Int?

REASY:
I don't think that it is because of Id. I used http://openjdk.java.net/projects/code-tools/jol/ to get the total size of map and 2-D array:
- map: 33370688 bytes
- 2-d array: 24503296 bytes

PR #1766, opened 2019-04-29
Avoid scenario loaded twice 4ci

carloscaldas:
Ccaldas/#1753 a

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1766)

Main changes:
- Warmstart was executed before loading specific configuration. 
- General refactoring: extracted methods and made things clear
- Script loaded twice: first Matsim config xml files are loaded and then the values are replaced by specific parameters (like Urbasim). This was not completely solved due to highly coupling but at least not is clear this behavior and some redundant steps were removed


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -460,7 +478,31 @@ trait BeamHelper extends LazyLogging {
...
+        matsimScenario
+      }
+    } else {
+      matsimScenario
+    }
+  }
+
+  def setupBeamWithConfig(
+    config: TypesafeConfig
+  ): (BeamConfig, org.matsim.core.config.Config, String, NetworkCoordinator) = {
```
REASY: Let's extract it to case class return type. Should be easier to read and maintain

PR #1770, opened 2019-05-01
Imperative way to `processLinkEvents`

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1770)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
@@ -17,7 +17,7 @@ import beam.agentsim.scheduler.BeamAgentScheduler.{CompletionNotice, ScheduleTri
 import beam.agentsim.scheduler.Trigger
 import beam.agentsim.scheduler.Trigger.TriggerWithId
 import beam.router.Modes.BeamMode
-import beam.router.Modes.BeamMode.{BIKE, RIDE_HAIL, TRANSIT, WALK}
+import beam.router.Modes.BeamMode.{TRANSIT, WALK}
 import beam.router.model.BeamLeg
 import beam.router.osm.TollCalculator
 import beam.sim.HasServices
@@ -30,6 +30,7 @@ import org.matsim.api.core.v01.events.{
   VehicleLeavesTrafficEvent
 }
 import org.matsim.api.core.v01.population.Person
+import org.matsim.core.api.experimental.events.EventsManager
 import org.matsim.vehicles.Vehicle
 
 import scala.collection.mutable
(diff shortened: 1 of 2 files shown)


REASY:
Run branch production-sfbay, `production/sfbay/smart/smart-baseline.conf` + `beam.agentsim.agentSampleSizeAsFractionOfPopulation = 0.02`

## Without this commit:
image
image

## With this commit:
image

image

REASY:
Need to add test to make sure if future changes happens it will be caught

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -76,6 +77,24 @@ object DrivesVehicle {
...
   case class StopDrivingIfNoPassengerOnBoardReply(success: Boolean, requestId: Int, tick: Int)
 
+  def processLinkEvents(eventsManager: EventsManager, vehicleId: Id[Vehicle], leg: BeamLeg): Unit = {
+    val path = leg.travelPath
+    if (path.linkTravelTime.nonEmpty & path.linkIds.size > 1) {
+      val links = path.linkIds
+      val linkTravelTime = path.linkTravelTime
+      var i: Int = 0
+      var curTime = leg.startTime
+      while (i < links.length - 1) {
```
JustinPihony: Should we keep the comment about why we are dropping the last item? Or methodize it in some way? Maybe that's where a test could come in handy?

JustinPihony:
Should this be merged? And an issue can be created to add a test?

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
REASY: Let me bring the comment back. Here is some examples, hope it is helpful:

If you have `linkIds = Array(1, 2, 3)` and `linkTravelTime = Array(1, 2, 3)`, then the following events will be generated:
```
- LinkLeaveEvent at time 1.0, linkId = 1

- LinkEnterEvent at time 1.0, linkId = 2
- LinkLeaveEvent at time 3.0, linkId = 2

- LinkEnterEvent at time 3.0, linkId = 3
```
If you have `linkIds = Array(1, 2, 3, 4)` and `linkTravelTime = Array(1, 2, 3, 4)`:
```
- LinkLeaveEvent at time 1.0, linkId = 1

- LinkEnterEvent at time 1.0, linkId = 2
- LinkLeaveEvent at time 3.0, linkId = 2

- LinkEnterEvent at time 3.0, linkId = 3
- LinkLeaveEvent at time 6.0, linkId = 3

- LinkEnterEvent at time 6.0, linkId = 4
```
What I want to show is that we have to start with `LinkLeaveEvent` and stop with `LinkEnterEvent`. Because we stop with `LinkEnterEvent`, so the whole travel time through that link (last link) is dropped. Time is cumulative here. (It seems to me it is MATSim specific. @wrashid @colinsheppard? )

PR #1771, opened 2019-05-01
New R5 version

### LINKED ISSUE #1684: RoutingResponse contains result in different coordinate system
Example of [RoutingRequest](https://pastebin.com/69jDAJZ0) which gives [RoutingResponse](https://pastebin.com/quRfKyy7) in different coordinate system.

### LINKED ISSUE #1689: update r5

### LINKED ISSUE #1743: R5 Network loading from directory must give some order to files to avoid discrepancy in link ids
[TransportNetwork fromDirectory](
https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/transit/TransportNetwork.java#L188) uses [File.listFiles()](https://docs.oracle.com/javase/7/docs/api/java/io/File.html#listFiles()) which is:

> There is no guarantee that the name strings in the resulting array will appear in any specific order; they are not, in particular, guaranteed to appear in alphabetical order.

So we have to make sure that there is some order in filenames to be able to use link ids across different operating systems/JVMs and use [TransportNetwork fromFiles](https://github.com/conveyal/r5/blob/dev/src/main/java/com/conveyal/r5/transit/TransportNetwork.java#L178)

michaz:
~(Can't CI this at the moment. Writing this up anyway.)~

Fixes #1689.
Fixes #1684.
Fixes #1743.

This has Beam use a new R5 version based on Conveyal's current `master`.

New R5 customization compared to the previous version:
- R5 now by default doesn't link (let trips start and end) on one-way streets. My guess is because they mainly do grid-based analysis, where those trips may be unrepresentative of a grid cell. I reverted that, mainly because it would destroy Beamville.

Changed behavior of R5 I had to adapt to:
- The `TransitMode` `TRANSIT`, which used to mean "any mode", doesn't mean that anymore. We rather have to pass the full set of TransitModes. On the way, I noticed that we only ever want "transit" or "no transit", so I removed the passing-around of this set for now and replaced it with a Bool parameter.
- `ProfileRequest` now needs a car speed, which needs to be an overestimation, or the A* heuristic will fail and we will miss attractive routes. Put 130 km/h there.

Other thing I noticed/changed on the way:
- `createPhyssimNetwork` is meanwhile happening before `initTransit`, but `initTransit` would sneakily modify the R5 network to insert dummy links for unlinked transit stops. This can't work. Those two can't mutually depend on each other. Removed the latter. We can't work with unlinked transit stops, our scenarios (or our test cases) should not contain them, R5 doesn't have a model for what to do with them, we shouldn't interpolate that. We still produce dummy _coordinates_ for them because transit drivers need them and our test cases _do_ contain unlinked stops.
- Added an assertion about being in WGS coordinates to `BeamPath` constructor. Fixed several places in code where coordinates were just wrong.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1771)


REASY:
@michaz is it possible to address this issue while updating R5? https://github.com/LBNL-UCB-STI/beam/issues/1743

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
Fixes #1684

michaz:
Note: Copy/pasted custom version of `McRaptorSuboptimalPathProfileRouter` was removed, I'm using the original here instead.

michaz:
This value

```
    // Warning: carSpeed is not used for link traversal (rather, the OSM travel time model is used),
    // but for R5-internal bushwhacking from network to coordinate, AND ALSO for the A* remaining weight heuristic,
    // which means that this value must be an over(!)estimation, otherwise we will miss optimal routes,
    // particularly in the presence of tolls.
    profileRequest.carSpeed = 36.11f // 130 km/h, WARNING, see ^^ before changing
```

has potential performance impact, may try to reduce if something is slow now, but may lead to suboptimal routes.

michaz:
Is this mergeable, now that the performance seems okay?

REASY:
Performance comparison:
https://docs.google.com/spreadsheets/d/1_Z8X1PJKynKumyjNQdtVLRB4-0_xfcL_ofYiaOyjTRI/edit?usp=sharing

PR #1784, opened 2019-05-04
Added test for CSV read/write and small refactoring

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1784)

### DIFF
--- a/src/main/scala/beam/router/BeamSkimmer.scala
+++ b/src/main/scala/beam/router/BeamSkimmer.scala
@@ -446,7 +446,7 @@ class BeamSkimmer @Inject()(val beamConfig: BeamConfig, val beamServices: BeamSe
                   }
 
                 writer.write(
-                  s"$timeBin,$mode,${origin.tazId},${destination.tazId},${theSkim.time},${theSkim.generalizedTime},${theSkim.cost},${theSkim.generalizedTime},${theSkim.distance},${theSkim.count},${theSkim.energy}\n"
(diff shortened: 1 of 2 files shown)


src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -390,95 +390,97 @@ class BeamSkimmer @Inject()(val beamConfig: BeamConfig, val beamServices: BeamSe
...
+  //  }
 
   def writeObservedSkims(event: IterationEndsEvent): Unit = {
-    val fileHeader =
-      "hour,mode,origTaz,destTaz,travelTimeInS,generalizedTimeInS,cost,generalizedCost,distanceInM,numObservations,energy"
     val filePath = event.getServices.getControlerIO.getIterationFilename(
       event.getServices.getIterationNumber,
       BeamSkimmer.observedSkimsFileBaseName + ".csv.gz"
     )
+    //    val writer = IOUtils.getBufferedWriter(filePath)
```
REASY: Commented code

src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -390,95 +390,97 @@ class BeamSkimmer @Inject()(val beamConfig: BeamConfig, val beamServices: BeamSe
...
+    } finally {
+      writer.close()
     }
-    writer.close()
   }
+
 }
 
 object BeamSkimmer extends LazyLogging {
+  type BeamSkimmerKey = (Int, BeamMode, Id[TAZ], Id[TAZ])
```
REASY: I know you just refactored, but what about extracting it as case class instead of Tuple4?

carloscaldas: I like the idea, but don want to change too much in this PR since its purpose was only to create a test. Small refactoring was only to avoid duplication code

PR #1795, opened 2019-05-09
Fix scheduler parallelism window and request buffer timeout condition

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1795)

### DIFF
--- a/src/main/scala/beam/utils/scripts/FailFast.scala
+++ b/src/main/scala/beam/utils/scripts/FailFast.scala
@@ -35,9 +35,9 @@ object FailFast extends LazyLogging {
     }
 
     if (beamServices.beamConfig.beam.agentsim.agents.rideHail.allocationManager.requestBufferTimeoutInSeconds != 0 &&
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/utils/scripts/FailFast.scala
```diff
@@ -35,9 +35,9 @@ object FailFast extends LazyLogging {
     }
 
     if (beamServices.beamConfig.beam.agentsim.agents.rideHail.allocationManager.requestBufferTimeoutInSeconds != 0 &&
-        beamServices.beamConfig.beam.agentsim.schedulerParallelismWindow < beamServices.beamConfig.beam.agentsim.agents.rideHail.allocationManager.requestBufferTimeoutInSeconds) {
+        beamServices.beamConfig.beam.agentsim.schedulerParallelismWindow > beamServices.beamConfig.beam.agentsim.agents.rideHail.allocationManager.requestBufferTimeoutInSeconds) {
```
wrashid: should be < because that would be the violation case

PR #1701, opened 2019-04-18
Added populationConverter script

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1701)

This PR contains:

CsvWriters from memory (Scenario/BeamServices):
- PopulationWriterCSV
- VehiclesCsvWriter
- HouseholdsCsvWriter
- PlansCsvWriter
- NetworkNodeCsvWriter
- NetworkLinkCsvWriter

Converter from Xml to Csv:

-   VehiclesXml2CsvConverter
-   PopulationXml2CsvConverter
-   PhyssimEventsXmlToCSV
-   HouseholdsXml2CsvConverter
-   PansXml2CsvConverter
-   NetworkXmlToCSV

PR #1804, opened 2019-05-12
Create CSV writers and conversion scripts

carloscaldas:
CsvWriters from memory (Scenario/BeamServices):
- PopulationCsvWriter
- VehiclesCsvWriter
- HouseholdsCsvWriter
- PlansCsvWriter
- NetworkNodeCsvWriter
- NetworkLinkCsvWriter
- NetworkMergeCsvWriter

Converter from Xml to Csv:
- PopulationXml2CsvConverter
- PhyssimEventsXmlToCSV
- HouseholdsXml2CsvConverter
- PansXml2CsvConverter
- NetworkXmlToCSV

Previous PR: https://github.com/LBNL-UCB-STI/beam/pull/1701

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1804)


carloscaldas:
test!

carloscaldas:
test!

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -291,17 +288,33 @@ class BeamSim @Inject()(
...
     BeamConfigChangesObservable.clear()
 
   }
 
+  private def writeScenario(scenario: Scenario, controlerIO: OutputDirectoryHierarchy) = {
+    PopulationCsvWriter.toCsv(scenario, controlerIO.getOutputFilename("population.csv"))
+    VehiclesCsvWriter(beamServices).toCsv(scenario, controlerIO.getOutputFilename("vehicles.csv"))
+    HouseholdsCsvWriter.toCsv(scenario, controlerIO.getOutputFilename("households.csv"))
+    PlansCsvWriter.toCsv(scenario, controlerIO.getOutputFilename("plans.csv"))
+    NetworkNodeCsvWriter.toCsv(scenario, controlerIO.getOutputFilename("networkNode.csv"))
```
wrashid: we only need the network (rename networkMerge.csv to network.csv). networkNode.csv and networkLink.csv are not needed

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -291,17 +288,33 @@ class BeamSim @Inject()(
...
       logger.debug(s"deleting output file: $fileName")
-      Files.deleteIfExists(Paths.get(event.getServices.getControlerIO.getOutputFilename(fileName)))
+      Files.deleteIfExists(Paths.get(controllerIO.getOutputFilename(fileName)))
     }
     BeamConfigChangesObservable.clear()
 
   }
 
+  private def writeScenario(scenario: Scenario, controlerIO: OutputDirectoryHierarchy) = {
+    PopulationCsvWriter.toCsv(scenario, controlerIO.getOutputFilename("population.csv"))
```
wrashid: we have config parameters (intervals) for deciding, when to write these out. can they be used here as well?

src/main/scala/beam/sim/BeamSim.scala
carloscaldas: updated

src/main/scala/beam/sim/BeamSim.scala
carloscaldas: as you can see in this line it is checked after every iteration ends
https://github.com/LBNL-UCB-STI/beam/pull/1804/files#diff-daac2f006f1a8b03e4c5a570b8b1ade7R189

IMO this should be responsibility of the caller. The executor should just write the scenario if asked for. It is easy to change... just tell me if I should keep the way it is or change

src/main/scala/beam/utils/csv/conversion/PlansXml2CsvConverter.scala
```diff
@@ -0,0 +1,99 @@
...
+    mode: String,
+    elementType: String
+  )
+
+  private def toActivity(node: Node, index: Int): ActivityOrLeg = {
+    ActivityOrLeg(
+      activityIndex = index,
+      activityType = Option(node.attributes("type")).map(_.toString.trim).getOrElse(""),
+      locationX = Option(node.attributes("x")).map(_.text.toDouble).getOrElse(0),
+      locationY = Option(node.attributes("x")).map(_.text.toDouble).getOrElse(0),
```
REASY: Typo here `locationY = Option(node.attributes("x")` should be `locationY = Option(node.attributes("y")`

src/main/scala/beam/utils/csv/conversion/VehiclesXml2CsvConverter.scala
```diff
@@ -0,0 +1,75 @@
...
+  override def contentIterator(sourceFile: File): Iterator[String] = {
+    val parser = ConstructingParser.fromFile(sourceFile, preserveWS = true)
+    val doc = parser.document().docElem
+    val peopleNodes: NodeSeq = doc \ "household"
+
+    //    val memberToHousehold = readMemberToHousehold()
+    //    val memberToRank: MemberToRank = readMember2Rank()
+
+//    peopleNodes.toIterator.map(node => toPerson(node, memberToHousehold, memberToRank).toString + LineSeparator)
+    Iterator()
```
REASY: Do you need empty iterator here? If yes, what all above things are doing?

src/main/scala/beam/utils/csv/conversion/Xml2CsvFileConverter.scala
```diff
@@ -0,0 +1,19 @@
+package beam.utils.csv.conversion
+
+import java.io.File
+
+trait Xml2CsvFileConverter {
+  protected def fields: Seq[String]
+  protected lazy val header: Iterator[String] = Iterator(fields.mkString(FieldSeparator), LineSeparator)
```
REASY: Why don't leave it as `def header`? Then you don't need to do the trick with `lazy val` because of init order

src/main/scala/beam/utils/csv/conversion/XmlConverter.scala
```diff
@@ -0,0 +1,114 @@
...
+    FileUtils.writeToFile(
+      csvFile.getAbsolutePath,
+      f(file)
+    )
+    csvFile
+  }
+
+  def extractFile(gzipFile: File): File = {
+    val result = File.createTempFile("somePrefix", ".xml")
+    val outputStream = new FileOutputStream(result)
```
REASY: If exception happens, the stream will remain open. Better to use `try catch`, or even better to use `IOUtils.getBufferedReader` which knows about gzipped file and create GZipStream on fly, example of usage [NetworkXmlToCSV.scala#L19](https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/scala/beam/utils/NetworkXmlToCSV.scala#L19)

src/main/scala/beam/sim/BeamSim.scala
wrashid: yes, but there are different parameters, which control different files

src/main/scala/beam/utils/csv/conversion/PlansXml2CsvConverter.scala
carloscaldas: nice catch

src/main/scala/beam/sim/BeamSim.scala
carloscaldas: Executing before the first iteration:
```
    PopulationCsvWriter.toCsv(scenario, controllerIO.getOutputFilename("population.csv"))
    VehiclesCsvWriter(beamServices).toCsv(scenario, controllerIO.getOutputFilename("vehicles.csv"))
    HouseholdsCsvWriter.toCsv(scenario, controllerIO.getOutputFilename("households.csv"))
    NetworkCsvWriter.toCsv(scenario, controllerIO.getOutputFilename("network.csv"))
```

Executing after the end of each iteration:
```Scala
      if (interval > 0 && iterationNumber % interval == 0) {
        PlansCsvWriter.toCsv(scenario, controllerIO.getIterationFilename(iterationNumber, "plans.csv"))
      }
```

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -710,6 +693,29 @@ trait BeamHelper extends LazyLogging {
     } else throw new NotImplementedError(s"ScenarioSource '$src' is not yet implemented")
   }
 
+  private def buildUrbansimScenarioSource(
+    injector: _root_.com.google.inject.Injector,
```
REASY: Please, fix the types

carloscaldas: done

PR #1805, opened 2019-05-12
Tools used to investigate travel time issues

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1805)


src/main/scala/beam/router/TravelTimeObserved.scala
```diff
@@ -166,6 +166,8 @@ object TravelTimeObserved extends LazyLogging {
   val histogramName: String = "simulation_vs_reference_histogram.png"
   val histogramBinSize: Int = 200
 
+  val MaxDistanceFromBeamTaz: Double = 500.0 // 500 meters
```
REASY: @wrashid some time back we discussed this (it's about how far TAZ from provided data are from our TAZs) . Is it fine if it goes to develop?

REASY:
test!

src/main/scala/beam/agentsim/events/PathTraversalEvent.scala
```diff
@@ -215,7 +215,7 @@ object PathTraversalEvent {
     val arrivalTime: Int = attr(ATTRIBUTE_ARRIVAL_TIME).toInt
     val mode: BeamMode = BeamMode.fromString(attr(ATTRIBUTE_MODE)).get
     val legLength: Double = attr(ATTRIBUTE_LENGTH).toDouble
-    val linkIdsAsStr = attr(ATTRIBUTE_LINK_IDS)
+    val linkIdsAsStr = Option(attr(ATTRIBUTE_LINK_IDS)).getOrElse("")
```
JustinPihony: What's the point of this? To handle a null from `attr`? If so wouldn't it be better to change `attr` or just do a null check?

src/main/scala/beam/utils/map/GeoJsonToGpxConvertor.scala
```diff
@@ -2,29 +2,42 @@ package beam.utils.map
...
-    val end = System.currentTimeMillis()
-    logger.info(s"Converted '$inputPath' to '$outputPath' in ${end - start} ms")
+    GeoJsonReader.read(inputPath, mapper)
   }
 
   def main(args: Array[String]): Unit = {
-    convert("""C:\temp\movement_data\san_francisco_censustracts.json""", "san_francisco_censustracts.gpx")
-    convert("""C:\temp\movement_data\san_francisco_taz.json""", "san_francisco_taz.gpx")
+    {
+      val censusTractsPoints = readGeoJson("""C:\temp\movement_data\san_francisco_censustracts.json""")
```
JustinPihony: Should this be part of the final merge? Maybe it could be read from args with a default to something meaningful? Or a helpful hint for those who might try to use it.

src/main/scala/beam/agentsim/events/PathTraversalEvent.scala
REASY: Yes, it is to make sure that there will be no NPE. It's common way to me to interact with java code when `null` is possible by just wrapping it to `Option`. I thought it's common knowledge. I can add comments why it's done this way

src/main/scala/beam/utils/map/GeoJsonToGpxConvertor.scala
REASY: Let me make it as app args with some comments

src/main/scala/beam/router/TravelTimeObserved.scala
wrashid: sure

REASY:
test!

src/main/scala/beam/agentsim/events/PathTraversalEvent.scala
JustinPihony: Nah - I'm aware of it - just wasn't sure if we wanted to make changes deeper. I'm fine with this though :)

Issue #1389, opened 2019-03-04
Find a better way to handle disrupted CAV trip


colinsheppard:
Sometimes a person missing the pickup (and sometimes the dropoff as well) of a HH CAV. This happens if the skims are misaligned with routes or if a person takes a slow mode on a previous trip (more common).

TheBoard/Alight triggers would cause stuckness, so the current fix is to just reschedule them to happen later... but this of course allows the person to actually take that CAV trip even though it has already left.

So better solution would be to abandon the trip for that person. 

I think easiest way to do this is to send a message cancelling all future CAV trips for that person and then reset the person's state data to no longer assume a CAV is coming (i.e. currentTourMode and _experiencedBeamPlan).

This will be easy to do from when(Moving) in PersonAgent but harder to do from when(Driving) in DrivesVehicle b/c it will have to transfer back over to PersonAgent (create a special state for this and use the timeout=0 trick).

Must pay careful attention to how this might impact the CAV plan. It can still do the same route it originally planned, but without passenger. Should work but needs to be tested.

PR #1344, opened 2019-02-26
remove all default vehicle types

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1344)

PR #1815, opened 2019-05-14
Use speed in vehicle type as maximum speed for router

michaz:
This affects a lot of files (sorry), mostly because it required changing a lot of parameter passing: Every piece of code that used global default vehicle types needs the `vehicleTypes` container now.

- Global vehicle types that are not in the data container are gone.
- There is only *one* vehicle type used in the core that has a special meaning: The default (and only) body type. To make it clear that this is a special vehicle type _that must still be in the data file_, its name is now a config parameter. (Preparing for the possibility that people get different bodies with different maxspeeds.)
- `Car` (as vehicle type id) does _not_ have a special role. There is no default car. Every place that creates vehicles has a configurable type (or must have one).
- Router uses speed on vehicle type as maximum velocity. For `CAR`, the speed is now `min(linkspeed, vehicletypespeed)`, for `BIKE` it is just `vehicletypespeed`. And I think that is the only difference between `CAR` and `BIKE`. That distinction is made by R5.
- Transit drivers at last don't live in the router anymore but in the agentsim.
- I started moving that and a couple other data containers from `BeamServices` to a `BeamScenario` which you can still get from `BeamServices` -- called it that due to the similarity to Matsim's `Scenario`. It complicates things if you mix hard-to-set-up infrastructure (like a `TollCalculator`) with simple data classes on the same big object. I was able to remove at least 4 byzantine mocks and overrides from tests now. :-)
- Removed tons of cruft, some of it from very central locations, which clearly people already spent time on refactoring, and which was either unused, or which dissolved upon pulling the right thread.
- Refactored CAV: It now behaves like transit, i.e. the agent has to "reserve", rather meaning "board", the vehicle. If the agent changes its mind for some reason, the CAV will just drive without her.

*The body type must be in the vehicle type file!*

NB: There are lots of hard-coded vehicle *type ids* remaining, mostly in peripheral code. I deliberately refrained from letting those use global constants, because that would look like that would work and should be copied.
I can't remove them all because they are used with different intentions -- sometimes indeed as a default, but often also as a dummy whose value never plays a role, or as a key for a map with a special meaning.

Please relay to everyone who doesn't read GitHub that hard-coded vehicle type ids should be removed on sight, except from tests. (Tests know which vehicle types the scenario has.)

Would make sense to run a test scenario that doesn't have a `Car` vehicle type to catch some of them, I guess.

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1815)


michaz:
Will need to address #1344 here.

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

REASY:
@michaz regarding extracting `BeamScenario`, we're fine that it will eventually become as `BeamServices`, meaning knowing too much. The main idea is just to be able to pass needed params by simply creating case class instead of _juggling_ with mocks, correct?

michaz:
@REASY 

> The main idea is just to be able to pass needed params by simply creating case class instead of juggling with mocks, correct?

Yes. Global-ish data containers are okay, as long as they are effectively read-only after a certain cut-off time. Like, when the agentsim starts. I think that's the case now.

But we should try to be more minimalist/orthogonal there. No need for several different representations of the same thing (I removed one thing that was effectively already contained in one other thing. And I think `privateVehicles` is just a convenience representation of something that's already somewhere else. That's dangerous, because one never knows if it's consistent.



Not okay is a wild mixture of

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
Test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

michaz:
test!

REASY:
test!

REASY:
test!

REASY:
test!

REASY:
test!

michaz:
test!

michaz:
Fixes #1389. CAV is now like transit. Agent has to explicitly board.

michaz:
test!

michaz:
test!

michaz:
test!

PR #1818, opened 2019-05-17
zn/new rh cost params

zneedell:
Updates the method for estimating solo and shared rh prices based on the chicago dataset. Previously it was a time and distance component for solo RH and a multiplier for pooled, now there is a separate intercept, distance, and time factor for both solo and pooled.

Also fixes a couple of small bugs in the skimmer

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1818)


src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -851,23 +856,34 @@ class RideHailManager(
     request: RideHailRequest,
     rideHailVehicleTypeId: Id[BeamVehicleType],
     trip: PassengerSchedule,
-    baseFare: Double
+    additionalCost: Double
   ): Map[Id[Person], Double] = {
-    val timeFare = DefaultCostPerSecond * surgePricingManager
+    var CostPerSecond = 0.0
```
JustinPihony: Can you change the naming to be camelCase since these are not constants?

src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -46,7 +46,7 @@ class BeamSkimmer @Inject()(val beamConfig: BeamConfig, val beamServices: BeamSe
   // The OD/Mode/Time Matrix
   private var previousSkims: BeamSkimmerADT = initialPreviousSkims()
   private var skims: BeamSkimmerADT = TrieMap()
-  private val modalAverage: TrieMap[BeamMode, SkimInternal] = TrieMap()
+  private val modalAverage: TrieMap[BeamMode, SkimInternal] = TrieMap() //TODO: Doesn't appear to be used
```
JustinPihony: I'd say to just remove this rather than leave a dangling comment

src/main/scala/beam/router/BeamSkimmer.scala
```diff
@@ -154,15 +175,8 @@ class BeamSkimmer @Inject()(val beamConfig: BeamConfig, val beamServices: BeamSe
...
-              time = 1.1,
-              generalizedTime = 1.1,
-              beamServices.beamConfig.beam.agentsim.agents.rideHail.pooledToRegularRideCostRatio * 1.1,
-              distance = 0,
-              beamServices.beamConfig.beam.agentsim.agents.rideHail.pooledToRegularRideCostRatio,
-              count = 0,
-              energy = 1.1
-            )
+            val cost = getRideHailCost(RIDE_HAIL_POOLED, solo.distance, solo.time)
+            return (1.1, cost / solo.cost)
```
JustinPihony: `return`s are dangerous in scala - Can you rework this to not use it? Ping me if you'd like to discuss ideas on how to do it.

zneedell: Thanks for these!

zneedell:
Test!

PR #1821, opened 2019-05-18
Use `Option` to make sure NPE won't happen

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1821)

### DIFF
--- a/src/main/scala/beam/utils/csv/writers/PopulationCsvWriter.scala
+++ b/src/main/scala/beam/utils/csv/writers/PopulationCsvWriter.scala
@@ -28,12 +28,17 @@ object PopulationCsvWriter extends ScenarioCsvWriter {
       val customAttributes: AttributesOfIndividual =
         person.getCustomAttributes.get("beam-attributes").asInstanceOf[AttributesOfIndividual]
 
-      val excludedModes = personAttributes
-        .getAttribute(person.getId.toString, "excluded-modes")
-        .toString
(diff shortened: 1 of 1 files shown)


REASY:
test!

REASY:
test!

REASY:
test!

wrashid:
Test!

src/main/scala/beam/utils/csv/writers/PopulationCsvWriter.scala
```diff
@@ -28,12 +28,17 @@ object PopulationCsvWriter extends ScenarioCsvWriter {
...
+      val excludedModes = Option(
+        personAttributes
+          .getAttribute(person.getId.toString, "excluded-modes")
+      ).map { attrib =>
+          attrib.toString
+            .replaceAll(",", ArrayItemSeparator)
+            .split(ArrayItemSeparator)
+            .mkString(ArrayStartString, ArrayItemSeparator, ArrayEndString)
+        }
+        .getOrElse("")
```
JustinPihony: Is an empty string ok though? I at least wanted to raise that, in case an exception would be better?

carloscaldas: the exception was there before (NPE) :-)
Agree it should at least be logged

PR #1829, opened 2019-05-19
extracted class - argumentsParser

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1829)

### DIFF
--- a/src/main/scala/beam/sim/ArgumentsParser.scala
+++ b/src/main/scala/beam/sim/ArgumentsParser.scala
@@ -1,0 +1,104 @@
+package beam.sim
+
+import beam.utils.BeamConfigUtils
+import com.typesafe.config.{Config => TypesafeConfig}
+import scopt.OptionParser
+
+object ArgumentsParser {
+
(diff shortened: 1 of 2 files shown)


carloscaldas:
test!

src/main/scala/beam/sim/ArgumentsParser.scala
```diff
@@ -0,0 +1,104 @@
...
+          else success
+      )
+    }
+  }
+
+  def parseArguments(parser: OptionParser[Arguments], args: Array[String]): Option[Arguments] = {
+    parser.parse(args, init = Arguments())
+  }
+
+  def parseArguments(args: Array[String]): Option[Arguments] = parseArguments(buildParser, args)
```
JustinPihony: This seems to be the only public method. Maybe we should hide the rest unless it becomes necessary to expose it?

carloscaldas: done

carloscaldas:
test!

PR #1838, opened 2019-05-20
Sync up with current lambda in AWS

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1838)


JustinPihony:
I just checked and the lambdas in develop are up to date, so this is our of date anyway

PR #1842, opened 2019-05-21
Add Reason for replanning and Generate Repanning Reason count CSV

### LINKED ISSUE #1840: add reason for replanning to replanningEvent -> investigate source replanning in sioux_faux

rajnikantsh:
Closes #1840 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1842)

### DIFF
--- a/src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
+++ b/src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
@@ -524,17 +524,14 @@ public class RealizedModeAnalysis extends BaseModeAnalysis {
     public void writeReplanningReasonCountCSV(Integer iteration) {
         String fileName = GraphsStatsAgentSimEventsListener.CONTROLLER_IO.getIterationFilename(iteration,"replanningEventReason.csv");
         String header = "ReplanningReason,Count";
-        try{
-            FileWriter writer = new FileWriter(new File(fileName));
-            BufferedWriter out = new BufferedWriter(writer);
+        try(FileWriter writer = new FileWriter(new File(fileName));
+            BufferedWriter out = new BufferedWriter(writer)){
             out.write(header);
             out.newLine();
             for (Map.Entry<String,Integer> entry : replanningReasonCount.entrySet()){
                 out.write(entry.getKey()+","+entry.getValue());
                 out.newLine();
             }
-            out.close();
-            writer.close();
         } catch (IOException ex) {
             log.error("error in generating csv ", ex);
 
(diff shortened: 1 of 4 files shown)


rajnikantsh:
Test!

rajnikantsh:
Test!

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
```diff
@@ -516,6 +521,26 @@ public void writeToReferenceCSV() {
...
+        try{
+            FileWriter writer = new FileWriter(new File(fileName));
+            BufferedWriter out = new BufferedWriter(writer);
+            out.write(header);
+            out.newLine();
+            for (Map.Entry<String,Integer> entry : replanningReasonCount.entrySet()){
+                out.write(entry.getKey()+","+entry.getValue());
+                out.newLine();
+            }
+            out.close();
```
JustinPihony: These should use the disposal pattern for safety

````
try(FileWriter writer = new FileWriter(new File(fileName));
  BufferedWriter out = new BufferedWriter(writer)) {
...
} catch ...
````

File: src/main/scala/beam/agentsim/agents/PersonAgent.scala [hunk 2]
```diff
@@ -825,6 +836,17 @@ class PersonAgent(
       }
   }
 
+  def replanningReasonBuilder(data: BasePersonData, prefix: String): String = {
```
JustinPihony: The name doesn't give the full definition. This builds AND outputs, so I would suggest something more like 

```suggestion
  def getReplanningReasonFrom(data: BasePersonData, prefix: String): String = {
```

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -825,6 +836,17 @@ class PersonAgent(
...
+    val reasonBuilder = new StringBuilder(prefix)
```
JustinPihony: How about this for terseness (especially since `None` has no effect and it is minimal concatenation):

```
(data.map{
  case mode if mode.isRideHail => s"$prefix RideHail"
  case mode if mode.isMassTransit || mode.isTransit => s"$prefix PT"
}).getOrElse(prefix)
```

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
rajnikantsh: this is done

src/main/scala/beam/agentsim/agents/PersonAgent.scala
rajnikantsh: method renaming is done

src/main/scala/beam/agentsim/agents/PersonAgent.scala
rajnikantsh: this refracting is done

rajnikantsh:
Test!

wrashid:
Test!

JustinPihony:
test!

JustinPihony:
test!

PR #1870, opened 2019-05-31
add FSM UML

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1870)


JustinPihony:
As this is documentation that seems fine I'm ok with it - not sure if @colinsheppard or @wrashid have any opinion before merging this in?

PR #1872, opened 2019-06-01
fixed csv reader - added missing household attributes

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1872)


src/main/scala/beam/utils/scenario/ScenarioSource.scala
```diff
@@ -3,6 +3,6 @@ package beam.utils.scenario
 trait ScenarioSource {
   def getPersons: Iterable[PersonInfo]
   def getPlans: Iterable[PlanElement]
-  def getHousehold: Iterable[HouseholdInfo]
+  def getHousehold(): Iterable[HouseholdInfo]
```
REASY: Why parenthesis here?

PR #1882, opened 2019-06-05
Add ridehail human charging

JustinPihony:
This branch is based off of jhrf/#1127-ev-charging and may end up being carved off in a smaller PR in the end.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1882)


src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -382,6 +424,26 @@ class RideHailAgent(
...
+    val beta3 = 0.001
+    val mnlUtilityFunction: Map[String, Map[String, UtilityFunctionOperation]] = Map(
+      "ParkingSpot" -> Map(
+        "energyPriceFactor" -> UtilityFunctionOperation("multiplier", -beta1),
+        "distanceFactor"    -> UtilityFunctionOperation("multiplier", -beta2),
+        "installedCapacity" -> UtilityFunctionOperation("multiplier", -beta3)
+      )
+    )
+    val mnl = MultinomialLogit[String, String](null)
+    val chargingInquiry = ChargingInquiry(Some(mnl), plugData = None, vehicle, vot = 0.0)
```
JustinPihony: @robfitzgerald @johanneshiry This is mostly a copy/paste from ChoosesParking. I know we had discussed the vot, but I wasn't sure how to work it so that fastcharging was preferred per @colinsheppard

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -228,6 +232,33 @@ class RideHailAgent(
           .copy(currentVehicle = Vector(vehicle.id), remainingShifts = shifts.get)
       }
   }
+  when(OfflineForCharging) {
+    case ev @ Event(ParkingInquiryResponse(stall, _), data) =>
+      log.debug("state(RideHailingAgent.OfflineForCharging.ParkingInquiryResponse): {}", ev)
+      vehicle.useParkingStall(stall)
```
JustinPihony: I still have to run tests, but can anybody verify that this would be enough to move the vehicle to this location? Or do I need to first create a routing request and move the vehicle. IIRC the manager will handle that - but it doesn't hurt to verify with those who might know ;)

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
rjf-nrel: @JustinPihony we are actually modifying MNL, ParkingInquiry, and ChargingInquiry today as part of our integration of these things. it's different since an hour ago, so, maybe pull down the changes for now.

for prioritizing fast charging, perhaps we could add a term which is the "duration until full" cost, and, for RHA, set that multiplier higher, so it dominates over the other choices. i wonder what @johanneshiry thinks of that idea.

i'll be busy for some bits of the middle of today but i'll try and keep up with the discussion.

PR #1883, opened 2019-06-05
Let emergency parking be expensive, but in the desired place

michaz:
This changes the emergency parking stall that is returned when there is no parking available.

Previously, this would be a parking stall that is expensive and somewhere far away in the world, at a random coordinate. This would put stress on the router by having it go through its quadtree a lot and give error messages.

Now, it is only expensive, and otherwise like the parking stall in the "abundant" parking lot, i.e. directly at the desired destination, so no routing is done.

This also fixes a test which was supposed to test for "no parking available", but the data file was bad, so there _was_ parking available. Possibly this was the only reason why for this case, the "far out" parking location would actually be necessary sometimes.

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1883)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -13,11 +13,9 @@ import beam.agentsim.agents.vehicles.AccessErrorCodes.RideHailNotRequestedError
 import beam.agentsim.agents.vehicles.VehicleProtocol.StreetVehicle
 import beam.agentsim.agents.vehicles.{VehiclePersonId, _}
 import beam.agentsim.events.{ModeChoiceEvent, SpaceTime}
-import beam.agentsim.infrastructure.{ParkingInquiry, ParkingInquiryResponse}
-import beam.agentsim.infrastructure.ParkingStall
+import beam.agentsim.infrastructure.{ParkingInquiry, ParkingInquiryResponse, ParkingStall}
 import beam.agentsim.scheduler.BeamAgentScheduler.{CompletionNotice, ScheduleTrigger}
 import beam.router.BeamRouter._
-import beam.router.Modes
 import beam.router.Modes.BeamMode
 import beam.router.Modes.BeamMode._
 import beam.router.model.{BeamLeg, EmbodiedBeamLeg, EmbodiedBeamTrip}
@@ -1146,9 +1144,9 @@ object ChoosesMode {
       parkingResponse = if (withParking) {
         None
       } else {
-        Some(ParkingInquiryResponse(ParkingStall.lastResortStall(), 0))
+        Some(ParkingInquiryResponse(ParkingStall.defaultStall(new Coord(0.0, 0.0)), 0))
       },
-      driveTransitParkingResponse = Some(ParkingInquiryResponse(ParkingStall.lastResortStall(), 0)),
+      driveTransitParkingResponse = Some(ParkingInquiryResponse(ParkingStall.defaultStall(new Coord(0.0, 0.0)), 0)),
       rideHailResult = if (withRideHail) {
         None
       } else {
--- a/src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
+++ b/src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
@@ -3,9 +3,9 @@ package beam.agentsim.infrastructure
 import scala.annotation.tailrec
 import scala.collection.JavaConverters._
 import scala.util.{Failure, Random, Success, Try}
-
 import akka.actor.{Actor, ActorLogging, ActorRef, Props}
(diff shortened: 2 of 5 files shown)


michaz:
test!

michaz:
@robfitzgerald  Also: "Default" TAZ IDs don't work in the long run, let's remove them. It's the same as with vehicle type ids: Having a TAZ ID means I can go to the TAZ map and get the TAZ by ID. This fails with made-up IDs.

Doesn't this rather mean that the model for ParkingStall is wrong and a TAZ ID is really optional? Then you wouldn't have to make up TAZ IDs.

Conversely, it seems that ParkingStall.pricingModel is actually *not* optional, since when a parking stall with pricingModel=NONE is written to an event file, the event reader crashes upon reading it back in.

michaz:
This seems to be non-deterministic! On my machine, the test case that fails up there ^^ is positive so far.

michaz:
@robfitzgerald Also (found this on develop, not on this branch), in the "empty" parking test scenario, drivers are happily using the parking reserved for ride hail agents!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1146,9 +1144,9 @@ object ChoosesMode {
       parkingResponse = if (withParking) {
         None
       } else {
-        Some(ParkingInquiryResponse(ParkingStall.lastResortStall(), 0))
+        Some(ParkingInquiryResponse(ParkingStall.defaultStall(new Coord(0.0, 0.0)), 0))
```
robfitzgerald: why is (0,0) better than a random stall location? i think it would be the same amount of work to re-map to a link in the network, and (0,0) makes assumptions about the coordinates of the running scenario (like, what about when (0,0) is a real location on the user's network?).

i should probably also mention my reasoning behind lastResortStall. it was suggested to me that it is acceptable to depend on connecting a random coordinate back to the scenario's road network. that, and the fact that the map is not in scope here, made it seem to be the best solution.

src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
```diff
@@ -258,7 +262,14 @@ object ZonalParkingManager extends LazyLogging {
       case Some(result) =>
         result
       case None =>
-        val newStall = ParkingStall.lastResortStall(random)
+        val newStall =
+          ParkingStall
+            .defaultStall(destinationUTM)
```
robfitzgerald: based on what you said, this could be the agent's home. however, if we just send them to the destination, then they may not learn that this was a bad route, because the remaining "score" of their day might lead them to believe they actually found parking. 

when the lastResortStall is produced, it really means, "ya, you're getting parking, but, you really shouldn't be."

robfitzgerald:
> @robfitzgerald Also (found this on develop, not on this branch), in the "empty" parking test scenario, drivers are happily using the parking reserved for ride hail agents!

now that is weird. as of right now, RideHailAgents call RideHailManager, and PersonAgents call ZonalParkingManager. i didn't know there was any way for them to interact.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
michaz: At _this_ point in the code, it means something different yet again, unfortunately. This stall is actually _never_ used, it is just a dummy object that we need to get through the logic of this class in a certain way. Bad, bad, bad, I know.

src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala
michaz: I know, that's why it's _expensive_. My point was that that should be enough, rather than _additionally_ have it be somewhere else, don't you think?

michaz:
test!

michaz:
Test!

michaz:
> now that is weird. as of right now, RideHailAgents call RideHailManager, and PersonAgents call ZonalParkingManager. i didn't know there was any way for them to interact.

Resolved.

robfitzgerald:
> @robfitzgerald Also: "Default" TAZ IDs don't work in the long run, let's remove them. It's the same as with vehicle type ids: Having a TAZ ID means I can go to the TAZ map and get the TAZ by ID. This fails with made-up IDs.
> 
> Doesn't this rather mean that the model for ParkingStall is wrong and a TAZ ID is really optional? Then you wouldn't have to make up TAZ IDs.
> 
> Conversely, it seems that ParkingStall.pricingModel is actually _not_ optional, since when a parking stall with pricingModel=NONE is written to an event file, the event reader crashes upon reading it back in.

these are some great points. i love using optional when it is declarative like this. i only became aware of this bijection expected in events very recently, like last week. i would like to see these changes you mentioned come through the pipeline at some point.

PR #1885, opened 2019-06-06
dropped use of SortedSet in favor of a List in Multinomial Logit function

robfitzgerald:
dropped use of SortedSet in favor of a List in MultinomialLogit's sampleAlternative function, per discussions about the issues with equality testing in SortedSet (see #1884 and [this link](https://stackoverflow.com/questions/4952374/scala-sortedset-sorted-by-one-ordering-and-unique-by-something-else) for details).

tracking infinitely-valued alternatives is now done by placing infinitely-valued alternatives at the tail of the list, and afterward, checking the `lastOption` of the list for a utility which equals Double.PositiveInfinity.

this also changed the Tuple2 of (AlternativeType, Double) into it's own case class, which is clearer.

in support of this change, the following happened:

- added a test for dealing with positively infinite values
- added a test for dealing with equally-valued alternatives
- recognized that the MNL's output wasn't reporting the utility, but instead, the e^u value, which is less meaningful from a logging/event standpoint. the true utility value is now output

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1885)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/choice/logit/MultinomialLogit.scala
+++ b/src/main/scala/beam/agentsim/agents/choice/logit/MultinomialLogit.scala
@@ -47,18 +47,18 @@ class MultinomialLogit[A, T](
               case Some(thisUtility) =>
                 if (thisUtility == Double.PositiveInfinity) {
                   // place on tail of list, allowing us to short-circuit the sampling in next step
-                  accumulator :+ AlternativeWithUtility(alt, math.exp(thisUtility))
+                  accumulator :+ AlternativeWithUtility(alt, thisUtility, math.exp(thisUtility))
                 } else {
-                  AlternativeWithUtility(alt, math.exp(thisUtility)) +: accumulator
+                  AlternativeWithUtility(alt, thisUtility, math.exp(thisUtility)) +: accumulator
                 }
             }
         }
 
       altsWithUtility.lastOption.flatMap {
-        case AlternativeWithUtility(possiblyInfiniteAlt, possiblyInfinite) =>
-          if (possiblyInfinite == Double.PositiveInfinity) {
+        case AlternativeWithUtility(possiblyInfiniteAlt, possiblyInfiniteUtility, possiblyInfiniteExpUtility) =>
+          if (possiblyInfiniteExpUtility == Double.PositiveInfinity) {
             // take any infinitely-valued alternative
-            Some { MultinomialLogit.MNLSample(possiblyInfiniteAlt, possiblyInfinite, 1.0, 1.0) }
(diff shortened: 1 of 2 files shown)


robfitzgerald:
@michaz i made a change to MNL here, just planning to add a test before i try and merge.

src/main/scala/beam/agentsim/agents/choice/logit/MultinomialLogit.scala
```diff
@@ -36,35 +38,39 @@ class MultinomialLogit[A, T](
...
+        alternatives.foldLeft(List.empty[AlternativeWithUtility[A]]) {
+          case (accumulator, (alt, attributes)) =>
             getUtilityOfAlternative(alt, attributes) match {
-              case None => set
+              case None => accumulator
               case Some(thisUtility) =>
-                set + ((alt, math.exp(thisUtility)))
+                if (thisUtility == Double.PositiveInfinity) {
+                  // place on tail of list, allowing us to short-circuit the sampling in next step
+                  accumulator :+ AlternativeWithUtility(alt, math.exp(thisUtility))
```
REASY: append to List is linear, so for the big input it can hurt performance. If order does not matter, use prepend:
https://docs.scala-lang.org/overviews/collections/performance-characteristics.html

robfitzgerald: right. i need a way to differentiate between infinitely and non-infinitely-valued alternatives. i decided to prepend non-infinitely-valued alternatives, since that is more common (and more performant, as you mentioned), and to append the infinite ones. that way, i know, if there is at least one infinitely-valued alternative, i know where i can find it.

robfitzgerald:
> One comment on this:
> 
> * from my understanding of the code, altsWithUtility is not sorted yet, but to build a valid cdf to sample from it has to be sorted in ascending order somewhere before the utility probabilities are stacked together

not sure it needs to be sorted. whatever their order, they just eventually need to be mapped to values between 0 and 1, and that needs to be in order. that happens a little bit later with the asProbabilitySpread code block.

johanneshiry:
> > One comment on this:
> > 
> > * from my understanding of the code, altsWithUtility is not sorted yet, but to build a valid cdf to sample from it has to be sorted in ascending order somewhere before the utility probabilities are stacked together
> 
> @johanneshiry not sure it needs to be sorted. whatever their order, they just eventually need to be mapped to values between 0 and 1, and that needs to be in order. that happens a little bit later with the asProbabilitySpread code block.

looked it up again and it seems like you're right - it's not directly _necessary_ to sort an empirical distribution function to sample from (but one way to do it anyway) and we also didn't do that in the past. Hence I think, the code you pointed to should be sufficient.

rjf-nrel:
Test!

robfitzgerald:
Test!

PR #1890, opened 2019-06-11
Ridehail Human Charging

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1890)


src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -382,19 +455,36 @@ class RideHailAgent(
...
+    val beta1 = 1
+    val beta2 = 1
+    val beta3 = 0.001
+    val commonUtilityParams: Map[String, UtilityFunctionOperation] = Map(
+      "energyPriceFactor" -> UtilityFunctionOperation("multiplier", -beta1),
+      "distanceFactor"    -> UtilityFunctionOperation("multiplier", -beta2),
+      "installedCapacity" -> UtilityFunctionOperation("multiplier", -beta3)
+    )
+    val mnl = new MultinomialLogit[ParkingZoneSearch.ParkingAlternative, String](Map.empty, commonUtilityParams)
+    val inquiry = ParkingInquiry(destinationUtm, "work", 0.0, mnl, 0.0)
```
JustinPihony: From https://github.com/LBNL-UCB-STI/beam/pull/1882 :


JustinPihony 6 days ago  Author Collaborator
@robfitzgerald @johanneshiry This is mostly a copy/paste from ChoosesParking. I know we had discussed the vot, but I wasn't sure how to work it so that fastcharging was preferred per @colinsheppard

  @rjf-nrel
rjf-nrel 6 days ago  Contributor
@JustinPihony we are actually modifying MNL, ParkingInquiry, and ChargingInquiry today as part of our integration of these things. it's different since an hour ago, so, maybe pull down the changes for now.

for prioritizing fast charging, perhaps we could add a term which is the "duration until full" cost, and, for RHA, set that multiplier higher, so it dominates over the other choices. i wonder what @johanneshiry thinks of that idea.

i'll be busy for some bits of the middle of today but i'll try and keep up with the discussion.

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -228,6 +234,64 @@ class RideHailAgent(
...
+        0L
+      })
+      holdTickAndTriggerId(tick, triggerId)
+      goto(OfflineForCharging)
+  }
+
+  when(OfflineForCharging) {
+    case ev @ Event(ParkingInquiryResponse(stall, _), _) =>
+      log.debug("state(RideHailAgent.OfflineForCharging.ParkingInquiryResponse): {}", ev)
+      vehicle.useParkingStall(stall)
```
JustinPihony: I still have to run tests, but can anybody verify that this would be enough to move the vehicle to this location? Or do I need to first create a routing request and move the vehicle. IIRC the manager will handle that - but it doesn't hurt to verify with those who might know ;)

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -228,6 +234,64 @@ class RideHailAgent(
           .copy(currentVehicle = Vector(vehicle.id), remainingShifts = shifts.get)
       }
   }
+
+  when(PendingOfflineForCharging) {
```
JustinPihony: This interim step was added only because agents were getting stuck otherwise. It matches the flow of what happens without adding my logic. My biggest concern is how tick and trigger id are handled since the RHA is not scheduling this, but instead handling it itself. In a sf-light .5k run I encountered 40 

> Current tick is empty when moving vehicle to OfflineForCharging. This likely will get the agent stuck

However none got stuck like previously, so maybe it is fine. But if there is a better, more seamless/direct way to do this than using onTransition and this pending state then I'm all for it.

JustinPihony:
@colinsheppard I believe this is working now, however am not a pro at checking the runs for these. Also, I only ran against a .5 sf-light. How would I go about setting this to run against a larger run so that we can review that?

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -292,6 +292,7 @@ class RideHailAgent(
   when(Offline) {
     case Event(TriggerWithId(StartShiftTrigger(tick), triggerId), _) =>
       log.debug("state(RideHailingAgent.Offline): starting shift {}", id)
+      holdTickAndTriggerId(tick, triggerId)
```
JustinPihony: Any reason why to move this? Did it make a difference? If so I'd be curious why

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
johanneshiry: @rjf-nrel @robfitzgerald @JustinPihony 
if the fast charging preference is still under discussion I like the idea of just adding such a term. I'm not sure if a "duration until full" would fit exactly what @JustinPihony is looking for or if sth. like a installed capacity multiplier (positive) would fit better, but we could both give a chance and just test it

PR #1901, opened 2019-06-15
Added SimpleRideHailUtilization

REASY:
Simple ridehail utilization which creates the stats from number of passengers to amount of trips. It is implemented as one more `IterationSummaryAnalysis `, so the stats are automatically added to `summaryStats.csv` file. Stats column name are `RideTripsWith0Passengers`, `RideTripsWith1Passengers` and so on. PNGs files are automatically created with name `RideTripsWith0Passengers.png`, `RideTripsWith1Passengers.png` and so on. Example:
image

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1901)

### DIFF
--- a/src/main/java/beam/analysis/StatsFactory.java
+++ b/src/main/java/beam/analysis/StatsFactory.java
@@ -29,7 +29,8 @@ public class StatsFactory {
         AboveCapacityPtUsageDuration,
         TollRevenue,
         AgencyRevenue,
-        ParkingDelay
+        ParkingDelay,
+        RideHailUtilization
     }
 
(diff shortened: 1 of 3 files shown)


wrashid:
can we do single graph with stacked bar chart?

rajnikantsh:
Single graph for all iteration

rideHailUtilisation

wrashid:
number labels not needed in stack bar

rajnikantsh:
Updated graph after removing number labels
rideHailUtilisation

PR #1902, opened 2019-06-16
ScalaTest HTML reports

michaz:
Screen Shot 2019-06-16 at 16 04 09

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1902)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -193,6 +193,7 @@ dependencies {
 
     // TEST Scala //
     testCompile group: 'org.scalatest', name: "scalatest_${scalaBinaryVersion}", version: '3.0.1'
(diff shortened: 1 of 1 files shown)


REASY:
Would be great to have it attached to PR or somehow accessible via link

michaz:
This was suggested by @wrashid a long time ago and I was in opposition, because at that time our build wasn't even _clean_ yet, in particular the console output of the CI server was full of log messages which we didn't understand how to turn off because of logger trouble.

And I was like "let's not mask that problem by looking at a HTML file instead of the console".

Now I thought it would be neat to try it out. :-)

JustinPihony:
Should we integrate this into CI? I believe there are addons to display it on each PR.

PR #1907, opened 2019-06-17
Pin BeamAgentScheduler, ZonalParkingManager and RideHailManager to dedicated threads

REASY:
This gives two things:
- Slightly improves performance because of decreasing context switching
- More important one is ability to see what's going on with those actors in Java Mission Control:
image

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1907)


michaz:
Wooow, fancy! :-)

JustinPihony:
Approved - Just a note from a previous case where the Akka team has suggested this - for FYI:

> This will, however, make the actors relatively heavyweight because it will allocate a dedicated thread for each actor – having 10s of thousands of these might become challenging, though not necessarily impossible given proper tuning.

This is not the case here, but wanted to put it out there.

PR #1912, opened 2019-06-18
Put the logging for the link speed quick fix back in, but at the source

michaz:
The router was logging cases with extremely low reported link speeds before, and I removed that while working on the travel times per vehicle type, and forgot to put it back in. (Only the logging, the fix was never removed.)

Here, I put in pack in, but where the link travel times are computed, not where they are used.

It happens a lot, even in BeamVille. There it is because of busses, they take like a day to traverse a link.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1912)

### DIFF
--- a/src/main/scala/beam/router/r5/R5RoutingWorker.scala
+++ b/src/main/scala/beam/router/r5/R5RoutingWorker.scala
@@ -12,7 +12,6 @@ import beam.agentsim.agents.vehicles.FuelType.FuelType
 import beam.agentsim.agents.vehicles.VehicleProtocol.StreetVehicle
 import beam.agentsim.agents.vehicles._
 import beam.agentsim.events.SpaceTime
-import beam.agentsim.infrastructure.taz.TAZTreeMap
 import beam.router.BeamRouter._
 import beam.router.Modes.BeamMode.{CAR, WALK}
 import beam.router.Modes._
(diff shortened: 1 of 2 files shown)


michaz:
test!

src/main/scala/beam/router/r5/R5RoutingWorker.scala
```diff
@@ -975,7 +984,11 @@ class R5RoutingWorker(workerParams: WorkerParameters) extends Actor with ActorLo
           val link = networkHelper.getLinkUnsafe(linkId)
           assert(link != null)
           val physSimTravelTime = travelTime.getLinkTravelTime(link, time, null, null).ceil.toInt
-          Math.min(Math.max(physSimTravelTime, minTravelTime), maxTravelTime)
+          val linkTravelTime = Math.max(physSimTravelTime, minTravelTime)
+          if (linkTravelTime > maxTravelTime) {
+            println("wurst")
```
REASY: Seems like this is left from debugging?

REASY: Seems like this is left from debugging?

michaz: Indeed.

michaz: Fixed.

michaz:
test!

PR #1916, opened 2019-06-20
avoid NetworkCoordinator init twice

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1916)


michaz:
`network` and `transportNetwork` are available on `beamScenario`, after `val beamScenario = loadScenario(beamConfig)`. Could you instead move `val emptyScenario = ScenarioBuilder(matsimConfig).withNetwork(networkCoordinator.network).build` one line down, then you have the `network` you need?

(Then you wouldn't need to refactor the tests..)

PR #1928, opened 2019-06-25
Refer to transit segments by index, not by leg

michaz:
As a precondition to #1711, don't require passengers to present the exact same sequence of legs to the bus driver that they are going to ride. Instead, tell them the "slice" of the trip we want to take, by start and end index into the stop sequence.

As a side effect, the big global transit map isn't required anymore.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1928)


michaz:
test!

PR #1933, opened 2019-06-26
CSV reader pointing to individual files

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1933)


test/input/beamville/beam-csv.conf
```diff
@@ -5,14 +5,11 @@ include "beam.conf"
 
 beam.agentsim.simulationName = "beamville-csv"
 beam.agentsim.agentSampleSizeAsFractionOfPopulation = 1.0
+beam.agentsim.agents.population.inputPopulationFilePath = ${beam.inputDirectory}"/csvInput/population.csv"
```
wrashid: use existing attribute

src/main/scala/beam/sim/config/BeamConfig.scala
```diff
@@ -606,13 +606,17 @@ object BeamConfig {
...
           useVehicleSampling: scala.Boolean
         )
 
         object Population {
 
           def apply(c: com.typesafe.config.Config): BeamConfig.Beam.Agentsim.Agents.Population = {
             BeamConfig.Beam.Agentsim.Agents.Population(
+              inputPopulationFilePath =
+                if (c.hasPathOrNull("inputPopulationFilePath")) c.getString("inputPopulationFilePath")
+                else "/test/input/beamville/population.csv",
```
wrashid: default is empty

PR #1936, opened 2019-06-27
Read plans, persons and households information in parallel

### LINKED ISSUE #1929: UrbanSimScenarioLoader load time can be improved by reading different files in parallel
We can easily speed-up scenario loading by reading the data in parallel, for example, we can read plans and person files in parallel instead of [sequentially](https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala#L39-L56). Here is the logs from my machine when I run production-sfbay:
```
00:14:31.099 INFO  b.u.s.urbansim.CsvScenarioReader$ - readPlansFile executed in 53429 ms
00:15:21.821 INFO  b.u.scenario.UrbanSimScenarioLoader - Read 18885330 plans
00:16:14.817 INFO  b.u.s.urbansim.CsvScenarioReader$ - readPersonsFile executed in 52995 ms
00:16:15.072 INFO  b.u.scenario.UrbanSimScenarioLoader - Read 8153481 persons
00:16:25.961 INFO  b.u.scenario.UrbanSimScenarioLoader - There are 3777066 persons with plans
00:16:39.688 INFO  b.u.s.urbansim.CsvScenarioReader$ - readHouseholdsFile executed in 8463 ms
00:16:43.130 INFO  b.u.s.urbansim.CsvScenarioReader$ - readUnitsFile executed in 3442 ms
00:16:51.085 INFO  b.u.s.urbansim.CsvScenarioReader$ - readParcelAttrFile executed in 7955 ms
00:16:55.571 INFO  b.u.s.urbansim.CsvScenarioReader$ - readBuildingsFile executed in 4486 ms
00:17:18.561 INFO  b.u.s.u.UrbanSimScenarioSource - [1] getUnitIdToCoord executed in 22990 ms
```

REASY:
Fixes #1929 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1936)


src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -29,30 +31,40 @@ class UrbanSimScenarioLoader(
...
     logger.info(s"There are ${personsWithPlans.size} persons with plans")
 
     val householdIdToPersons: Map[HouseholdId, Iterable[PersonInfo]] = personsWithPlans.groupBy(_.householdId)
 
-    val householdsWithMembers = {
-      val households = scenarioSource.getHousehold
-      logger.info(s"Read ${households.size} households")
-      households.filter(household => householdIdToPersons.contains(household.householdId))
-    }
+    val households = Await.result(householdsF, 500.seconds)
```
carloscaldas: extract a variable `timeout`

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -29,30 +31,40 @@ class UrbanSimScenarioLoader(
...
+      logger.info(s"Read ${plans.size} plans")
+      plans
+    }
+    val personsF = Future {
       val persons: Iterable[PersonInfo] = scenarioSource.getPersons
       logger.info(s"Read ${persons.size} persons")
-      getPersonsWithPlan(persons, plans)
+      persons
+    }
+    val householdsF = Future {
```
carloscaldas: move this `Future` to `ScenarioSource`. Makes sense an adapter (`ScenarioSource`) returning a Future for expensive operation. In this case you have to change `BeamScenarioLoader` as well

REASY: What is the purpose of moving this to `ScenarioSource`? Why should I break `ScenarioSource` interface? This was simple thing to do in `UrbanSimScenarioLoader`, otherwise good solution will be make it async all the way up to `BeamScenarioReader`

REASY:
test!

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -29,30 +31,40 @@ class UrbanSimScenarioLoader(
...
+      val households = scenarioSource.getHousehold
+      logger.info(s"Read ${households.size} households")
+      households
     }
+    val plans = Await.result(plansF, 500.seconds)
+    val persons = Await.result(personsF, 500.seconds)
+
+    val personsWithPlans = getPersonsWithPlan(persons, plans)
```
JustinPihony: Can we keep the flow going instead of awaiting? Await is a smell, so if we can push the futures all the way until they are needed then that's probably better:

i.e.

```
val personsWithPlansF = for{
  plans <- plansF
  persons <- personsF
} yield getPersonsWithPlan(persons, plans)
```

And so on until the future is forced.

REASY: Agree than almost always blocking for the result is a bad idea, but in this particular case we can't do anything _better_ because the code is not async all the way. To be able to run the simulation we need to have properly populated scenario.  Also, `val population: Population = scenario.getPopulation` is Java class and it is blackbox for me, so I can't do side-effects on this class in the thread other than `main` because I'm not sure will be the changes properly visible in `main` thread (Future executes in another thread).

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
carloscaldas: The interface describes an Adapter to external source of information (in this case files xml/csv/parquet). It makes sense using futures to slow operations (like reading files, socket, database, etc.).
 At the same time we have redundancy: `BeamScenarioSource` and `UrbansimScenarioSource` that implements the same interface and could take advantage of Future approach.

PR #1938, opened 2019-06-28
refactored BeamScenarioLoader to build population without side effects

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1938)

### DIFF
--- a/src/main/scala/beam/sim/BeamHelper.scala
+++ b/src/main/scala/beam/sim/BeamHelper.scala
@@ -563,7 +563,7 @@ trait BeamHelper extends LazyLogging {
         val externalFolderExists: Boolean = Option(scenarioConfig.folder).exists(new File(_).isDirectory)
         if (externalFolderExists) {
           val beamScenario = loadScenario(beamConfig)
-          val emptyScenario = ScenarioBuilder(matsimConfig).withNetwork(beamScenario.network).build
+          val emptyScenario = ScenarioBuilder(matsimConfig, beamScenario.network).build
           val scenario = {
             val source = buildUrbansimScenarioSource(new GeoUtilsImpl(beamConfig), beamConfig)
             new UrbanSimScenarioLoader(emptyScenario, beamScenario, source, new GeoUtilsImpl(beamConfig)).loadScenario()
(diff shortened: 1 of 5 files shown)


carloscaldas:
test!

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -576,13 +576,13 @@ trait BeamHelper extends LazyLogging {
...
-            val emptyScenario = ScenarioBuilder(matsimConfig).withNetwork(beamScenario.network).build
             val scenario = {
               val source = new BeamScenarioSource(
                 beamConfig,
                 rdr = readers.BeamCsvScenarioReader
               )
-              new BeamScenarioLoader(emptyScenario, beamScenario, source, new GeoUtilsImpl(beamConfig)).loadScenario()
+              val scenarioBuilder = ScenarioBuilder(matsimConfig, beamScenario.network)
+              new BeamScenarioLoader(scenarioBuilder, beamScenario, source, new GeoUtilsImpl(beamConfig)).loadScenario()
             }.asInstanceOf[MutableScenario]
```
REASY: Won't it fail with `ClassCastException`  because of casting `BeamScenarioLoader` to `MutableScenario`?

src/main/scala/beam/utils/scenario/BeamScenarioLoader.scala
```diff
@@ -205,6 +213,16 @@ class BeamScenarioLoader(
     population
   }
 
+  private def selectedPlanFromPersonOrBuildNew(person: Person): Plan = {
+    var result = person.getSelectedPlan
```
REASY: Why using `var`?  You could use `Option(person.getSelectedPlan).getOrElse{}`

carloscaldas: replaced

src/main/scala/beam/sim/BeamHelper.scala
carloscaldas: No. there is a method call `.loadScenario()` at the end

Issue #1853, opened 2019-05-25
Organize startup into stages for easier debugging and code navigation


JustinPihony:
With BEAM moving towards generating files on demand, with providing pre-loaded data as optional configurations it would be easier to understand the flow if it were in clear stages:

- [ ] Stage 1.) Read all available data in the first stage. This will include both warmstart and root files (ie vehicles.csv/households.csv)
  - This may need split up between warmstart and general config, such that warmstart takes an entire directory and automatically looks in the corresponding locations, and root files act as a backup for anything not in warmstart. As such, warmstart would run first, then reading root files thereafter. @colinsheppard @wrashid Can you verify that this order makes sense
- [ ] Stage 2.) Any missing data should then be procedurally generated
- [ ] Stage 3.) Write the final state of the data to the output directory, such that it can be pointed back to during subsequent runs (as part of stage 1)
- [ ] Stage 4.) Execute the simulation

JustinPihony:
Also, if you'd like the bullets to be broken into separate issues (especially stage 1) then we can do that.

PR #1939, opened 2019-06-29
Refactoring of AvailableModeUtils: extracted method, simplified code …

carloscaldas:
Improvements:
- Converted class to object
- Extracted method to avoid code duplication
- Code simplification: avoided redundant tests
- Tightened methods visibility level
- Removed unused method

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1939)

### DIFF
--- a/src/main/scala/beam/utils/plan/PlansBuilder.scala
+++ b/src/main/scala/beam/utils/plan/PlansBuilder.scala
@@ -3,7 +3,7 @@ package beam.utils.plan
 import java.io.{BufferedWriter, File, FileWriter}
 import java.nio.file.{Files, Paths}
 
-import beam.utils.plan.sampling.AvailableModeUtils.{AllowAllModes, setAvailableModesForPerson}
+import beam.utils.plan.sampling.AvailableModeUtils.{setAvailableModesForPerson, AllowAllModes}
 import beam.utils.plan.sampling.HouseholdAttrib.{HomeCoordX, HomeCoordY, HousingType}
 import beam.utils.plan.sampling.PlansSampler.newPop
 import beam.utils.plan.sampling.PopulationAttrib.Rank
@@ -24,7 +24,7 @@ import org.matsim.utils.objectattributes.{ObjectAttributes, ObjectAttributesXmlW
 import org.matsim.vehicles.{Vehicle, VehicleUtils, VehicleWriterV1, Vehicles}
 
 import scala.collection.JavaConverters._
-import scala.collection.{JavaConverters, immutable}
+import scala.collection.{immutable, JavaConverters}
 import scala.util.Random
 
 object PlansBuilder {
--- a/src/main/scala/beam/utils/plan/PlansBuilder.scala
+++ b/src/main/scala/beam/utils/plan/PlansBuilder.scala
@@ -3,8 +3,7 @@ package beam.utils.plan
 import java.io.{BufferedWriter, File, FileWriter}
 import java.nio.file.{Files, Paths}
 
-import beam.sim.population.PopulationAdjustment
-import beam.sim.population.PopulationAdjustment.BEAM_ATTRIBUTES
+import beam.utils.plan.sampling.AvailableModeUtils.{AllowAllModes, setAvailableModesForPerson}
 import beam.utils.plan.sampling.HouseholdAttrib.{HomeCoordX, HomeCoordY, HousingType}
 import beam.utils.plan.sampling.PlansSampler.newPop
 import beam.utils.plan.sampling.PopulationAttrib.Rank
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/utils/plan/sampling/AvailableModeUtils.scala
```diff
@@ -50,81 +51,89 @@ object AvailableModeUtils extends LazyLogging {
...
+    Option(gerPersonCustomAttributes(person)).getOrElse {
+      val attribs: AttributesOfIndividual =
+        PopulationAdjustment.createAttributesOfIndividual(beamScenario, population, person, household)
       person.getCustomAttributes.put(PopulationAdjustment.BEAM_ATTRIBUTES, attribs)
       attribs
     }
-    setModesForPerson(person, population, permissibleModes, attributesOfIndividual)
+  }
+
+  private def gerPersonCustomAttributes(person: Person): AttributesOfIndividual = {
```
JustinPihony: Typo: ger -> get

src/main/scala/beam/utils/plan/PlansBuilder.scala
```diff
@@ -43,8 +43,7 @@ object PlansBuilder {
   val newHH: HouseholdsImpl = new HouseholdsImpl()
   val newHHAttributes: ObjectAttributes = newHH.getHouseholdAttributes
 
-  val modeAllocator: AvailableModeUtils.AllowAllModes =
-    new AvailableModeUtils.AllowAllModes
+  val modeAllocator: AvailableModeUtils.AllowAllModes.type = AvailableModeUtils.AllowAllModes
```
JustinPihony: If this is truly a static class (functions don't rely on internal state), which it appears is the case, then the standard is to reference it directly. This technically works, but should it be in a variable or just imported?

    import AvailableModeUtils.AllowAllModes

then it can just be referenced:

    AllowAllModes.METHOD

src/main/scala/beam/utils/plan/sampling/AvailableModeUtils.scala
```diff
@@ -50,81 +51,89 @@ object AvailableModeUtils extends LazyLogging {
...
     }
   }
 
+  private def getAvailableModesOfPerson(
+    person: Person,
+    population: Population,
+    permissibleModes: Seq[String]
+  ): Seq[String] = {
+    val excludedModes = getExcludedModesForPerson(population, person.getId.toString)
+    permissibleModes.filterNot(am => excludedModes.exists(em => em.equalsIgnoreCase(am)))
```
JustinPihony: You lost the existence check on `excludedModes`. I assume this was a micro-optimization, but it is more optimized. It should probably stay

src/main/scala/beam/utils/plan/PlansBuilder.scala
carloscaldas: Actually this is not a method. is an `Object` inside another `Object`. that is why the word `type` is needed

src/main/scala/beam/utils/plan/sampling/AvailableModeUtils.scala
carloscaldas: In this case outcome is exactly the same. Benefits: smaller code with at same level of readability(IMO is better readability) and micro optimization.

src/main/scala/beam/utils/plan/sampling/AvailableModeUtils.scala
carloscaldas: fixed

src/main/scala/beam/utils/plan/PlansBuilder.scala
carloscaldas: fixed

src/main/scala/beam/utils/plan/PlansBuilder.scala
```diff
@@ -123,7 +122,7 @@ object PlansBuilder {
     val permissibleModes = modeAllocator
       .getPermissibleModes(person.getSelectedPlan)
       .asScala
-    AvailableModeUtils.setAvailableModesForPerson(person, newPop, permissibleModes.toSeq)
+    setAvailableModesForPerson(person, newPop, permissibleModes.toSeq)
```
JustinPihony: Actually, this one I'm not a fan of. I prefer taking it to the final object and then leaving it there when it comes to statics. Otherwise it makes it just a bit harder to trace where the method is defined - especially for new devs

carloscaldas: This PR is a pre-step to fulfil the the issue https://github.com/LBNL-UCB-STI/beam/issues/1853

Having an "static" object to apply mutations to an object that should be immutable in the first place looks like spreading the problem around the system.

The final goal here is moving the responsibility yo the loader (`BeamScenarioLoader`). There is no reason to load scenario partially and in other steps fix the inconsistent scenario. 

To sum up: ScenarioLoader will be responsible to load persons (like is already is) and its attributes (like is partially is). Ideally the scenario will become immutable and once is loaded cannot change anymore

carloscaldas: in other words: mutating a state passed as parameter is a bad practice.
```
setAvailableModesForPerson(person, newPop, permissibleModes.toSeq)
```
will be replaced with
```
person.availableModes = calculateAvailableModes(newPop, permissibleModes.toSeq)
```
when building the scenario

carloscaldas: done

carloscaldas:
test!

PR #1941, opened 2019-07-01
all pattern matching ops on ChargingPointType now cover all cases

### LINKED ISSUE #1940: Cover new and old ChargingPointTypes in all match statements
10:45:49.912 [ClusterSystem-akka.actor.default-dispatcher-226] ERROR akka.actor.OneForOneStrategy - invalid argument: Level1
java.lang.IllegalArgumentException: invalid argument: Level1

This error was experienced due to the fact that the `ChargingPointType` class constructor and the functions on ChargingPointType do not cover all cases that exist.

This task is to fully cover all defined types in all cases where `ChargingPointType` is pattern-matched.

robfitzgerald:
errors were occurring because the legacy `ChargingPointType`s were not properly "cased out" in pattern matching over `ChargingPointType` values.

i also modified the ChargingPointType Regex so that it is "unanchored" and can grab occurrences of the named charging point from a row at any position. that is consistent with the design, but, now that i'm thinking about it, if a user used a reserved ChargingPointType or PricingModel name in their TAZ name, then the regex would find that. maybe we should restrict our parsing to the row/column instead of the entire row.

anyhow, this fixes the bug. perhaps additionally we should add test class(es) for the charging code. 

Closes #1940.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1941)

### DIFF
--- a/src/main/scala/beam/agentsim/infrastructure/charging/ChargingPointType.scala
+++ b/src/main/scala/beam/agentsim/infrastructure/charging/ChargingPointType.scala
@@ -56,30 +56,30 @@ object ChargingPointType {
 
   }
 
-  private[ChargingPointType] val CustomChargingPointRegex: Regex = "(\\w+)\\((\\d+),(\\w+)\\)".r
+  private[ChargingPointType] val CustomChargingPointRegex: Regex = "(\\w+)\\((\\d+),(\\w+)\\)".r.unanchored
 
   // matches either the standard ones or a custom one
   // these were breaking some tests with a ChargingPoint parsing error caused by Event handlers
   def apply(s: String): Option[ChargingPointType] = {
     s.trim.toLowerCase match {
-//      case "householdsocket"              => Some(HouseholdSocket)
-//      case "bluehouseholdsocket"          => Some(BlueHouseholdSocket)
-//      case "cee16asocket"                 => Some(Cee16ASocket)
-//      case "cee32asocket"                 => Some(Cee32ASocket)
-//      case "cee63asocket"                 => Some(Cee63ASocket)
-//      case "chargingstationtype1"         => Some(ChargingStationType1)
-//      case "chargingstationtype2"         => Some(ChargingStationType2)
-//      case "chargingstationccscombotype1" => Some(ChargingStationCcsComboType1)
-//      case "chargingstationccscombotype2" => Some(ChargingStationCcsComboType2)
-//      case "teslasupercharger"            => Some(TeslaSuperCharger)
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/agentsim/infrastructure/charging/ChargingPointType.scala
```diff
@@ -99,6 +99,12 @@ object ChargingPointType {
       case ChargingStationCcsComboType2 => 50
       case TeslaSuperCharger            => 135
       case CustomChargingPoint(_, v, _) => v
+      // legacy charging points (values taken from 2018 BEAM code)
```
robfitzgerald: @johanneshiry plz double-check the values i put in place for the legacy types, if you don't mind - both AC/DC and the KwH. i just grabbed them from the old implementation. i think it should work as expected.

REASY:
@robfitzgerald there is [scala formatiing failure:](https://beam-ci.tk/job/beam-forced-prs/1405/console)
```
23:10:19 :checkScalafmt FAILED
```
Could you please apply formatting? Thanks!

rjf-nrel:
test!

robfitzgerald:
@REASY @johanneshiry if you would be so kind to review, that would be rad :pray:

src/main/scala/beam/agentsim/infrastructure/charging/ChargingPointType.scala
johanneshiry: The "worse" thing about the old definitions is, that Level1 / Level2 Charging rate are not fixed defined but more of a range of installed charging capacity. Anyway the values you picked are in range or close to the definitions of Level1/2 charging. So we can keep them for now until the personal ev branch will be merged in.

src/main/scala/beam/agentsim/infrastructure/charging/ChargingPointType.scala
```diff
@@ -116,7 +122,13 @@ object ChargingPointType {
       case ChargingStationCcsComboType2 => DC
       case TeslaSuperCharger            => DC
       case CustomChargingPoint(_, _, c) => c
-      case _                            => throw new IllegalArgumentException("invalid argument")
+      // legacy charging points
+      case Level2    => AC
+      case Level1    => AC
+      case DCFast    => DC
+      case UltraFast => DC
+      case NoCharger => AC
```
johanneshiry: doesn't make sense as a call on NoCharger should throw an IllegalArgumentException but we can keep this anyway for now to make it work.

PR #1942, opened 2019-07-01
Fix warm start

### LINKED ISSUE #1893: sioux faux despite warmstart takes 5 iterations to reduce congestion - see if this is the case with beam in general now and why

rajnikantsh:
Closes #1893 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1942)

### DIFF
--- a/src/main/scala/beam/sim/BeamHelper.scala
+++ b/src/main/scala/beam/sim/BeamHelper.scala
@@ -556,8 +556,6 @@ trait BeamHelper extends LazyLogging {
 
     val fileFormat = scenarioConfig.fileFormat
 
-    warmStart(beamConfig, matsimConfig)
-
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -795,6 +796,8 @@ trait BeamHelper extends LazyLogging {
 
     new UrbanSimScenarioSource(
       scenarioFolder = beamConfig.beam.exchange.scenario.folder,
+      isWarmMode = beamConfig.beam.warmStart.enabled,
+      planFile = matsimConfig.plans.getInputFile,
```
carloscaldas: in this case is loading the wrong file. `matsimConfig.plans` points to xml and urbansim is supposed to load to csv/parquet files

JustinPihony:
@rajnikantsh @REASY Can this branch be deleted?

PR #1951, opened 2019-07-04
avoid warmstart loading multiple times and added as part of config

### LINKED ISSUE #1893: sioux faux despite warmstart takes 5 iterations to reduce congestion - see if this is the case with beam in general now and why

carloscaldas:
Closes #1893 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1951)


carloscaldas:
test!

src/main/scala/beam/sim/BeamWarmStart.scala
```diff
@@ -2,98 +2,78 @@ package beam.sim
...
+      case None =>
+        throwErrorFileNotFound(description, srcPath)
     }
   }
 
-  private def loadPopulation(runPath: String, populationFile: String): String = {
-    val plansPath = Paths.get(runPath, "warmstart_plans.xml")
-    unGunzipFile(populationFile, plansPath.toString, false)
-    plansPath.toString
+  private def throwErrorFileNotFound(fileDesc: String, path: String): String = {
```
REASY: There is (FileNotFoundException)[https://docs.oracle.com/javase/7/docs/api/java/io/FileNotFoundException.html] in Java. Why don't use it? At least the name of the method contains file not found

src/main/scala/beam/utils/FileUtils.scala
```diff
@@ -96,6 +96,7 @@ object FileUtils extends LazyLogging {
   def downloadFile(source: String, target: String): Unit = {
     assert(source != null)
     assert(target != null)
+    println(s"@@@@@ downloading $source to $target")
```
REASY: Please, remove `println` and use logging instead

src/main/scala/beam/sim/BeamWarmStart.scala
```diff
@@ -2,98 +2,78 @@ package beam.sim
...
-          } else {
-            logger.warn("Travel times failed to warm start, stats not found at path ( {} )", statsPath)
-            None
-          }
-        case None =>
-          logger.warn("Travel times failed to warm start, stats not found at path ( {} )", srcPath)
+  def readTravelTime: Option[TravelTime] = {
+    getWarmStartFilePath("linkstats.csv.gz", rootFirst = false) match {
+      case Some(statsPath) =>
+        if (Files.isRegularFile(Paths.get(statsPath))) {
```
REASY: Why this has been changed to `Files.isRegularFile`? What is the reason behind it?

src/main/scala/beam/utils/FileUtils.scala
carloscaldas: wip. missing the label

src/main/scala/beam/sim/BeamWarmStart.scala
carloscaldas: `exists` method should be avoided. if there is a directory with that name returns true and you get an exception further. what the method actually needs is a regular file instead of something with that name

src/main/scala/beam/sim/BeamWarmStart.scala
carloscaldas: done

src/main/scala/beam/utils/FileUtils.scala
carloscaldas: done

PR #1956, opened 2019-07-06
Added an ability to override links params

### LINKED ISSUE #1955: Add an ability to override links params
Beam should be able to override link's capacity, length and free flow speed from CSV file if it is provided. CSV format is the following:
```
link_id,capacity,free_speed,length
```

REASY:
- Added new param `beam.physsim.overwriteLinkParamPath` which can poi…nt to the CSV file with params for the links
- `NetworkCoordinator` will overwrite link params in case it finds params in CSV file `beam.physsim.overwriteLinkParamPath`

Fixes #1955 

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1956)


src/main/scala/beam/router/r5/NetworkCoordinator.scala
```diff
@@ -111,4 +151,27 @@ trait NetworkCoordinator extends LazyLogging {
...
+    val path = beamConfig.beam.physsim.overwriteLinkParamPath
+    val filePath = new File(path).toPath
+    if (path.nonEmpty && Files.exists(filePath) && Files.isRegularFile(filePath)) {
+      try {
+        BeamVehicleUtils.readCsvFileByLine(path, scala.collection.mutable.HashMap[Int, LinkParam]()) {
+          case (line: java.util.Map[String, String], z) =>
+            val linkId = line.get("link_id").toInt
+            val capacity = Option(line.get("capacity")).map(_.toDouble)
+            val freeSpeed = Option(line.get("free_speed")).map(_.toDouble)
+            val length = Option(line.get("length")).map(_.toDouble)
```
wrashid: number of lanes not set here - does it default to 1?

PR #1966, opened 2019-07-10
zn/assign vehicles by income

zneedell:
Gives the option to assign vehicles by to different income brackets with different probabilities. This is done by adding a new column to vehicleTypes.csv, `sampleProbabilityString`, that gives the probability that a certain vehicle type is selected for different groups of users. In the string, different probability groups are separated by `;`, the name and different keys within a group are separated by a pipe "|", and within a key the key's range and its associated the probability is separated by `:`. Spaces are ignored.

For example, `ridehail | all:0.6; income | 0-17:0.5 | 17-36:0.25 | 61-86:1.0 | 86-133 : 0.0 | 133-999 : 0.0` means that that vehicle type makes up 60% of the ridehail fleet,. For personal vehicles, probabilities are different for households in different household ranges. For households with income between 0 and 17000, the probability is 50%, for incomes between 17000 and 36000 the probability is 25%, and so on. RIght now only income is implemented, but it will be easy to have probabilities also associated with different household characteristics. The bins aren't hard coded in anymore, instead they are inferred from the set of probability strings. All of the ranges are pooled over all of the vehicles, so when pooled the ranges have to be consistent and can't overlap (e.g. if you have a range of 0-17 and 17-36 for one vehicle, you can't have a range of 0-36 for another vehicle). If the bins are defined such that a household belongs to multiple bins, a warning is thrown and a bin is chosen from the matched ones. If there is an income that isn't covered by the ranges, a warning is thrown as well.

This behavior can be turned on by setting the parameter `beam.agentsim.agents.vehicles.vehicleAdjustmentMethod = "INCOME_BASED"` in the config file, but it is turned off as a default and uniform vehicle adjustment is used instead (if `beam.agentsim.agents.population.useVehicleSampling = True`).

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1966)


zneedell:
Test!

zneedell:
test!

zneedell:
test!

src/main/scala/beam/sim/vehicles/IncomeBasedVehiclesAdjustment.scala
```diff
@@ -0,0 +1,110 @@
...
+        vehicleTypesAndProbabilityByCategoryAndGroup.head._2
+    }
+    (1 to numVehicles).map { _ =>
+      val newRand = realDistribution.sample()
+      val (vehType, _) = vehTypeWithProbability.find { case (_, prob) => prob >= newRand }.get
+      vehType
+    }.toList
+  }
+
+  private def getHouseholdIncomeGroup(
```
colinsheppard: Why couldn't we use Ranges instead of hard coding? What if someone who is not American wants to use BEAM, do they have to convert their incomes into USD? Why the random dividing points on these ranges? Why not 0 - 20k, 20k - 40k, etc?

I don't want to slow things down seeking perfection, but I'm also worried these issues will require a refactor in the future at a time that is also inconvenient

src/main/scala/beam/sim/vehicles/VehiclesAdjustment.scala
```diff
@@ -28,12 +28,17 @@ trait VehiclesAdjustment extends LazyLogging {
 }
 
 object VehiclesAdjustment {
-  val DEFAULT_ADJUSTMENT = "DEFAULT_ADJUSTMENT"
-  val UNIFORM_ADJUSTMENT = "UNIFORM_ADJUSTMENT"
+  val DEFAULT_ADJUSTMENT = "DEFAULT"
```
colinsheppard: what is difference between DEFAULT and UNIFORM?

src/main/scala/beam/sim/vehicles/IncomeBasedVehiclesAdjustment.scala
zneedell: That makes sense. The current oddly-specific ranges are the ones used by NREL, but it wouldn't be too hard to have the ranges themselves defined by the group ID provided in the input file.

src/main/scala/beam/sim/vehicles/VehiclesAdjustment.scala
zneedell: Not totally sure--both were there before so I just kept them. I'll merge them to simplify.

zneedell:
Changed the string format to be a bit more flexible and allow for the income ranges to be defined by the input file. It works and is flexible but not the most straightforward, so I'd welcome suggestions.

zneedell:
test!

test/input/sf-light/sf-light-1k.conf
```diff
@@ -1,4 +1,3 @@
-include "../common/akka.conf"
```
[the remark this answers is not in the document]
zneedell: Oops--not sure how that got in there

src/main/scala/beam/sim/vehicles/IncomeBasedVehiclesAdjustment.scala
```diff
@@ -0,0 +1,126 @@
...
+    vehTypes: Array[BeamVehicleType]
+  ): scala.collection.mutable.Map[(VehicleCategory, (String, String)), Array[(BeamVehicleType, Double)]] = {
+    var groupIDs = scala.collection.mutable.Map[(VehicleCategory, (String, String)), Array[(BeamVehicleType, Double)]]()
+    var groupIDlist: scala.collection.mutable.ListBuffer[(VehicleCategory, (String, String))] =
+      scala.collection.mutable.ListBuffer()
+    var vehicleTypeAndProbabilityList: scala.collection.mutable.ListBuffer[(BeamVehicleType, Double)] =
+      scala.collection.mutable.ListBuffer()
+    vehTypes.foreach { vehType =>
+      vehType.sampleProbabilityString.getOrElse("All").replaceAll("\\s", "").toLowerCase.split(";").foreach { group =>
+        val keyAndValues = group.split('|')
```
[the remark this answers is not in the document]
zneedell: This is cool, thanks!

zneedell:
test!

zneedell:
test!

zneedell:
test!

zneedell:
test!

zneedell:
test!

PR #1979, opened 2019-07-15
Add default write interval for routeHistory to 10

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1979)


src/main/resources/beam-template.conf
```diff
@@ -436,6 +436,7 @@ beam.outputs.writeLinkTraversalInterval = "int | 0"
 beam.outputs.writeSkimsInterval = "int | 0"
 beam.physsim.writeEventsInterval = "int | 0"
 beam.physsim.writePlansInterval = "int | 0"
+beam.physsim.writeRouteHistoryInterval = "int | 10"
```
JustinPihony: Should the default be 10, or 1 for backwards compatibility. Where the bigger thing is that it is now configurable?

src/main/scala/beam/router/RouteHistory.scala
```diff
@@ -107,19 +107,26 @@ class RouteHistory @Inject()(
...
+        toCsv(routeHistory),
+      )
+    }
+
     previousRouteHistory = routeHistory
     routeHistory = new TrieMap()
   }
 
+  private def shouldWriteInIteration(iterationNumber: Int, interval: Int): Boolean = {
+    interval == 1 || (interval > 0 && iterationNumber % interval == 0)
```
JustinPihony: This will not write at all if it is a negative or 0. In an effort to handle misconfiguration maybe it should write every iteration in those cases?

src/main/resources/beam-template.conf
rajnikantsh: Hi @JustinPihony this param control routeHistory file writing which is used for warmStart. For warmstart we need around 20 to 30 previous iteration. So we found 10 as default value is appropriate for this param because Sometime user forgot to set this param so its good we write this file in 10,20,30.. iterations

src/main/scala/beam/router/RouteHistory.scala
rajnikantsh: for negative we will do fail fast for this config param and for 0 we will write file in all iterations

JustinPihony:
Neither of those are reflected in this PR though

src/main/scala/beam/router/RouteHistory.scala
JustinPihony: OK, I had been on mobile and see you added fail fast. And misread your 0 scenario - so approving

rajnikantsh:
test!

rajnikantsh:
test!

PR #1989, opened 2019-07-18
remove default energy primary and secondary file path from vehicleTyp…

### LINKED ISSUE #1988: Remove default-energy-primary.csv from all vehicleTypes.csv files
Across all input data, including in all production branches, please find all instances of a vehicleTypes.csv file (and the variations on that same file) and remove any reference to 

"default-energy-primary.csv"

or 

"default-energy-secondary.csv"

in these files. We don't need to specify a file path for the columns "primaryVehicleEnergyFile" and "secondaryVehicleEnergyFile" so these can be left blank instead of referring to some default energy file.

If there is a value that refers to an actual file (e.g. "/AllAboutMe_longTerm_highTech/conv-car-gas.csv.gz") then definitely leave that

rajnikantsh:
remove default energy primary and secondary file path from vehicleType.csv

Fixes #1988 

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1989)


JustinPihony:
Before merging it might be useful to review [my comment in the issue](https://github.com/LBNL-UCB-STI/beam/issues/1988#issuecomment-513084591).

PR #1993, opened 2019-07-22
jhrf/#1127 ev charging filter results

zneedell:
Filter the parking stalls returned from parking search based on context to allow or not allow vehicles to park in stalls with chargers. Intended to keep chargers from filling up with non-EVs and to ensure that EVs find any available chargers when they need to charge.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1993)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
@@ -534,7 +534,7 @@ class RideHailAgent(
       "installedCapacity" -> UtilityFunctionOperation("multiplier", -beta3)
     )
     val mnl = new MultinomialLogit[ParkingZoneSearch.ParkingAlternative, String](Map.empty, commonUtilityParams)
-    val inquiry = ParkingInquiry(destinationUtm, "work", 0.0, mnl, 0.0)
+    val inquiry = ParkingInquiry(destinationUtm, "charge", 0.0, mnl, 0.0, Option(vehicle))
     parkingManager ! inquiry
   }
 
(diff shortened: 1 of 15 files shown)


src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -534,7 +534,7 @@ class RideHailAgent(
       "installedCapacity" -> UtilityFunctionOperation("multiplier", -beta3)
     )
     val mnl = new MultinomialLogit[ParkingZoneSearch.ParkingAlternative, String](Map.empty, commonUtilityParams)
-    val inquiry = ParkingInquiry(destinationUtm, "work", 0.0, mnl, 0.0)
+    val inquiry = ParkingInquiry(destinationUtm, "charge", 0.0, mnl, 0.0, Option(vehicle))
```
johanneshiry: I have no idea if this is correct or not - maybe @colinsheppard knows if this is fine?

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -894,9 +894,9 @@ class RideHailManager(
     Map(request.customer.personId -> fare)
   }
 
-  def findRefuelStationAndSendVehicle(rideHailAgentLocation: RideHailAgentLocation): Unit = {
+  def findRefuelStationAndSendVehicle(rideHailAgentLocation: RideHailAgentLocation, beamVehicle: BeamVehicle): Unit = {
     val destinationUtm = rideHailAgentLocation.currentLocationUTM.loc
-    val inquiry = ParkingInquiry(destinationUtm, "work")
+    val inquiry = ParkingInquiry(destinationUtm, "charge", Option(beamVehicle))
```
johanneshiry: Same here ...

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
colinsheppard: sorry, can't tell by glancing at it

PR #1999, opened 2019-07-23
zn/#1998 fix energy calcs

### LINKED ISSUE #1998: RouteE energy calculations don't work
I ran production-sfbay baseline with the new RouteE energy files from NREL, and there were two problems:

1: BEAM couldn't load the rate files because they were missing a `num_lanes_int_bins` column.

2: When that is fixed, the energy calculations revert to the default energy in J per meter even though the correct files exist. This appears to be due to `onlyLength_Id_And_Type` in BeamVehicle.scala  https://github.com/LBNL-UCB-STI/beam/blob/c9d42e1f1227ea3a7d62c32137baa491cd5ae996/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L294. This input is set to `True` when a vehicle energy file exists for the current beamVehicleType, which leads to a `FuelConsumptionData` being created with None for the link speed. When `onlyLength_Id_And_Type == false`, the actual link speed is used. This seems backwards, and negating `onlyLength_Id_And_Type` appears to lead to the correct behavior.

zneedell:
Two small fixes:

1: when the `num_lanes_int_bins` column doesn't exist, it assumes that all rows are for a bin of 0-100 lanes.

2: negates what was previously called `onlyLength_Id_And_Type` in BeamVehicle.scala so that the link speed is set as none if the vehicle mapping file does not exist.

Fixes #1998

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/1999)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
+++ b/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
@@ -120,12 +120,13 @@ class BeamVehicle(
     *
     */
   def useFuel(beamLeg: BeamLeg, beamScenario: BeamScenario, networkHelper: NetworkHelper): FuelConsumed = {
+    val fuelConsumptionDataWithOnlyLength_Id_And_Type = !beamScenario.vehicleEnergy.vehicleEnergyMappingExistsFor(beamVehicleType)
     val fuelConsumptionData =
       BeamVehicle.collectFuelConsumptionData(
         beamLeg,
         beamVehicleType,
         networkHelper,
-        beamScenario.vehicleEnergy.vehicleEnergyMappingExistsFor(beamVehicleType)
+        fuelConsumptionDataWithOnlyLength_Id_And_Type
       )
 
     val primaryEnergyForFullLeg =
@@ -291,12 +292,12 @@ object BeamVehicle {
     beamLeg: BeamLeg,
     theVehicleType: BeamVehicleType,
     networkHelper: NetworkHelper,
-    vehicleMappingExists: Boolean = false
+    fuelConsumptionDataWithOnlyLength_Id_And_Type: Boolean = false
   ): IndexedSeq[FuelConsumptionData] = {
     //TODO: This method is becoming a little clunky. If it has to grow again then maybe refactor/break it out
     if (beamLeg.mode.isTransit & !Modes.isOnStreetTransit(beamLeg.mode)) {
       Vector.empty
-    } else if (!vehicleMappingExists) {
+    } else if (fuelConsumptionDataWithOnlyLength_Id_And_Type) {
       beamLeg.travelPath.linkIds
         .drop(1)
         .map(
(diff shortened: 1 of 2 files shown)


src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
```diff
@@ -116,7 +116,11 @@ class ConsumptionRateFilterStoreImpl(
       .foreach(csvRecord => {
         val speedInMilesPerHourBin = convertRecordStringToRange(csvRecord.getString(speedBinHeader), isDouble = true)
         val gradePercentBin = convertRecordStringToRange(csvRecord.getString(gradeBinHeader), isDouble = true)
-        val numberOfLanesBin = convertRecordStringToRange(csvRecord.getString(lanesBinHeader))
+        val numberOfLanesBin = if (csvRecord.getMetaData.containsColumn(lanesBinHeader)) {
+          convertRecordStringToRange(csvRecord.getString(lanesBinHeader))
+        } else {
+          convertRecordStringToRange("(0,100]")
```
JustinPihony: As long as this is valid then I'm fine with it. The biggest question to me is if a lanes header is a required field. I'll approve, but pinging @wrashid and @colinsheppard so they can chime in if needed

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -291,12 +291,12 @@ object BeamVehicle {
     beamLeg: BeamLeg,
     theVehicleType: BeamVehicleType,
     networkHelper: NetworkHelper,
-    onlyLength_Id_And_Type: Boolean = false
+    vehicleMappingExists: Boolean = false
```
JustinPihony: This is not an accurate refactoring of what is meant here. This method was in the process of being pared down to reduce calculating things it didn't need any longer. So, this if else is supposed to indicate that we care to get back `FuelConsumptionData` with ONLY length, id and type filled in. And the only time this happens currently is if the mapping exists. A better rename would be to be even more explicit - `fuelConsumpionDataWithOnlyLength_Id_And_Type`.

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -291,12 +291,12 @@ object BeamVehicle {
...
   ): IndexedSeq[FuelConsumptionData] = {
     //TODO: This method is becoming a little clunky. If it has to grow again then maybe refactor/break it out
     if (beamLeg.mode.isTransit & !Modes.isOnStreetTransit(beamLeg.mode)) {
       Vector.empty
-    } else if (onlyLength_Id_And_Type) {
+    } else if (!vehicleMappingExists) {
```
JustinPihony: The reason negating this results in the correct output is because it goes back to pulling in more data than is needed. A more accurate fix would be to include the speed. I could've swore when we ran this originally it pulled this in...but I can't recall what happened and why this was turned into a None at some point. Either way, keeping the boolean check as is is correct - but providing the necessary speed is the better path.

zneedell: OK sounds good, thanks--so to clarify:

As originally written, collectFuelConsumptionData can return fuelConsumptionData in two ways: One with only link length, link ID, and vehicle type, and one with those plus things like link speed, free flow speed, and number of lanes. The first way is called when a vehicle energy rate mapping file exists, and the second is called when it doesn't. One way we could get RouteE to correctly calculate energy use is by passing speed as well when the first way is used.

I'm still a little confused because it seems to me that the first way (return just length, id, and vehicle type) is all that we would need to estimate energy using the default energy consumption in J/m, whereas the second way already includes the information (such as speed and # lanes) that also can be used in RouteE but isn't used for the default energy calculations. Free flow speed is also passed in the second way, and as far as I can see that is not used in RouteE, so that does seem like something we could get rid of to avoid passing unnecessary information.

zneedell:
@JustinPihony in the interest of time I went ahead and pushed an update going along with negating the boolean and passing the simplified collectFuelConsumptionData when the mapping file doesn't exist. Definitely happy to circle back and do it the other way, though, if that sounds better to you.

In getting rid of unnecessary data from collectFuelConsumptionData when the mapping does exist, I noticed that getRateUsing in VehicleEnergy.scala was using the free flow speed rather than the actual link speed to look up the rate. I do remember fuel economy seeming kinda high in older runs, so this could make sense.

zneedell:
test!

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
JustinPihony: I believe the 2nd way is in fact going to go away sometime soon afaik, but we didn't do it yet.

Also, you are correct that lanes need to be included. I am not sure what happened that this was lost at some point...I swear it was there...or else it wouldn't have worked at first....

JustinPihony:
I think your new changes make sense...and we can even get rid of the branching altogether if we've commented out the rest of the unnecessary stuff (and it hasn't been used since). Some of the changes (default lanes and now using the different speed) should be reviewed by @colinsheppard or @wrashid imho, though.

zneedell:
test!

PR #2013, opened 2019-07-29
Cs/pool en route

colinsheppard:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2013)


src/main/scala/beam/agentsim/agents/MobilityRequest.scala
```diff
@@ -35,6 +40,39 @@ case class MobilityRequest(
       case Some(p) => p.personId.toString
       case None    => "None"
     }
-    s"${formatTime(time)}|$tag|${personid}|${activity.getType}| => ${formatTime(serviceTime)}"
+//    s"${formatTime(baselineNonPooledTime)}|$tag|${personid}|${activity.getType}| => ${formatTime(serviceTime)}"
+    s"${baselineNonPooledTime}|$tag|${personid}|${activity.getCoord}| => ${serviceTime}"
+  }
+  override def equals(that: Any): Boolean = {
```
REASY: This is tricky and can cause different issues in the future when you have a collection (List/Vector/Map) of `MobilityRequest` and want to find an instance of `MobilityRequest`.

src/main/resources/beam-template.conf
```diff
@@ -144,7 +144,7 @@ beam.agentsim.agents.rideHail.allocationManager.repositionLowWaitingTimes.waitin
 beam.agentsim.agents.rideHail.allocationManager.repositionLowWaitingTimes.demandWeight = "double | 4.0"
 beam.agentsim.agents.rideHail.allocationManager.repositionLowWaitingTimes.produceDebugImages = true
 beam.agentsim.agents.rideHail.allocationManager.alonsoMora.pickupTimeWindowInSec = "int | 360"
-beam.agentsim.agents.rideHail.allocationManager.alonsoMora.dropoffTimeWindowInSec = "int | 600"
+beam.agentsim.agents.rideHail.allocationManager.alonsoMora.dropoffTimeWindowInSec= "double | 0.2"
```
wrashid: this seems too low 0.2 seconds...

colinsheppard: It has been changed into a multiplier in a linear model. But it should have a new name.

src/main/scala/beam/agentsim/agents/MobilityRequest.scala
colinsheppard: Agree, but there's an important reason for doing it.

src/main/resources/beam-template.conf
wrashid: ok, got it

PR #2014, opened 2019-07-30
PersonAgent realizes when CAV was missed

### LINKED ISSUE #2012: New Household CAV can schedule triggers in the past
I think there’s a bug in the latest changes you made to Household CAV coordination (i.e. when you made them act like transit).

If you run sf-light-1k on branch develop with these params:

```
beam.agentsim.agents.vehicles.vehicleTypesFilePath = ${beam.inputDirectory}"/vehicleTypes-cavs-all.csv"
beam.agentsim.agents.vehicles.vehiclesFilePath = ${beam.inputDirectory}"/sample/1k/vehicles-cav.csv"
```

We end up with error, e.g. 

```
14:36:51.862 ERROR beam.agentsim.agents.PersonAgent - Cannot schedule an event ScheduleTrigger(BoardVehicleTrigger(37320,603-0),Actor[akka://ClusterSystem/user/BeamMobsim.iteration/population/030101-2016001422251-0/030101-2016001422251-0-7062567#1456612010],0) at tick 37320 when 'nowInSeconds' is at 37565}
14
```

michaz:
Fixes #2012 

By letting the passenger agent do the same as with transit: Realize that it missed the departure and go to replanning rather than sending a reservation request for a trip that it missed.

(We can't replan CAV vehicles, or let them wait for delayed people, but we couldn't do that before either.)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2014)


michaz:
One test fails, but has been failing before.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -749,6 +749,26 @@ class PersonAgent(
...
       )
       goto(WaitingForReservationConfirmation)
+    // CAV but too late
+    // TODO: Refactor so it uses literally the same code block as transit
+    case Event(StateTimeout, data @ BasePersonData(_, _, nextLeg :: _, _, _, _, _, _, _, _, _))
+        if nextLeg.beamLeg.startTime < _currentTick.get =>
+      // We've missed the CAV. This occurs when something takes longer than planned (based on the
+      // initial inquiry). So we replan but change tour mode to WALK_TRANSIT since we've already done our non-transit
+      // portion.
+      log.error("Missed CAV pickup, late by {} sec", _currentTick.get - nextLeg.beamLeg.startTime)
```
JustinPihony: Curious, should this be a warning, or even less? Does it really warrant error? Error to me means there is a bug in something (could be config vs code) and this is here to notify us - my two cents at least

colinsheppard: I agree @JustinPihony, can you make this a WARN on the RH depot charging branch?

michaz: Yes, agree, too. I _think_ I made this an error because before this fix it used to be, .. well, an actual exception, and I thought maybe we still want to know how often this happens for a short while.

PR #2016, opened 2019-07-30
Add File-Based Surge Pricing

sfwatergit:
Refactor surge pricing by creasing interface, extracting the original adaptive surge pricing mechanism, and adding read-from file-based surge pricing. Also added default "Dummy" implementation of interface (surge pricing factor stays at 1.0).


JustinPihony:
This is a bit old - should it be cleaned up and merged still?

@sfwatergit

PR #2022, opened 2019-08-01
BeamScoringFunctionFactory: Make sure we access only exist elements

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2022)


REASY:
test!

PR #2028, opened 2019-08-02
provide ZonalParkingManager's maxSearchRadius as a config parameter

robfitzgerald:
the default value for maxSearchRadius was 1000 meters, and hard-coded. TAZs may be more spread out, and so the search for parking would never find these alternatives. since it may vary by scenario, this should be provided as a config parameter.

```
beam.agentsim.agents.parking.maxSearchRadius = "double | 8046.72"
```

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2028)


robfitzgerald:
test!

PR #2046, opened 2019-08-11
treat param named with units of seconds like it has units of seconds

colinsheppard:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2046)


wrashid:
Test!

wrashid:
will be replaced by #2048

PR #2048, opened 2019-08-12
Hl/#2017 review rh enroute

htmlrb:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2048)


htmlrb:
test!

src/main/resources/beam-template.conf
```diff
@@ -258,6 +258,11 @@ beam.replanning.Module_4 = "TimeMutator"
 beam.replanning.ModuleProbability_4 = 0.0
 beam.replanning.fractionOfIterationsToDisableInnovation = "double | Double.PositiveInfinity"
 
+#Skimmer
+beam.beamskimmer.writeObservedSkims = "boolean | true"
+beam.beamskimmer.writeAllModeSkimsForPeakNonPeakPeriods = "boolean | true"
+beam.beamskimmer.writeObservedSkimsPlus = "boolean | true"
```
colinsheppard: Is there a reason we are using booleans instead of an "interval" param like all of our other output controls?

colinsheppard: If no, then we should make this an interval where 0 is "false" and a positive integer means only write every nth interval. Also, at a minimum you should document what "observerdSkimsPlus" means right here in the config template so others can understand without having to read the source code.

src/main/scala/beam/agentsim/agents/ridehail/AlonsoMoraPoolingAlgForRideHail.scala
```diff
@@ -339,7 +340,8 @@ object AlonsoMoraPoolingAlgForRideHail {
                 Trip(theActivity, None, null),
                 BeamMode.RIDE_HAIL,
                 Relocation,
-                leg.startTime
+                leg.startTime,
+                leg.startTime + Int.MaxValue - 30000000
```
colinsheppard: why 30000000? seems very arbitrary and confusing... it's roughly 346 days, is that meaningful?

src/main/resources/beam-template.conf
htmlrb: > Is there a reason we are using booleans instead of an "interval" param like all of our other output controls?

okay I will switch that to an integer

htmlrb: > If no, then we should make this an interval where 0 is "false" and a positive integer means only write every nth interval. Also, at a minimum you should document what "observerdSkimsPlus" means right here in the config template so others can understand without having to read the source code.

I can document that, but it is soon going to be deprecated with the new skimmer coming along

htmlrb:
test!

htmlrb:
test!

htmlrb:
test!

htmlrb:
test!

htmlrb:
test!

htmlrb:
test!

htmlrb:
test!

PR #2078, opened 2019-08-16
Rjf/#2054 rha fast charge only

### LINKED ISSUE #2054: Ride Hail Agents should only request fast charging
RHA CAVs are using the utility function sampling method to find charging alternatives, but, since the utility function does not accurately model their utility, they are choosing alternatives which are "not fast". We need a quick fix.

- Add fast charging flag to ParkingInquiry so that only those types of chargers are returned for RHA requests
- set the threshold internally to only return chargers with 20kW or greater rates

robfitzgerald:
ride hail agents, when requesting charging locations from ZPM, will only ask for fast charging.

this is assuming anyone specifying a cav ride hail depot charging file will be able to restrict to fast charging by config. procedurally generated rha depot charging will create ubiquitous 50 kWh charging.

Closes #2054.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2078)


colinsheppard:
test!

Issue #2088, opened 2019-08-18
Write Events File in Parquet Format


colinsheppard:
Please create a new BeamEventsFileFormats 

PARQUET(suffix = "parquet")

And make it an option to write events in this format. For an example of a Parquet reader already developed in BEAM, see class ParquetReader.

irishwithaxe:
implemented in https://github.com/LBNL-UCB-STI/beam/pull/2092

PR #2092, opened 2019-08-19
inm/#2088-write-events-file-in-parquet-format

irishwithaxe:
#2088

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2092)


irishwithaxe:
Test!

irishwithaxe:
If a new field added to an event or even new event added everything should be fine - all new fields will be treated like strings.

irishwithaxe:
> What happens if a new field is added to an event? Will this still function? If it fails will it crash the simulation or fail gracefully with a message?

If a new field added to an event or even new event added everything should be fine - all new fields will be treated like strings.

PR #2102, opened 2019-08-21
Add automatic processing of python scripts at iteration end

JustinPihony:


src/main/scala/beam/analysis/AnalysisProcessor.scala
```diff
@@ -0,0 +1,42 @@
...
+import com.zaxxer.nuprocess.{NuAbstractProcessHandler, NuProcess, NuProcessBuilder}
+import java.nio.ByteBuffer
+import java.nio.charset.StandardCharsets
+
+import beam.utils.logging.ExponentialLazyLogging
+
+object AnalysisProcessor extends ExponentialLazyLogging {
+
+  def iterationEndBulkAnalysisOutput_Async(currentIterationNumber: Int) = {
+    fireAndForgetPythonScript("src/main/python/events_analysis/drive_to_transit.py")
```
colinsheppard: Need to pass in the arg so the script can know what events file to handle.

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -304,6 +304,7 @@ class BeamSim @Inject()(
 
     logger.info("Ending Iteration")
     delayMetricAnalysis.generateDelayAnalysis(event)
+    beam.analysis.AnalysisProcessor.iterationEndBulkAnalysisOutput_Async(event.getIteration)
```
colinsheppard: probably better to pass the iteration output path than just the iteration

JustinPihony:
I just pushed a change that I didn't test (from the python script perspective). Some todos to optimize/clean:

* Optimize two python scripts to coordinate event file reading instead of reading twice
* Verify that events.csv is created...especially since it could be events.csv.gz
* Validate python scripts

zneedell:
> * Optimize two python scripts to coordinate event file reading instead of reading twice

We can get rid of `analyze_events` now--I moved everything important from there over to `pooling_metrics`. Although maybe now we want to rename that.

JustinPihony:
Thanks @zneedell that simplifies things :) I would think we should rename pooling_metrics to analyze_events then :)?

JustinPihony:
@colinsheppard I have this working - however in my simple tests it doesnt finish because of only one iteration. Of course I can add another iteration, but this raises the question of how we want to handle the final iteration - should BEAM exit and we can run manually, or should BEAM wait until the script is completed to exit?

src/main/scala/beam/analysis/AnalysisProcessor.scala
JustinPihony: Done

src/main/scala/beam/sim/BeamSim.scala
JustinPihony: Done

JustinPihony:
I am using `beam.outputs.writeEventsInterval` as it only makes sense to do when events are written. I guess a non-existent event file wouldn't cause a crash, but it might create some unnecessary confusion if the two values are not in sync or an interval of the other. It would be a fairly simple change to go to what you propose, though @colinsheppard

JustinPihony:
test!

JustinPihony:
test!

PR #2103, opened 2019-08-21
change park protocol to allow release of stall when fast charging done

colinsheppard:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2103)


colinsheppard:
This is passing CI but still needs to be tested on a couple of production scenarios.

PR #2119, opened 2019-08-25
Cs/skim no zero

colinsheppard:
Skimmer shouldn't return distance 0 for intra TAZ results.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2119)


wrashid:
Test!

PR #2127, opened 2019-08-27
rjf/#2126 more encouragement for PEV home charging

### LINKED ISSUE #2126: add range cost of "tomorrow" to PEVs headed home
while we don't have "tomorrow", we do have the first leg of our day. we can use that when considering range anxiety, to encourage agents to park at charging stalls when they are headed home.

robfitzgerald:
added two features to help encourage PEV agents to charge at home:

1. when headed home, there is no remaining tour. tours help create range anxiety, which encourages charging. in real life, this agent would have plans for the next day as well, and they would consider that. to simulate, we use the first car trip of the agent's day and add it in as a remaining tour distance (along with the leg they are currently on).
2. all agents who are 1) headed home, 2) to a residential stall have been getting a boost to their parking stall's utility. i split this out so that, in the case of electric vehicles, they only get the boost when the stall has a charger.

this is a draft while tests are running on these features, and feedback would help in case anyone can think of issues with these features.

Closes #2126.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2127)


JustinPihony:
test!

PR #2140, opened 2019-08-31
look at trips with mode ride_hail_pooled in events analysis

zneedell:
Average wait times and the ride hail waiting time duration plots were only looking at mode choice events with mode `ride_hail` not `ride_hail_pooled`

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2140)


wrashid:
Test!

PR #2141, opened 2019-08-31
Jhrf/#2135 parking utility event

johanneshiry:
First implementation of a ParkingUtilityEvent for analyzing and debugging purposes of the parking zone seach. Focus/main goal is to get more insights in the search results for pev charging.

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -24,14 +24,14 @@ import beam.sim.{BeamServices, Geofence}
 import beam.sim.population.AttributesOfIndividual
 import beam.utils.plan.sampling.AvailableModeUtils._
 import com.vividsolutions.jts.geom.Envelope
-import org.matsim.api.core.v01.population.{Activity, Leg}
+import org.matsim.api.core.v01.population.{Activity, Leg, Person}
 import org.matsim.api.core.v01.{Coord, Id}
 import org.matsim.core.population.routes.NetworkRoute
 import org.matsim.core.utils.misc.Time
 import org.matsim.vehicles.Vehicle
+
 import scala.concurrent.duration._
 import scala.concurrent.{ExecutionContext, Future}
-
 import beam.agentsim.infrastructure.parking.ParkingMNL
 
 /**
@@ -776,7 +776,8 @@ trait ChoosesMode {
       remainingTripData,
       attributes.valueOfTime,
       duration,
-      reserveStall = false
+      reserveStall = false,
+      currentDriverId = id.toString
     )
     parkingManager ! inquiry
     Some(inquiry.requestId)
--- a/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
+++ b/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
@@ -6,13 +6,12 @@ import beam.agentsim.agents.BeamAgent._
 import beam.agentsim.agents.PersonAgent._
 import beam.agentsim.agents._
 import beam.agentsim.agents.choice.logit.{MultinomialLogit, UtilityFunctionOperation}
-import beam.agentsim.agents.modalbehaviors.DrivesVehicle.{StartLegTrigger}
+import beam.agentsim.agents.modalbehaviors.DrivesVehicle.StartLegTrigger
 import beam.agentsim.agents.parking.ChoosesParking.{ChoosingParkingSpot, ReleasingParkingSpot}
 import beam.agentsim.agents.vehicles.FuelType.{Electricity, Gasoline}
 import beam.agentsim.agents.vehicles.VehicleProtocol.StreetVehicle
-import beam.agentsim.agents.vehicles.{PassengerSchedule}
+import beam.agentsim.agents.vehicles.PassengerSchedule
 import beam.agentsim.events.{LeavingParkingEvent, SpaceTime}
-
 import beam.agentsim.scheduler.BeamAgentScheduler.{CompletionNotice, ScheduleTrigger}
 import beam.agentsim.scheduler.Trigger.TriggerWithId
 import beam.router.BeamRouter.{RoutingRequest, RoutingResponse}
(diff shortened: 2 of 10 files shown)


src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -24,14 +24,14 @@ import beam.sim.{BeamServices, Geofence}
 import beam.sim.population.AttributesOfIndividual
 import beam.utils.plan.sampling.AvailableModeUtils._
 import com.vividsolutions.jts.geom.Envelope
-import org.matsim.api.core.v01.population.{Activity, Leg}
+import org.matsim.api.core.v01.population.{Activity, Leg, Person}
```
robfitzgerald: you don't need to add this `Person` import since you are `toString`'ing the Id below

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -66,7 +67,8 @@ trait ChoosesParking extends {
         this.currentTourBeamVehicle,
         remainingTripData,
         attributes.valueOfTime,
-        parkingDuration
+        parkingDuration,
+        currentDriverId = id.asInstanceOf[Id[Person]].toString
```
robfitzgerald: what does the `.asInstanceOf[Id[Person]]` give you here? i think you can probably remove it.

src/main/scala/beam/agentsim/events/ParkingUtilityEvent.scala
```diff
@@ -0,0 +1,124 @@
...
+  driverId: String,
+  beamVehicle: Option[BeamVehicle],
+  activityType: String,
+  parkingDuration: Double,
+  agentVoT: Double,
+  parkingZoneSearchStats: ParkingZoneSearchStats,
+  selectedStallPrice: Double,
+  selectedStallParkingType: ParkingType,
+  selectedStallChargingPointType: Option[ChargingPointType],
+  tick: Double = -1 // todo maybe we need the currentTick? If yes we have to get it into ZonalParkingManager
```
robfitzgerald: yes. i do think we need the current tick here in order to make sense of what we observe, for example, to compare against trends in ParkEvents and RefuelSessionEvents (estimated vs experienced parking utility).

we can probably send it via the ParkingInquiry from the various spots where we send those for PEVs and ride hail.

src/main/scala/beam/agentsim/events/ParkingUtilityEvent.scala
```diff
@@ -0,0 +1,124 @@
...
+) extends Event(tick)
+    with ScalaEvent {
+
+  import ParkingUtilityEvent._
+
+  private lazy val sampledStallsChargingTypeDist: String = "[" + parkingZoneSearchStats.sampledStallsChargingTypes
```
robfitzgerald: regarding `sampledStallsChargingTypeDist | sampledStallsParkingTypeDist | sampledStallsCostsDist`,

i'm a little worried about this approach and how much clutter it will add, and how it won't be easy to convert for use in data analysis. i know, when i started logging parking utility data, i was grabbing this kind of stuff, but it was mostly for debug reasons. maybe we leave these out from the Event?

i think if we want to study this kind of data, we should make additional events for it. a `SampledStallsChargingTypeDist` event, etc., though i would want to know what kinds of analysis tasks that would support, i.e., what questions we are asking to necessitate grabbing that data.

src/main/scala/beam/agentsim/events/ParkingUtilityEvent.scala
```diff
@@ -0,0 +1,124 @@
...
+  private lazy val sampledStallsCostsDist: String = "[" +
+  parkingZoneSearchStats.sampledStallsCosts
+    .groupBy(identity)
+    .mapValues(_.size)
+    .map(tuple => tuple._1 + ": " + tuple._2)
+    .mkString(",") + "]"
+
+  private lazy val vehIdString: String = beamVehicle match {
+    case Some(vehicle) => vehicle.id.toString
+    case None          => "no beamVehicle in parking inquiry"
```
robfitzgerald: i suggest "" for the None case, i think that's the convention in Beam Events.

robfitzgerald: and, for all "None" cases following, the same.

src/main/scala/beam/agentsim/events/ParkingUtilityEvent.scala
robfitzgerald: and, i also see that this would mean removing like 50-60 lines of your work, so, let me say that i appreciate your efforts on this. i just don't know that we need that data or would want it in that form?

src/main/scala/beam/agentsim/events/ParkingUtilityEvent.scala
johanneshiry: Agree. I tried to implement it exactly like this, but parkingInquiry is send from several places and not everywhere it is send from is the current tick available. 
=> it would need some bigger refactoring in several places

But I'm going to adapt theses places as well then

src/main/scala/beam/agentsim/events/ParkingUtilityEvent.scala
johanneshiry: Agree that it won't be easy to convert this kinda data into a usable format for data analysis. I left it inside mainly because of the reason that I assumed we initially wanted this event for debugging purposes. 

I think a `SampledStallsChargingTypeDist`event wouldn't add much value for data analysis - only for debugging purposes and to understand what's going on in the sampling. So maybe we just drop it for now?

johanneshiry: Okay I just dropped it for now in it's own commit 

`ParkingZoneSearch`still includes `parkingZoneIdsSampled` which hasn't been mentioned in your review as dropping candidate. As this is part of the existing code, let me know if you think this param adds some value on data analysis steps and I'll include it into the event or drop it otherwise as well.

src/main/scala/beam/agentsim/events/ParkingUtilityEvent.scala
johanneshiry: done

PR #2146, opened 2019-09-02
scale utilities in mode choice

zneedell:
Adds a scaling factor to multinomial logit, which lets you scale how important the random noise is in the mode choice probabilities. Right now this is only implemented in mode choice, but it should be straightforward to add to parking if needed. 

As this factor gets large, the choice becomes more deterministic towards the option with the highest utility, as it gets small the choice becomes closer to uniform between the available options. 1 is default. Note that for large values (~= 100), the exponential utilities can get very small and numerical issues might come up.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2146)


JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

Issue #2153, opened 2019-09-06
why is generalized time sometimes >> time?


zneedell:
In some of the skims from the production runs we're seeing generalized time values that are 10+ times greater than the observed time values. Based on the VOTT input parameters, we would only expect generalized time to be at a max 2x greater than the real time.

PR #2155, opened 2019-09-06
Fix inflated generalized times (#2153)

zneedell:
We've observed some generalized times in the skims that are higher than they should be based on the config VOTT multipliers, implying that generalized times are getting artificially inflated somewhere. 

I think this is because generalized TT is calculated by looping over all links in car paths, whereas the driver does not necessarily need to drive the full distance of the first and last link. By matsim convention this means ignoring the first link when calculating travel time and distance.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2155)


zneedell:
test!

PR #2163, opened 2019-09-10
calculate distance along non-road transit routes

zneedell:
so that beamLegs for transit no longer have 0 distance

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2163)


zneedell:
Would be great for someone to take a pretty close look at this if possible--it was a bit of a shot in the dark, and even if it is correct there might be much faster ways of doing it.

PR #2175, opened 2019-09-16
Add mode chosen available alternative count

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2175)


src/main/java/beam/analysis/plots/ModeChosenAnalysis.java
```diff
@@ -347,4 +374,40 @@ public void writeToRootCSVForReference(String fileBaseName) {
...
+                return false;
+            }
+
+            ModeChosenAvailableAlternatives modeChosenAvailableAlternatives = (ModeChosenAvailableAlternatives) o;
+
+            return modeChosenAvailableAlternatives.mode.equals(mode) &&
+                    modeChosenAvailableAlternatives.availableModes.equals(availableModes);
+        }
+
+        @Override
```
wrashid: what is this?

rajnikantsh: i have to override java hashCode and equals method because this class is used as key in map

PR #2176, opened 2019-09-17
Consider commute in whether to retire household vehicles

zneedell:
We want to prioritize retiring household vehicles in households that have more plentiful other commute options.

Part 1 is prioritizing removing vehicles from people with short commutes. This is a bit tricky because there is not a 1:1 relationship between workers and vehicles.

Part 2 is making it less likely to remove someone's vehicle if they take a trip when it falls outside of typical transit operating hours. This is done by penalizing that person with the equivalent of an extra 5 km of commute distance per trip that can't be taken on transit.

Appears to produce expected results on production-sfbay

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2176)


zneedell:
test!

zneedell:
Test!

wrashid:
Test!

wrashid:
Test!

wrashid:
Test!

PR #2203, opened 2019-10-01
Add Repositioning Analysis

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2203)

### DIFF
--- a/src/main/scala/beam/utils/RespositioingValidation.scala
+++ b/src/main/scala/beam/utils/RespositioingValidation.scala
@@ -2,10 +2,8 @@ package beam.utils
 
 import beam.agentsim.events.PathTraversalEvent
 import org.matsim.api.core.v01.events.Event
-import org.matsim.api.core.v01.{Coord, Id}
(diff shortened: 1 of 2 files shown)


src/main/scala/beam/utils/RespositioingValidation.scala
```diff
@@ -0,0 +1,98 @@
...
+  }
+
+  private def readCSVEvents(path: String): (Int, Int) = {
+    val (eventSeq, _) = EventReader.fromCsvFile(path, _ => true)
+    var eventCount = 0
+    var foundEvent = 0
+    eventSeq.foreach(event => {
+      if (event.getEventType == "PathTraversal") {
+        eventCount += 1
+        if (processForRepositioningDebug(event))
```
JustinPihony: Why does this one not `PathTraversalEvent.apply`?

rajnikantsh: Make the changes as per above comment

PR #2214, opened 2019-10-11
Stagger initial tick for repositioning

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2214)

### DIFF
--- a/src/main/scala/beam/sim/BeamMobsim.scala
+++ b/src/main/scala/beam/sim/BeamMobsim.scala
@@ -315,7 +315,7 @@ class BeamMobsimIteration(
 
   private def scheduleRideHailManagerTimerMessages(): Unit = {
     if (config.agents.rideHail.repositioningManager.timeout > 0)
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/sim/BeamMobsim.scala
```diff
@@ -314,8 +314,13 @@ class BeamMobsimIteration(
   }
 
   private def scheduleRideHailManagerTimerMessages(): Unit = {
-    if (config.agents.rideHail.repositioningManager.timeout > 0)
-      scheduler ! ScheduleTrigger(RideHailRepositioningTrigger(0), rideHailManager)
+    if (config.agents.rideHail.repositioningManager.timeout > 0) {
```
REASY: @colinsheppard @wrashid do you think this comment is enough to understand why it is done in this way?

colinsheppard: that's fine... I will eventually solve this so that stagger is no longer needed but this will do for now

PR #2234, opened 2019-10-20
Ability to process RoutingResponse

REASY:
- Added `RoutingResponse.request` to have request-response pair
- Added `RoutingRequest.initiatedFrom` to keep track of the places where `RoutingRequest` was created
- `BeamRouter` sends `RoutingResponseEvent` to events manager
- Added `GeofenceAnalyzer` as an example of `RoutingResponseEvent` handler

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2234)


REASY:
test!

REASY:
Test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

src/main/scala/beam/router/BeamRouter.scala
```diff
@@ -211,6 +214,7 @@ class BeamRouter(
         sendWorkTo(worker, work, originalSender, receivePath = "GimmeWork")
       }
     case routingResp: RoutingResponse =>
+      eventsManager.processEvent(RoutingResponseEvent(routingResp))
```
JustinPihony: Is there any noticeable overhead from this?

src/test/scala/beam/agentsim/agents/PersonAndTransitDriverSpec.scala
```diff
@@ -404,7 +404,8 @@ class PersonAndTransitDriverSpec
             )
           )
         ),
-        requestId = 1
+        requestId = 1,
+        request = None
```
JustinPihony: Rather than making all of these changes required should we have the default be `None`?

REASY: I thought about this in the first place, but then it would make compiler happy and I could skip important places where I had to change

src/main/scala/beam/router/BeamRouter.scala
REASY: Good point. I don't think it will, but I can run prod run before actual merge.

src/test/scala/beam/agentsim/agents/PersonAndTransitDriverSpec.scala
JustinPihony: I figured that might be the case...sgtm.

JustinPihony:
@REASY Is this ready to merge?

JustinPihony:
A bunch of `java.lang.ArithmeticException: / by zero`

REASY:
@JustinPihony thanks, I fixed it locally, but didn't push it, let me do it now, commit 3f43becdf4c6f87694714d186f5671dffd8c6263

JustinPihony:
@REASY Seems this is ready - except it has some merge conflicts which need pulled in?

PR #2249, opened 2019-10-30
Add csv for non arrived agents

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2249)

### DIFF
--- a/src/main/java/beam/analysis/plots/PersonTravelTimeAnalysis.java
+++ b/src/main/java/beam/analysis/plots/PersonTravelTimeAnalysis.java
@@ -15,6 +15,8 @@ import org.matsim.api.core.v01.population.Person;
 import org.matsim.core.controler.OutputDirectoryHierarchy;
 import org.matsim.core.controler.events.IterationEndsEvent;
 import org.matsim.core.utils.collections.Tuple;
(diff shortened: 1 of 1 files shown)


src/main/java/beam/analysis/plots/PersonTravelTimeAnalysis.java
```diff
@@ -311,6 +312,25 @@ private void createNonArrivalAgentAtTheEndOfSimulationGraph(int iterationNumber)
...
+            out.newLine();
+            Set<String> modes = personLastDepartureEvents.keySet();
+            for(String mode: modes){
+                Map<Id<Person>, PersonDepartureEvent> personDepartureEventMap = personLastDepartureEvents.get(mode);
+                out.write(mode+","+personDepartureEventMap.size());
+                out.newLine();
+            }
+            out.flush();
+        } catch (IOException e) {
+            e.printStackTrace();
```
REASY: Please, use logger to log an exception

rajnikantsh: use logger for exception

PR #2253, opened 2019-10-31
Art/prod error for r5

REASY:
This is just to show the difference

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2253)


JustinPihony:
Was this just for show? Should it be closed or will it become a full PR at some point?

PR #2266, opened 2019-11-06
inm/script to transform BEAM events file into MATSim events file

irishwithaxe:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2266)


irishwithaxe:
Test!

irishwithaxe:
Test!

src/main/scala/beam/utils/beamToVia/IO/Reader.scala
```diff
@@ -0,0 +1,241 @@
+package beam.utils.beamToVia.IO
```
REASY: Please, do not use upper case in the namespace because it will cause some issue during launching the app run in Linux (case sensitive file system). By JAVA convention [Naming a Package](https://docs.oracle.com/javase/tutorial/java/package/namingpkgs.html): Package names are written in all lower case to avoid conflict with the names of classes or interfaces.

src/main/scala/beam/utils/beamToVia/IO/Writer.scala
```diff
@@ -0,0 +1,108 @@
...
+    pw.println("</events>")
+    pw.close()
+
+    progress.finish()
+
+    Console.println("via events written into " + outputPath)
+  }
+
+  def writeSeqOfString(script: Traversable[String], outputPath: String): Unit = {
+    val pw = new PrintWriter(new File(outputPath))
```
REASY: If exceptions happen the stream will remain opened. It's always good to make sure that resources are closed every time. You could use `FileUtils.using` for that purpose.

src/main/scala/beam/utils/beamToVia/appsForVisualizations/visualization_11.scala
```diff
@@ -0,0 +1,71 @@
+package beam.utils.beamToVia.appsForVisualizations
```
REASY: Does it make sense to attach some document/actual vidoe what does visualization 11 and others mean?

src/main/scala/beam/utils/beamToVia/IO/Reader.scala
irishwithaxe: ok

src/main/scala/beam/utils/beamToVia/IO/Writer.scala
irishwithaxe: ok, will use FileUtils.using

src/main/scala/beam/utils/beamToVia/appsForVisualizations/visualization_11.scala
irishwithaxe: a good question ..

irishwithaxe:
Test!

irishwithaxe:
Test!

PR #2289, opened 2019-11-22
Idle vehicle should be in range by remainingPrimaryRangeInM to be able to serve customer

REASY:
`getClosestIdleVehiclesWithinRadiusByETA` makes sure that idle vehicle has enough fuel (with some marginFactor) to make a trip to the customer, serve customer and go to charging.
Total travel distance can be represented graphically:
Untitled Diagram

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2289)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
+++ b/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
@@ -18,8 +18,9 @@ import beam.sim.BeamScenario
 import beam.sim.common.GeoUtils.TurningDirection
 import beam.utils.NetworkHelper
 import beam.utils.logging.ExponentialLazyLogging
-import org.matsim.api.core.v01.Id
+import org.matsim.api.core.v01.{Coord, Id}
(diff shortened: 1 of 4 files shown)


src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -364,11 +365,28 @@ object BeamVehicle {
...
 
+  object BeamVehicleState {
+
+    def haveEnoughFuel(
+      beamVehicleState: BeamVehicleState,
+      rideHailLocationUTM: Coord,
+      pickupLocationUTM: Coord,
+      travelDistance: Double,
+      distanceToDepot: Double,
+      marginFactor: Double = 2.0 // TODO Better name? Extract to config?
```
wrashid: we need to be able to control this through a parameter in config fiile

wrashid: e.g. beam.agentsim.agents.rideHail.rangeSafetyMarginFactor

PR #2295, opened 2019-12-06
Bl/#2294/geofence is being violated 4ci

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2295)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
@@ -378,7 +378,7 @@ class RideHailAgent(
       updateLatestObservedTick(tick)
       log.debug("state(RideHailingAgent.Offline.EndRefuelTrigger): {}", ev)
       holdTickAndTriggerId(tick, triggerId)
-      handleEndRefuel(energyInJoules, tick, sessionStart.toInt)
+      handleEndRefuel(energyInJoules, tick, sessionStart.toInt, geofence)
       goto(Idle)
     case ev @ Event(TriggerWithId(StartLegTrigger(_, _), triggerId), data) =>
       log.warning(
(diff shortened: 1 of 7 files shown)


src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -624,7 +624,7 @@ class RideHailAgent(
 
   override def logPrefix(): String = s"RideHailAgent $id: "
 
-  def handleEndRefuel(energyInJoules: Double, tick: Int, sessionStart: Int): Unit = {
+  def handleEndRefuel(energyInJoules: Double, tick: Int, sessionStart: Int, geofence: Option[Geofence]): Unit = {
```
REASY: @bhavyalatha26 I'm not sure why we need to pass param `geofence` because it is already in the class fields of `RideHailAgent` class.
@colinsheppard what is the idea behind passing geofence in `NotifyVehicleIdle` class? Is it supposed to be dynamic? I see this as just extra steps which is not needed. Instead we can make sure that RHM knows about vehicle geofence and assign them accordingly.

bhavyalatha26: ok will make the change @REASY

colinsheppard: Good question... right now we don't treat geofences as dynamic, so we could change their treatment to static and nothing would break. 

I can imagine, however, a future situation where we would want to simulate something dynamic. For example, Uber already has a feature that allows a driver who is far from home to get matched with customers that are going in the direction of home, allowing the driver to make their way back while still serving rides. This kind of functionality is similar to geofencing and if/when we decide to implement it, it could probably be done through that mechanism. 

So I think we should leave it for now unless we are finding that it is somehow impacting performance or making other parts of the code more difficult to maintain and debug.

PR #2302, opened 2019-12-11
zn/#2301 fix charging power plots

### LINKED ISSUE #2301: Charging power plots aren't correct
The plots showing charging power over the day (by charger type, parking type, and user) can show average power values that are greater than the number of chargers that exist * their rated power.

I think this is because there is a bug in how loads from different charging events are averaged within hour bins.

zneedell:
The average load during an hour in kW should be the sum of the energy delivered (in kWh) of all charging events during that hour. This still doesn't handle charging sessions that span multiple hour bins, but that effect shouldn't be too big with fast chargers.

Fixes #2301

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2302)


src/main/scala/beam/analysis/LoadOverTimeAnalysis.scala
```diff
@@ -33,20 +33,18 @@ class LoadOverTimeAnalysis extends GraphAnalysis with ExponentialLazyLogging {
                 .contains("ridehail")) {
             if (vehicleType.isCaccEnabled) "CAV RideHail" else "Human RideHail"
           } else "Personal"
-        val energyInJoules = refuelSessionEvent.energyInJoules
-        val sessionDuration = refuelSessionEvent.sessionDuration
-        val currentEventAverageLoad = if (sessionDuration != 0) energyInJoules / sessionDuration / 1000 else 0
+        val energyInkWh = refuelSessionEvent.energyInJoules / 3.6e6
```
JustinPihony: Doesn't session duration need to be factored because we want the average...or do we want the total energy?

zneedell: Would be good to get another set of eyes here. @colinsheppard or @wrashid maybe?

I think that session duration is being factored in implicitly because the time bins have fixed length. So if we count the number of kWh (energy) during a time bin and then divide by the bin duration (1 hr, so not explicitly done here) we get the average rate of power consumption, in kW.

JustinPihony: Yah, but isn't session duration often less than the hour bin?

zneedell: Right--let's say we have a 30 minute charging session that delivers 10 kWh. If that's the only session, the charger was delivering 20 kW for 30 minutes and 0 kW for 30 minutes, so it had an average load of 10 kW for the hour. I guess this isn't the only reasonable definition for average charging power, but I _think_ this one will end up being the most useful.

PR #2307, opened 2019-12-14
Procedural charging locations (for demo)

zneedell:
IN PROGRESS

Generates charging stall locations to minimize distances from charging events using a method based off of FCS-Plan, working separately for CAV and Human driven ride hail. For now, the parameters that can vary are:

- **Number of charging stations**: These are scattered across the simulation area to minimize average distance to the nearest charging event. Each station can have multiple plugs
- **Charging power**: DC power at each plug
- **Maximum Queuing Probability**: For every charging station, the number of plugs is determined based on the probability that a new user will find that all plugs are occupied during the busiest hour of the day (for now using the [Erlang C formula](https://en.wikipedia.org/wiki/Erlang_(unit)#Erlang_C_formula), although that is not perfect). The number of plugs is set to the smallest value that keeps this probability below the threshold parameter. 

Right now, it is set up to generate depot and public parking files for ~20 different numbers of charging stations, 2 values for charging power, and 3 values for the max queuing probability, but this can be changed. Summary metrics are the average distance from a charging event (given unlimited infrastructure) and the nearest charger and the number of plugs.

tradeoffs

The idea is that a user could look at this plot and determine which of these three variables to use, and then the user could pick out the parking file generated that matches the values of these three variables and move it over to the inputs for a new BEAM run.

Todo:
- [ ] Make new (non-charging) parking files to account for limited population
- [ ] Plots of charger locations?
- [ ] Automate running the python script at the end of a beam run
- [ ] Check how robust it is to different simulation inputs

It also makes a plot of charging demand, but this could definitely be improved, and in general this would also probably benefit from some more interesting visualizations of charging activity (in space and in time).

chargind_demand_map


JustinPihony:
A reminder that you can use the "Open as Draft" feature for WIPs. I wish you could revert back once put into final review mode.

zneedell:
Updated maps: 
charger_locations
public

Still to do:

- [ ] Figure out why the contextily module isn't always working
- [ ] Adjust BEV range from config?

PR #2308, opened 2019-12-15
Add EBS Volume Size Notifier

### LINKED ISSUE #2290: Create an alarm to make sure EBS storage costs do not rise too high again
Maybe the current process changes are good enough, but it's worth noting the possibility of something as a CYA thing.

JustinPihony:
Closes #2290 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2308)


JustinPihony:
Yah, boto3 is pretty much a simple wrapper around the aws API - using similar syntax as the aws cli. Also, that notification was when I had the size check backwards (for testing) - but has been fixed so that it should only report when greater than the max :) - just noting for posterity

PR #2315, opened 2019-12-27
Rh matching fix

haitamlaarabi:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2315)


wrashid:
Test!

haitamlaarabi:
Test!

src/main/scala/beam/agentsim/agents/ridehail/MatchmakingUtils.scala
```diff
@@ -0,0 +1,394 @@
...
+    calc.setDestinationGeographicPoint(point3.getX, point3.getY)
+    val azimuth2 = calc.getAzimuth
+    val degrees = azimuth2 - azimuth1
+    Math.abs(degrees) < 45.0
+  }
+
+  def checkDistance(r: MobilityRequest, schedule: List[MobilityRequest], searchRadius: Double): Boolean = {
+    schedule.foreach { s =>
+      if (GeoUtils.distFormula(r.activity.getCoord, s.activity.getCoord) <= searchRadius)
+        return true
```
JustinPihony: returns in scala are *very* much a no no. Not just for syntax but functionality can be different than you expect. Here I would use `exists`:

```
schedule.exists(s => GeoUtils.distFormula(r.activity.getCoord, s.activity.getCoord) <= searchRadius)
````

That should be the only thing you need.

File: src/main/scala/beam/agentsim/agents/ridehail/MatchmakingUtils.scala [hunk 2]
```diff
@@ -0,0 +1,394 @@
...
+    val center = requestWithCurrentVehiclePosition.activity.getCoord
+
+    if (requestWithCurrentVehiclePosition.tag == EnRoute) {
+      // if vehicle is EnRoute, then filter list of customer based on the destination of the passengers
+      val i = v.schedule.indexWhere(_.tag == EnRoute)
+      val mainTasks = v.schedule.slice(0, i)
+      demand
+        .filter(
+          r =>
+            mainTasks.foldLeft(false) { (acc, t) =>
```
JustinPihony: Do you want everything reviewed, or would an exists be more appropriate here? It seems that way to me.

src/main/scala/beam/agentsim/agents/ridehail/MatchmakingUtils.scala
```diff
@@ -0,0 +1,394 @@
...
+      //.sortBy(r => GeoUtils.minkowskiDistFormula(center, r.pickup.activity.getCoord))
+    } else {
+      // if vehicle is empty, prioritize the destination of the current closest customers
+      val customers = demand.sortBy(r => GeoUtils.minkowskiDistFormula(center, r.pickup.activity.getCoord))
+      val mainRequests = customers.slice(0, Math.min(customers.size, v.getSeatingCapacity))
+      mainRequests ::: customers
+        .drop(mainRequests.size)
+        .filter(
+          r =>
+            mainRequests.foldLeft(false) { (acc, t) =>
```
JustinPihony: Same, I think an exists is all you want/need here?

src/main/scala/beam/agentsim/agents/ridehail/MatchmakingUtils.scala
```diff
@@ -0,0 +1,394 @@
...
+          skimmer.beamScenario.beamConfig.beam.agentsim.agents.rideHail.initialization.procedural.vehicleTypeId,
+          classOf[BeamVehicleType]
+        )
+      )
+      val serviceTime = prevReq.serviceTime + tdc.time
+      val serviceDistance = prevReq.serviceDistance + tdc.distance.toInt
+      if (serviceTime <= curReq.upperBoundTime && serviceDistance <= remainingVehicleRangeInMeters) {
+        newPoolingList.append(curReq.copy(serviceTime = serviceTime, serviceDistance = serviceDistance))
+      } else {
+        return None
```
JustinPihony: Another return. Please find another way to solve this. I would go with `forall` instead:

```
val isValid = sortedRequests.forall(curReq =>
    ...
    if (serviceTime <= curReq.upperBoundTime && serviceDistance <= remainingVehicleRangeInMeters) {
        newPoolingList.append(curReq.copy(serviceTime = serviceTime, serviceDistance = serviceDistance))
        true
    } else {
        false
    }
)
if isValid Some(newPoolingList.toList) else None
```

haitamlaarabi:
Test!

haitamlaarabi:
Test!

src/main/scala/beam/agentsim/agents/ridehail/MatchmakingUtils.scala
```diff
@@ -147,25 +148,29 @@ object MatchmakingUtils {
...
-        prevReq.activity.getCoord,
-        curReq.activity.getCoord,
-        prevReq.baselineNonPooledTime,
-        BeamMode.CAR,
-        Id.create(
-          skimmer.beamScenario.beamConfig.beam.agentsim.agents.rideHail.initialization.procedural.vehicleTypeId,
-          classOf[BeamVehicleType]
+
+    var isValid = true
+    breakable {
```
JustinPihony: This works - although I am curious - was the code I put up not working? I wrote it in the browser, so it wasn't tested...but it seemed right :)

wrashid:
Test!

src/main/scala/beam/agentsim/agents/ridehail/MatchmakingUtils.scala
haitamlaarabi: yes, it is right, but it slows it down. breaking out of the loop as soon as the matching is invalid is critical.

JustinPihony: Interesting...it should break:

```
scala> List(true, true, false, true).forall(x=>{println(x);x})
true
true
false
res0: Boolean = false

scala> List(true, true, true, true).forall(x=>{println(x);x})
true
true
true
true
res1: Boolean = true
```

haitamlaarabi:
Test!

PR #2357, opened 2020-02-04
Ez/#2350 census database

### LINKED ISSUE #2350: Build a database from Census Tract State-based Shapefile files
We need to use [US Cenus Bureau](https://www.census.gov/) data to build a database of USA. We're interested in [Census Tract State-based Shapefile Record Layout](https://www2.census.gov/geo/pdfs/maps-data/data/tiger/tgrshp2019/TGRSHP2019_TechDoc.pdf):

image

Terms dictionary:

- FIPS: Federal Information Processing Standards
- [Census tracts](https://transition.fcc.gov/form477/Geo/more_about_census_tracts.pdf): Census tracts are geographic entities within counties (or the statistical equivalent of counties). The entire area of a county is covered by census tracts, just as the entire area of a state is covered by counties and county equivalents.

Useful links: 

- [Concept & Definitions](https://www.census.gov/programs-surveys/acs/geography-acs/concepts-definitions.html)
- [Actual census tract data for 2019](https://www2.census.gov/geo/tiger/TIGER2019/TRACT/) 

The database should contains the following fields:
- state_fp: State FIPS code
- county_fp: County FIPS code
- tract_code: Census tract code
- geo_id: Census tract identifier; a concatenation of state FIPS code, county FIPS code, and census tract code
- land_area: Land area
- water_area: Water area
- multi_polygon: Multi-polygon of that census tract

We need to have a way to query the data in at least two ways:
- Provide bounding box and get back all the records which are in that bounding box
- Provide state and county FIPS code and get back all the records which belongs to that state and county

There can be multiple ways of implementing this database, some of them:
 
- Use PostgreSQL with [PostGIS](https://postgis.net/) which has capabilities for geo spatial queries
- Use Elasticsearch with [geo queries](https://www.elastic.co/guide/en/elasticsearch/reference/current/geo-queries.html)

SHaaD94:
Resolves #2350 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2357)


JustinPihony:
Test!

src/main/scala/beam/taz/PointGenerator.scala
```diff
@@ -0,0 +1,43 @@
...
+    def generate(n: Int): Seq[Coord] = {
+      pointsBuilder.setNumPoints(n)
+      val multiPoint = pointsBuilder.getGeometry.intersection(geometry)
+
+      val points = (0 until multiPoint.getNumGeometries)
+        .map(multiPoint.getGeometryN)
+        .filter(x => x.getGeometryType.contains("Point"))
+
+      if (points.size >= nPoints)
+        return points
```
JustinPihony: return is dangerous in scala - use an if/else with each returning accordingly

src/main/scala/beam/taz/PointGenerator.scala
```diff
@@ -0,0 +1,43 @@
...
+
+  override def generate(geometry: Geometry, nPoints: Int): Seq[Coord] = {
+    val pointsBuilder = new RandomPointsInGridBuilder(geometryFactory)
+    pointsBuilder.setExtent(geometry.getEnvelopeInternal)
+
+    @tailrec
+    def generate(n: Int): Seq[Coord] = {
+      pointsBuilder.setNumPoints(n)
+      val multiPoint = pointsBuilder.getGeometry.intersection(geometry)
+
```
JustinPihony: Have you tested the performance of this? As we are messing with large scales at times we need to be cognizant of that. I wonder if a view would be useful since you end up using a take in the longrun. No need to constantly run the code if we can put it off until the final point and reduce CPU cycles.

src/main/scala/census/db/creator/DatabaseCreator.scala
```diff
@@ -0,0 +1,55 @@
...
+
+  val orchestrator = system.actorOf(Props(new OrchestrationActor()))
+  val tazWriter = system.actorOf(Props(new TazSavingActor(Hardcoded.config, orchestrator)))
+  val shapeReader = system.actorOf(Props(new ShapeReadingActor(Hardcoded.config, tazWriter)))
+  val unzipper = system.actorOf(Props(new UnzipActor(Hardcoded.config, shapeReader)))
+  val downloader = system.actorOf(Props(new DownloadActor(Hardcoded.config, unzipper)))
+
+  for {
+    files <- new FileDownloadService(Hardcoded.config).getFileNames()
+  } yield {
```
JustinPihony: No need for yield since you are not outputting anything.

src/main/scala/census/db/creator/DatabaseCreator.scala
```diff
@@ -0,0 +1,55 @@
...
+
+/*
+
+RUN POSTGIS DOCKER WITH FOLLOWING COMMAND:
+
+docker run --name postgis -e POSTGRES_PASSWORD=postgres1 -e POSTGRES_DB=census -p 5432:5432 -d mdillon/postgis
+
+ */
+
+private[creator] object DatabaseCreator extends App {
```
JustinPihony: App is generally not suggested btw - it's probably fine here, but also good practice to build your own main method

src/main/scala/census/db/creator/GeometryUtil.scala
```diff
@@ -0,0 +1,27 @@
+package census.db.creator
+import com.vividsolutions.jts.geom._
+import com.vividsolutions.jts.io.{WKBReader, WKTReader}
+
+object GeometryUtil {
+  val projection = 4326
+  val geometryFactory = new GeometryFactory(new PrecisionModel(), projection)
+
+  private val wktReader = new WKTReader(geometryFactory)
```
JustinPihony: What is wkt and wkb - can you expand the naming to help future readers of the code?

src/main/scala/census/db/creator/domain/Domain.scala
```diff
@@ -0,0 +1,12 @@
+package census.db.creator.domain
+import com.vividsolutions.jts.geom.prep.PreparedGeometry
+
+case class TazInfo(
```
JustinPihony: Why Domain.scala when it is really TazInfo?

src/main/scala/census/db/creator/service/actors/DownloadActor.scala
```diff
@@ -0,0 +1,67 @@
...
+  actorSystem.scheduler.schedule(Duration.Zero, 1.second, self, CheckQueueMessage)
+
+  private val filesToDownload = mutable.Queue[String]()
+  private val downloadingsInProgress = new AtomicInteger(0)
+
+  override def receive: Receive = {
+    case DownloadMessage(url) =>
+      filesToDownload += url
+    case CheckQueueMessage =>
+      while (downloadingsInProgress.get() < 20 && filesToDownload.nonEmpty) {
```
JustinPihony: This is simple enough that it is probably ok here. For the future maybe think of spinning up 20 actors for fanning the download. Then you would't have to worry about the AtomicInteger. Anytime you start thinking about concurrency when using actors then you are likely in need of another hierarchy in your actor system. If this actor weren't closed by System.exit in its usage here then this scheduler would be unnecessary when it empties and should really shut itself  down. Again, with an actor approach you would only need a scheduler to check for dead downloads. But even that could probably be handled otherwise. 

TLDR - it's fine, and here are some future tips :)

src/main/scala/census/db/creator/service/actors/UnzipActor.scala
```diff
@@ -0,0 +1,45 @@
+package census.db.creator.service.actors
```
JustinPihony: I'd suggest simplifying this by using https://github.com/srikanth-lingala/zip4j

```
new ZipFile("filename.zip").extractAll("/destination_directory");
````

REASY:
Closing PR because we didn't go in this way, but a lot of code from this PR is used in https://github.com/LBNL-UCB-STI/beam/pull/2471
Also I marked it with `whishlist` label because it can be useful somewhere in the future.

PR #2376, opened 2020-02-25
Bl/#2373 actor that collects charging events

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2376)


JustinPihony:
Test!

src/main/scala/beam/agentsim/agents/vehicles/ChargingEventsAccumulator.scala
```diff
@@ -0,0 +1,57 @@
...
+      chargingEventsBuffer += e
+
+    case Finish =>
+      informExternalSystem(chargingEventsBuffer)
+      context.stop(self)
+
+  }
+
+  private def informExternalSystem(chargingEventsBuffer: ListBuffer[org.matsim.api.core.v01.events.Event]): Unit = {
+    //TODO implement this stub later
```
JustinPihony: Is this still a draft PR? Should this be implemented?

src/main/resources/beam-template.conf
```diff
@@ -218,8 +218,10 @@ beam.agentsim.agents.vehicles.sharedFleets = [
     }
   }
 ]
-beam.agentsim.agents.vehicles.collectChargingEvents=true
-beam.agentsim.agents.vehicles.collectChargingEventsIntervalInSeconds=300
+
+beam.agentsim.agents.vehicles.collectEvents=true
```
wrashid: default should be false

test/input/beamville/beam.conf
```diff
@@ -47,8 +47,10 @@ beam.agentsim.agents.vehicles.fuelTypesFilePath = ${beam.inputDirectory}"/beamFu
 beam.agentsim.agents.vehicles.vehicleTypesFilePath = ${beam.inputDirectory}"/vehicleTypes.csv"
 beam.agentsim.agents.vehicles.vehiclesFilePath = ${beam.inputDirectory}"/vehicles.csv"
 beam.agentsim.agents.vehicles.sharedFleets = []
-beam.agentsim.agents.vehicles.collectChargingEvents=true
-beam.agentsim.agents.vehicles.collectChargingEventsIntervalInSeconds=300
+
+beam.agentsim.agents.vehicles.collectEvents=true
+beam.agentsim.agents.vehicles.collectEventsIntervalInSeconds=300
+beam.agentsim.agents.vehicles.collectEventTypes="charging"
```
wrashid: where is this used in the code?

src/main/scala/beam/sim/LoggingEventsManager.scala
```diff
@@ -34,18 +34,18 @@ class LoggingEventsManager @Inject()(config: Config) extends EventsManager with
...
   }
 
   override def processEvent(event: Event): Unit = {
-    chargingEventsAccumulator match {
+    eventsAccumulator match {
       case Some(accumulator) =>
         event match {
           case e @ (_: ChargingPlugInEvent | _: ChargingPlugOutEvent | _: RefuelSessionEvent) =>
-            accumulator ! e
+            accumulator ! EventsAccumulator.ProcessChargingEvents(e)
```
wrashid: need to make this generic

src/main/scala/beam/agentsim/agents/vehicles/ChargingEventsAccumulator.scala
bhavyalatha26: No this stub will have no implementation for now. Ya, the PR is still a draft one

src/main/resources/beam-template.conf
bhavyalatha26: ok

wrashid: use beam.agentsim.collectEvents

src/main/resources/beam-template.conf
```diff
@@ -218,8 +218,10 @@ beam.agentsim.agents.vehicles.sharedFleets = [
...
+beam.agentsim.agents.vehicles.collectEventsIntervalInSeconds=300
+beam.agentsim.agents.vehicles.collectEventTypes="charging"
```
wrashid: let us remove collectEventTypes for now

bhavyalatha26:
Test!

src/main/resources/beam-template.conf
wrashid: remove this line

haitamlaarabi:
@wrashid @bhavyalatha26 
I don't see the reason of collecting events so to be transmitted as messages to an Actor. Why wouldn't be the eventsAcumulator both an Actor an EventHandler itself, without need to transmit messages ? 
+ bonus point: better modularity

wrashid:
> @wrashid @bhavyalatha26
> I don't see the reason of collecting events so to be transmitted as messages to an Actor. Why wouldn't be the eventsAcumulator both an Actor an EventHandler itself, without need to transmit messages ?
> 
> * bonus point: better modularity

EventHandlers are running single threaded and some are writing outputs. We do not want to slow down our main simulation, by making it dependent on them (e.g. see that the agentsim scheduler is awaiting trigger release). For the same reason we are sending the events to the accumulator actor in the first line of processEvent to avoid any overheads which might be cause by event handlers.

PR #2386, opened 2020-03-04
Add configuration of ride hail matching algorithm for PoolingAlonsoMora

SHaaD94:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2386)


SHaaD94:
test!

SHaaD94:
test!

src/main/scala/beam/agentsim/agents/ridehail/allocation/PoolingAlonsoMora.scala
```diff
@@ -179,12 +179,30 @@ class PoolingAlonsoMora(val rideHailManager: RideHailManager)
...
-      val alg =
-        new VehicleCentricMatchingForRideHail(
-          spatialPoolCustomerReqs,
-          availVehicles,
-          rideHailManager.beamServices
-        )
+
+      val alg: RHMatchingAlgorithm =
+        rideHailManager.beamServices.beamConfig.beam.agentsim.agents.rideHail.allocationManager.alonsoMora.matchingAlgorithm match {
+          case algoName if algoName == null || algoName.isEmpty || algoName == "VehicleCentricMatchingForRideHail" =>
```
REASY: I don't think you need to check against `null` because it will never be null due to https://github.com/LBNL-UCB-STI/beam/pull/2386/files#diff-1fe0fcebc66b3b894e52f8d586f67b36R750

SHaaD94: indeed, removed null check

PR #2388, opened 2020-03-04
joint distribution

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2388)

### DIFF
--- a/src/main/scala/beam/utils/data/ctpp/JointDistribution.scala
+++ b/src/main/scala/beam/utils/data/ctpp/JointDistribution.scala
@@ -1,24 +1,18 @@
 package beam.utils.data.ctpp
 
-import java.util
 import java.util.{Map => JavaMap}
+import org.apache.commons.math3.util.{Pair => CPair}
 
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/utils/data/ctpp/JointDistribution.scala
```diff
@@ -0,0 +1,148 @@
...
+/**
+  *
+  * @param mappedArray
+  * @param rndGen Random number generator. If an instance of `JointDistribution` is shared across multiple threads,
+  *               make sure you use thread-safe instance of `RandomGenerator`, for example, `SynchronizedRandomGenerator`
+  *               or wrap it by your own implementation which uses ThreadLocal
+  * @param columnMapping
+  * @param scale
+  */
+class JointDistribution(
```
REASY: @rajnikantsh please, add some comments to the constructor explaining what are those input parameters

rajnikantsh: @REASY now we have added details on constructor inputs

PR #2399, opened 2020-03-05
bug-fix : EventsAccumulator does not auto trigger after every require…

bhavyalatha26:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2399)


bhavyalatha26:
can we merge this @haitamlaarabi ?

PR #2406, opened 2020-03-09
BugFix: Clear the state of BeamVehicle in the end of iteration

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2406)


src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -54,6 +54,7 @@ class BeamVehicle(
 
   val rand: Random = new Random(randomSeed)
 
+  @volatile
   var spaceTime: SpaceTime = _
```
REASY: `volatile` because there are places outside of `BeamVehicle` which set `spaceTime`

Issue #2415, opened 2020-03-17
missing beamLog.out for tests


wrashid:

carloscaldas:
Closed by https://github.com/LBNL-UCB-STI/beam/issues/2415

PR #2420, opened 2020-03-19
generate Integration tests log in the right output folder

carloscaldas:
Issue: https://github.com/LBNL-UCB-STI/beam/issues/2415

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2420)

### DIFF
--- a/src/main/java/beam/utils/LoggingUtil.java
+++ b/src/main/java/beam/utils/LoggingUtil.java
@@ -12,11 +12,15 @@ import java.io.IOException;
 import java.io.InputStream;
 
 public class LoggingUtil {
+
+    public static final String LOG_OUTPUT_DIRECTORY_KEY = "log-path";
+
     private static boolean keepConsoleAppenderOn = true;
 
     public static void initLogger(String outputDirectory, boolean keepConsoleAppenderOn) throws JoranException, IOException {
-        String logFileName = System.getProperty(ContextInitializer.CONFIG_FILE_PROPERTY);
-        if (logFileName != null) {
+        String logbackConfigFile = System.getProperty(ContextInitializer.CONFIG_FILE_PROPERTY);
+
+        if (logbackConfigFile != null) {
             LoggingUtil.keepConsoleAppenderOn = keepConsoleAppenderOn;
             // https://logback.qos.ch/faq.html#sharedConfiguration
             LoggerContext context = (LoggerContext) LoggerFactory.getILoggerFactory();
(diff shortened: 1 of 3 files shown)


src/main/java/beam/utils/LoggingUtil.java
```diff
@@ -30,24 +34,23 @@ public static void initLogger(String outputDirectory, boolean keepConsoleAppende
             }
             // ################################################
 
-            InputStream resourceAsStream = context.getClass().getClassLoader().getResourceAsStream(logFileName);
+            InputStream resourceAsStream = context.getClass().getClassLoader().getResourceAsStream(logbackConfigFile);
             if (resourceAsStream != null) {
                 try {
+                    context.reset();
```
REASY: Why did you move the these lines upper? According the documentation http://logback.qos.ch/faq.html#sharedConfiguration
```java
LoggerContext context = (LoggerContext) LoggerFactory.getILoggerFactory();
  JoranConfigurator jc = new JoranConfigurator();
  jc.setContext(context);
  context.reset(); // override default configuration
  // inject the name of the current application as "application-name"
  // property of the LoggerContext
  context.putProperty("application-name", NAME_OF_CURRENT_APPLICATION);
  jc.doConfigure("/path/to/the/above/configuration/file.xml"); 
```

src/test/resources/logback-test.xml
```diff
@@ -15,8 +15,18 @@
...
+    <appender name="FILE" class="ch.qos.logback.core.FileAppender">
+        <File>${log-path}/beamLog.out</File>
+        <layout class="ch.qos.logback.classic.PatternLayout">
+            <Pattern>
+                %d{HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n
+            </Pattern>
+        </layout>
+    </appender>
+
+    <root level="INFO">
```
wrashid: beamLog.out should be info, but console should be Error level

src/main/java/beam/utils/LoggingUtil.java
carloscaldas: just code organization. Does not makes difference since `jc.setContext(context);` only saves the reference
```Java
    public void setContext(Context context) {
        if (this.context == null) {
            this.context = context;
        } else if (this.context != context) {
            throw new IllegalStateException("Context has been already set");
        }
    }
```

src/test/resources/logback-test.xml
carloscaldas: done

PR #2456, opened 2020-03-25
Refactoring: 30 small improvements/fixes

carloscaldas:
# Changes applied per file:
Road:
- Transformed instance variables into class variables: 1
- Added suffix to avoid loose of precision: 1
- Assigned variable never used: 1
- Removed redundant initialization: 1

PhyssimCalcLinkSpeedStats:
- Transformed static variables into constants: 3
- Transformed mutable object into immutable: 1
- Extracted method: 1
- Transformed instance variable into final instance variables: 5
- Removed redundant comments: 8
- Removed premature declaration: 1
- Transformed Boxed type into primitive type: 1

VehicleMilesTraveledAnalysis:
- Transformed instance variables into class variables: 1
- Added suffix to avoid loose of precision: 1
- Transformed instance variable into final instance variables: 2
- Avoid NullPointerException: 1
- Removed redundant comments: 1

#Relevant changes:
## Bug fix:
Previous:
```Java
LinkedList<Double> gap = getGap();
if (gap == null) {
    gap = new LinkedList<>();
} else {
    gap.clear();
}
```
After:
```Java
if (gap == null) {
    gap = new LinkedList<>();
} else {
    gap.clear();
}
```
**Comments**: Assignment to the reference `gap` only updates the local variable that is not used

## Avoid NPE:
Before:
```Java
this.vehicleTypes = vehicleTypes;
```
After:
```Java
this.vehicleTypes = vehicleTypes == null
    ? Set$.MODULE$.empty()
    : vehicleTypes;
```
**Comments**: The usage can lead to NPE. Other possible improvement would be to use Java Set instead of Scala Set since is a java class and no special feature from Scala collection is needed

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2456)

### DIFF
--- a/src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedStats.java
+++ b/src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedStats.java
@@ -57,9 +57,9 @@ public class PhyssimCalcLinkSpeedStats {
 
     private int getNumOfBinsFromConfig(BeamConfig beamConfig) {
         double endTime = Time.parseTime(beamConfig.matsim().modules().qsim().endTime());
-        Double noOfTimeBins = endTime / this.beamConfig.beam().physsim().linkStatsBinSize();
-        noOfTimeBins = Math.floor(noOfTimeBins);
-        return noOfTimeBins.intValue() + 1;
+        Double numOfTimeBins = endTime / this.beamConfig.beam().physsim().linkStatsBinSize();
+        numOfTimeBins = Math.floor(numOfTimeBins);
+        return numOfTimeBins.intValue() + 1;
     }
 
     public void notifyIterationEnds(int iteration, TravelTimeCalculator travelTimeCalculator) {
--- a/src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedStats.java
+++ b/src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedStats.java
@@ -33,16 +33,17 @@ import java.util.stream.IntStream;
  */
 public class PhyssimCalcLinkSpeedStats {
 
+    static final String OUTPUT_FILE_NAME = "physsimLinkAverageSpeedPercentage";
     private static final int BIN_SIZE = 3600;
     private static final List<Color> COLORS = Collections.unmodifiableList(
             Arrays.asList(Color.GREEN, Color.BLUE)
     );
     private static final int NUM_OF_BIN_TEST_MODE = 24;
+
     private final int numOfBins;
     private final BeamConfig beamConfig;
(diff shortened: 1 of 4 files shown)


src/main/java/beam/physsim/jdeqsim/cacc/sim/Road.java
```diff
@@ -172,7 +172,7 @@ private double  getNextAvailableTimeForLeavingStreet(double simTime){
     public void procsessFilledRoad(org.matsim.core.mobsim.jdeqsim.Vehicle vehicle, double simTime) {
         LinkedList<Double> gap = getGap();
         if (gap == null) {
-            gap = new LinkedList<>();
+            setGap(new LinkedList<>());
```
REASY: `setGap` will set the class field, but here was no such intention. I guess it is better not to do it

carloscaldas: The variable gap was not used after assignment. So either was redundant assignment or the intention was to set the field.

src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedStats.java
```diff
@@ -32,53 +33,48 @@
...
-            noOfBins = noOfTimeBins.intValue() + 1;
-        }
+        numOfBins = isTestMode()
+                ? NUM_OF_BIN_TEST_MODE
+                : getNumOfBinsFromConfig(beamConfig);
+    }
+
+    private int getNumOfBinsFromConfig(BeamConfig beamConfig) {
+        double endTime = Time.parseTime(beamConfig.matsim().modules().qsim().endTime());
+        Double noOfTimeBins = endTime / this.beamConfig.beam().physsim().linkStatsBinSize();
```
JustinPihony: If changing to num, then let's be uniform and change it here also

src/main/java/beam/physsim/jdeqsim/cacc/sim/Road.java
JustinPihony: I agree with Carlos - that seems to be a bug to me. Since it's a difference of null versus empty it likely didn't make a difference - if they are treated the same.

rajnikantsh: Yes, Gap is redundent. We can remove it. Only intention was if we use gap in method for further processing in future then we find not null value of gap

rajnikantsh: Also keep this gap.clear() with null check. because we are clearing gap value in this method

PR #2459, opened 2020-03-27
Jp/#2447 spot instances

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2459)

### DIFF
--- a/aws/build.gradle
+++ b/aws/build.gradle
@@ -93,8 +93,10 @@ task deploy(type: AWSLambdaInvokeTask) {
   "shutdown_wait": "${ext.getParameterValue('shutdownWait')}",
   "shutdown_behaviour": "${ext.getParameterValue('shutdownBehaviour')}",
   "command": "deploy",
-  "run_grafana" : "${ext.getParameterValue('runGrafana') ?: false}"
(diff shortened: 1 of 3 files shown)


JustinPihony:
@nikolayilyin What do you mean? Are you talking about the fact that spot instances can be shut down on you due to their nature? I'm not sure where the times you are referring are coming from 1.5-15 min?

nikolayilyin:
@JustinPihony, I mean that there are few delays during execution of lambda function if one requested a spot instance, and spot instance can not be ready right after it was requested.
So, i think, people may run gradle command with request of spot instance end expect that instance appears ready shortly, but, they may see, that instance is not ready. 
In that case people may think that there may be some error in parameters or something like that, or they may try to run gradle command once more, etc.

nikolayilyin:
those delays:

https://github.com/LBNL-UCB-STI/beam/blob/develop/aws/src/main/python/beam_lambda/lambda_function.py#L249

https://github.com/LBNL-UCB-STI/beam/blob/develop/aws/src/main/python/beam_lambda/lambda_function.py#L258

https://github.com/LBNL-UCB-STI/beam/blob/develop/aws/src/main/python/beam_lambda/lambda_function.py#L292

Issue #2457, opened 2020-03-26
create template for integrating python script call into beam


wrashid:
-> call script
-> get reply string
-> if anything goes wrong with script, continue after fixed timeout.

PR #2468, opened 2020-03-28
Added Scala wrapper for calling CarRideStats python script

carloscaldas:
Resolves: https://github.com/LBNL-UCB-STI/beam/issues/2457

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2468)

### DIFF
--- a/src/main/python/calibration/generate_rides_with_google_maps.py
+++ b/src/main/python/calibration/generate_rides_with_google_maps.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """Generate A CSV file containing Google Maps links and apply filters from the Beam files: {iteration}.personal.CarRideStats.csv.gz
 Example of usage:
 python generate_rides_with_google_maps.py \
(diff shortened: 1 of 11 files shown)


src/main/scala/beam/analysis/AnalysisProcessor.scala
```diff
@@ -53,39 +54,14 @@ object AnalysisProcessor extends ExponentialLazyLogging {
         }
       )
       logger.info(s"Running python script: $scriptPath with args $args")
-      PythonProcess(Some((Seq("python", scriptPath) ++ args).mkString(" ").run(processLogger)))
+      val str = (Seq("python3", scriptPath) ++ args).mkString(" ")
```
JustinPihony: This should be `processName`, right?

carloscaldas: right. thx

src/main/scala/beam/analysis/carridestats/CarRideStatsAreaBoundBox.scala
```diff
@@ -0,0 +1,12 @@
+package beam.analysis.carridestats
+
+case class CarRideStatsPoint(x: Double, y: Double) {
+  override def toString: String = s"($x,$y)"
+}
+
+case class CarRideStatsFilterAreaBoundBox(topLeft: CarRideStatsPoint, bottomRight: CarRideStatsPoint)
```
REASY: I think you don't need all of these. There is already a usage of so called `Envelope` type from jts. The usage can be found, for example, in ZonalParkingManager: https://github.com/LBNL-UCB-STI/beam/blob/0dd3a931369a53f8b1930daf4213a9592783042e/src/main/scala/beam/agentsim/infrastructure/ZonalParkingManager.scala#L254

src/main/python/calibration/generate_rides_with_google_maps.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
REASY: How can we be sure that use has Python3 and all the needed dependencies? Should we have `requirement.txt` for `pip` and make sure users install it before running Beam?

PR #2473, opened 2020-03-30
Rw/#2472 scale network speed

wrashid:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2473)


wrashid:
Test!

src/main/java/beam/router/r5/R5MnetBuilder.java
```diff
@@ -37,18 +37,19 @@
     private final Map<Coord, Id<Node>> coordinateNodes = new HashMap<>();
 
     private int matsimNetworkNodeId = 0;
+    private BeamConfig.Beam beamConfig;
```
REASY: nitpick: can be marked as `final`

Issue #2486, opened 2020-04-02
Add docker dependencies into updateDependencies lambda 


nikolayilyin:
Docker and docker-compose need to be installed in AMI, but there was a problem with the latest version of docker-compose.
So, it should be of specific version in order to work.

PR #2491, opened 2020-04-03
Added docker and docker-compose to AMI build script

nikolayilyin:
for #2486 

added `docker-ce`
added `docker-compose` of a specific version, because, there is a bag in the modern version of docker-compose.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2491)

### DIFF
--- a/aws/src/main/python/updateDependencies/lambda_function.py
+++ b/aws/src/main/python/updateDependencies/lambda_function.py
@@ -52,6 +52,10 @@ runcmd:
   - sudo apt-get install curl
   - curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
   - sudo apt-get install nodejs -y
+  - sudo apt-get install docker-ce docker-ce-cli containerd.io -y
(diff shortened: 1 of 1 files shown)


aws/src/main/python/updateDependencies/lambda_function.py
```diff
@@ -52,6 +52,10 @@
   - sudo apt-get install curl
   - curl -sL https://deb.nodesource.com/setup_8.x | sudo -E bash -
   - sudo apt-get install nodejs -y
+  - sudo apt-get install docker-ce docker-ce-cli containerd.io -y
```
REASY, line 55: This list of things keeps growing. We already have a mechanism (lambda `updateDependencies`) to update our image. What if we change the way how we install all required libraries to be there, so it will be happening only one time when w update our AMI instead of running it for every single run.

REASY: @nikolayilyin as discussed in person, I didn't pay an attention that it is already `updateDependencies`, so forget what I said :D

nikolayilyin:
test!

REASY:
test!

PR #2499, opened 2020-04-06
Generate parking from ctpp

zneedell:
Generates parking files based on regression from San Francisco data. Ex:

> taz,parkingType,pricingModel,chargingType,numStalls,feeInCents,reservedFor
00000123,Public,Block,None,50,0.0,Any
00000123,Workplace,Block,None,183,0.0,Any
00001536,Public,Block,None,30,0.0,Any
00001536,Workplace,Block,None,121,0.0,Any
00000537,Public,Block,None,5,0.0,Any
00000537,Workplace,Block,None,21,0.0,Any
00000236,Public,Block,None,61,0.0,Any
00000236,Workplace,Block,None,247,0.0,Any
00000585,Public,Block,None,14,0.0,Any
00000585,Workplace,Block,None,58,0.0,Any
00000117,Workplace,Block,None,9,0.0,Any
00000117,Residential,Block,None,7,0.0,Any
00001008,Public,Block,None,1,0.0,Any
00001008,Workplace,Block,None,20,0.0,Any
00001008,Residential,Block,None,13,0.0,Any
00001949,Public,Block,None,4,0.0,Any
00001949,Workplace,Block,None,17,0.0,Any

Did some maybe weird stuff to get `generate` to give the mapping from taz -> residents and workers

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2499)

### DIFF
--- a/src/main/scala/beam/utils/data/synthpop/models/Models.scala
+++ b/src/main/scala/beam/utils/data/synthpop/models/Models.scala
@@ -11,7 +11,7 @@ object Models {
   case class State(value: String) extends AnyVal
   case class County(value: String) extends AnyVal
 
-  class GenericGeoId()
+  abstract class GenericGeoId()
 
   case class PumaGeoId(state: State, puma: String) extends GenericGeoId {
     def asUniqueKey: String = s"${state.value}${puma}"
--- a/src/main/scala/beam/utils/data/synthpop/models/Models.scala
+++ b/src/main/scala/beam/utils/data/synthpop/models/Models.scala
@@ -11,7 +11,9 @@ object Models {
   case class State(value: String) extends AnyVal
   case class County(value: String) extends AnyVal
 
-  case class PumaGeoId(state: State, puma: String) {
+  class GenericGeoId()
+
+  case class PumaGeoId(state: State, puma: String) extends GenericGeoId {
     def asUniqueKey: String = s"${state.value}${puma}"
   }
 
@@ -24,7 +26,7 @@ object Models {
     }
   }
 
-  case class PowPumaGeoId(state: State, puma: String) {
+  case class PowPumaGeoId(state: State, puma: String) extends GenericGeoId {
     def asUniqueKey: String = s"${state.value}${puma}"
   }
 
@@ -37,11 +39,11 @@ object Models {
     }
   }
 
-  case class BlockGroupGeoId(state: State, county: County, tract: String, blockGroup: String) {
+  case class BlockGroupGeoId(state: State, county: County, tract: String, blockGroup: String) extends GenericGeoId {
     def asUniqueKey: String = s"${state.value}${county.value}${tract}${blockGroup}"
   }
 
-  case class TazGeoId(state: State, county: County, taz: String) {
+  case class TazGeoId(state: State, county: County, taz: String) extends GenericGeoId {
     def asUniqueKey: String = s"${state.value}${county.value}${taz}"
   }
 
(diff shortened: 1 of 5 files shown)


src/main/scala/beam/utils/data/synthpop/models/Models.scala
```diff
@@ -11,7 +11,9 @@ object Models {
   case class State(value: String) extends AnyVal
   case class County(value: String) extends AnyVal
 
-  case class PumaGeoId(state: State, puma: String) {
+  class GenericGeoId()
```
REASY, line 14: nitpick: let's make it as a trait or abstract class

src/main/scala/beam/utils/scenario/generic/writers/ParkingInfoWriter.scala
```diff
@@ -0,0 +1,80 @@
...
+    val onStreetParking =
+      ((numberPersons + numberWorkers) * (0.5559 - 0.1454 * (populationDensity + employmentDensity)))
+        .max(0)
+        .min(numberPersons + numberWorkers)
+    val offStreetCost = (-35.9285 + 5.2970 * employmentDensity + 15.7921 * populationDensity).max(0).min(2000)
+    val onStreetCost = (-22.6038 + 1.6954 * employmentDensity + 7.5305 * populationDensity).max(0).min(2000)
+    val residentialParking = if (populationDensity + employmentDensity < 50) {
+      (numberPersons / 2 * (50 - (populationDensity + employmentDensity)) / 50).toInt
+    } else {
+      0
```
REASY, line 72: Are all these constants model parameters? Is it possible to extract then to something meaningful or it is just coefficients?

src/main/scala/beam/utils/scenario/generic/writers/ParkingInfoWriter.scala
```diff
@@ -0,0 +1,80 @@
...
+  val destinationCoordSystem = CRS.decode("EPSG:3395", true)
+  val sourceCoordSystem = CRS.decode("EPSG:4326", true)
+  val mt: MathTransform = CRS.findMathTransform(sourceCoordSystem, destinationCoordSystem, true)
+
+  override def write(path: String, geoService: GeoService, tazCounts: Map[GenericGeoId, (Int, Int)]): Unit = {
+    val csvWriter = new CsvWriter(path, headers)
+    try {
+      tazCounts.foreach {
+        case (tazGeoId, (numberPersons, numberWorkers)) =>
+          val scaledTazArea = geoService.tazGeoIdToGeom.get(tazGeoId.asInstanceOf[TazGeoId]) match {
```
REASY, line 29: Now I see why you `GenericGeoId`. I left `PumaLevelScenarioGenerator.scala` just to have it as some reference that we did something on Puma level. If it starts to create so much headache, maybe we can stop having it as one of implementation of `ScenarioGenerator`

src/main/scala/beam/utils/scenario/generic/writers/ParkingInfoWriter.scala
```diff
@@ -0,0 +1,80 @@
...
+import scala.util.Try
+
+trait ParkingInfoWriter {
+  def write(path: String, geoService: GeoService, tazCounts: Map[GenericGeoId, (Int, Int)]): Unit
+}
+
+object CsvParkingInfoWriter extends ParkingInfoWriter {
+  private val headers: Array[String] =
+    Array("taz", "parkingType", "pricingModel", "chargingType", "numStalls", "feeInCents", "reservedFor")
+  val destinationCoordSystem = CRS.decode("EPSG:3395", true)
```
REASY, line 20: Shall we derive destination coord system from shape file?

src/main/scala/beam/utils/scenario/generic/writers/ParkingInfoWriter.scala
zneedell: Just regression parameters, no particularly meaningful physical interpretation unfortunately

src/main/scala/beam/utils/scenario/generic/writers/ParkingInfoWriter.scala
zneedell: Might be worth keeping in in case we move over to TAZs, but I haven't tried it on Puma IDs so there might be other ways that switching geometry breaks it

PR #2511, opened 2020-04-09
initial push for base review

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2511)


nikolayilyin:
Accordingly to the first TODO.
Is there a special way to handle lambdas which are died by timeout? Like a specific lambda function which AWS call if another lambda function died by timeout. That would be a simple and elegant solution.
Or maybe we should just count for 10 minutes (or any other time) and then terminate a spot instance request.

Also, just a reminder: it is a good idea to warn a Gradle function user that spot instance request may take a lot of time.

JustinPihony:
@nikolayilyin This is the last issue I have....there might be a way to set up a cloudwatch....but that seems overly complicated....compared to option 2 - terminate spot instance before the timeout itself and error out. Would you concur with that analysis?

JustinPihony:
test!

JustinPihony:
test!

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -232,44 +336,90 @@ def deploy_spot_instance(script, instance_type, region_prefix, shutdown_behaviou
...
+            print 'Waiting 30 seconds to let spot fleet cancel and then shutting down due to getting too close to lambda timeout'
+            time.sleep(30)
+            exit(123)
+        else:
+            time.sleep(30)
+            spot = ec2.describe_spot_fleet_requests(SpotFleetRequestIds = [spot_fleet_req_id]).get('SpotFleetRequestConfigs')[0]
+            status = spot.get('ActivityStatus')
+            state = spot.get('SpotFleetRequestState')
+    if state != 'active' or status != "fulfilled":
         exit(1)
```
nikolayilyin, line 397: Can we be sure at that point we do not have any dangling instances?
Like request was fulfilled, but the instance is still initializing or something else.
If something like that happens then will our AWS-idle-notification catch that?

nikolayilyin:
> @nikolayilyin This is the last issue I have....there might be a way to set up a cloudwatch....but that seems overly complicated....compared to option 2 - terminate spot instance before the timeout itself and error out. Would you concur with that analysis?

Yep, the second approach looks simpler and because of that - better.
In any case it is possible to work with cloudwatch later if necessary.

aws/src/main/python/beam_lambda/lambda_function.py
JustinPihony: This is more of a failsafe - but I can add the cancel call to be safe - I think if it gets here then it failed altogether, though. But the idle check would indeed be the ultimate failsafe

JustinPihony: Fixed via https://github.com/LBNL-UCB-STI/beam/pull/2511/commits/749e812dead43ef7c267bec0a2c78491e3439942

JustinPihony:
test!

PR #2517, opened 2020-04-10
An attempt to fix `akka.actor.InvalidActorNameException`

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2517)


JustinPihony:
Test!

JustinPihony:
For posterity - in case the build falls out of scope:

> PersonAgentSpec:
> A PersonAgent
> should know how to take a walk_transit trip when it's already in its plan *** FAILED *** (3 seconds, 15 milliseconds)
> java.lang.AssertionError: assertion failed: timeout (3 seconds) during expectMsgClass waiting for class beam.agentsim.agents.vehicles.TransitReservationRequest
> ...
> should also work when the first bus is late *** FAILED *** (3 seconds, 24 milliseconds)

...

> PersonAndTransitDriverSpec:
> A PersonAgent
> should know how to take a walk_transit trip when it's already in its plan *** FAILED *** (6 seconds, 75 milliseconds)


But there are plenty of actor errors that I do not believe were there...but maybe were?

REASY:
Closed in favor of https://github.com/LBNL-UCB-STI/beam/pull/2534. Instead of introducing unique name, @dimaopen just made sure that actor got terminated. https://github.com/akka/akka/issues/24165#issuecomment-360046988 is stated that this would work in case if Terminated received by parent actor and this is exactly our case.

PR #2522, opened 2020-04-10
also collect non-work activity data from NHTS

zneedell:
`activity-intercepts` and `activity-params` are the two input files taken by the secondary activity generation module, `activity-calibration-values` gives values for the average trip distance and average number of trips per capita to access different activity types.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2522)


zneedell:
Test!

PR #2532, opened 2020-04-15
Use now for ride hail reservation time

colinsheppard:
During trip planning (in ChoosesMode) we build a trip with legs that looks like e.g. WALK,RIDE_HAIL,WALK. When we do this we make the start time of the WALK leg be “now” and start time of the RIDE_HAIL be set to the pickup time, which therefore allows our mode choice alg to include wait time in the option.

So far so good. 

But then when we execute the trip we are taking the dummy WALK leg which takes 0 seconds and we are sending a reservation request to RHM with the start time of the RIDE_HAIL leg as the requested departure time. So this is a problem because the RHM takes requested departure time on faith and won’t dispatch a vehicle more quickly even if it were possible.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2532)


wrashid:
Test!

PR #2548, opened 2020-04-20
Performance improvement for RideHailFleetAnalysis

nikolayilyin:
Performance improvements mostly contain getting rid of creating intermediate collections and unnecessary type castings.

The whole algorithm contains a part that collects and group events and the part which processes collected events. The part which processes collected events is called very often. Once per group of events. For big simulation there are lots of groups, so, every taken action in the part which process group of events is expensive, especially creating new collections and unnecessary type casting.

So, mostly, the performance improvement was made by getting rid of zipWithIndex and getting events fields values directly from fields instead of 'event.getAttributes.get(<field name>)' method.

Also big performance improvement was achieved by parallel processing of events groups and with non-blocking adding results of such processing to the final array of values.

Also inactive tests for RideHailFleetAnalysis added. 
The reason why tests are inactive: the tests were created to check if code returns the same metrics from analyzing the same events file. So, in order to work the tests require a preselected events file and hardcoded values, generated based on events file, to check. Having preselected events file in git repo is not a good idea, so, the tests are inactive and only for manual usage. 

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2548)


nikolayilyin:
According to synthetic tests performance improvement is about 5 times.
According to production runs (production/sfbay/smart/smart-baseline.conf) performance degradation when collecting RH EV-CAV metrics is about 3% for iteration execution time.

nikolayilyin:
test!

nikolayilyin:
test!

PR #2550, opened 2020-04-20
wait time fix 

haitamlaarabi:
a recent update to matching algorithms allowed situations where a dropoff can happen before a pickup. This led to the agent to be stuck Waiting forever to be picked up, causing multiple tests to fail.
Longer duration of wait time increased the chance of these situations to happen as the solution search space increases.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2550)


haitamlaarabi:
Test!

haitamlaarabi:
Test!

PR #2595, opened 2020-05-01
Refactor - 20 small improvements

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2595)

### DIFF
--- a/src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
+++ b/src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
@@ -458,7 +458,7 @@ public class RealizedModeAnalysis extends BaseModeAnalysis {
                 if (benchMarkData.get(d) == null) {
                     builder.append(",0.0");
                 } else {
-                    builder.append("," + benchMarkData.get(d) * 100 / sum);
+                    builder.append(",").append(benchMarkData.get(d) * 100 / sum);
                 }
             }
             out.write(builder.toString());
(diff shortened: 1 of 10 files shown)


wrashid:
Test!

src/main/java/beam/analysis/plots/RealizedModeAnalysis.java
```diff
@@ -453,12 +453,12 @@ public void writeToReferenceCSV() {
...
 
             StringBuilder builder = new StringBuilder("benchmark");
-            double sum = benchMarkData.values().stream().reduce((x, y) -> x + y).orElse(0.0);
+            double sum = benchMarkData.values().stream().reduce(Double::sum).orElse(0.0);
             for (String d : cumulativeReferenceMode) {
                 if (benchMarkData.get(d) == null) {
                     builder.append(",0.0");
                 } else {
-                    builder.append("," + benchMarkData.get(d) * 100 / sum);
+                    builder.append(",").append(benchMarkData.get(d) * 100 / sum);
```
JustinPihony, line 461: Is this more performant? It seems to me that it would be better to concatenate inside the string?

carloscaldas: @JustinPihony it is more performant. Concatenating inside the string generates one String object that is appended afterwards and discarded. With the change there is no redundant object creation

PR #2600, opened 2020-05-01
keep track of vehicles not in beamVehicles due to Missing ChargingPlugOutEvents

### LINKED ISSUE #2533: Missing ChargingPlugOutEvents
Hello BEAM team,

I am using beam to simulate the ev charging infrastructure in Munich for my student thesis and encountered an issue with missing ChargingPlugOut Events. 
For the majority of charging events everything works fine. Vehicles get plugged in, plugged out and a RefuelSession Event occurs.

But if the agent tries to charge his vehicle at the last activity of the day AND this activity is a home activity, the vehicle never gets plugged out and its fuel level at the end of the iteration doesn’t get updated. 

I linked a code segment inside the PersonAgent.scala class which intends to handle charging events at the end of the iteration. As far as I could tell, the vehicle is still connected to the charging point when this loop is executed, but it is no longer assigned to this PersonAgent if the last activity is a ‘Home’ activity. If the last activity is of type ‘Other’ for example, the vehicle is still assigned to the agent at this point and everything works fine.

https://github.com/LBNL-UCB-STI/beam/blob/0bea90e5ae351c8940c854a93b1381352afaf66e/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L436-L449

The unassignment happens here

https://github.com/LBNL-UCB-STI/beam/blob/0bea90e5ae351c8940c854a93b1381352afaf66e/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L1002-L1006

Is this behaviour intended? If yes, is there an alternative way to finish every charging process at the end of an iteration?

Thanks in advance.

haitamlaarabi:
Closes #2533

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2600)


JustinPihony:
Should we write these out somewhere?

haitamlaarabi:
Test!

PR #2621, opened 2020-05-06
fix average speed vs freeflow speed graph

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2621)


rajnikantsh:
close via https://github.com/LBNL-UCB-STI/beam/pull/2625

PR #2623, opened 2020-05-06
Add percentToSimulate to beam-template.conf

### LINKED ISSUE #2553: Make Experiment_5_0 parameterized
Experiment_5_0 is hardcoded to have only 10 iterations of routing assignments, 10% of people loaded on each iteration:
https://github.com/LBNL-UCB-STI/beam/blob/b3716666f526afe70c54cf0adda643c0fbd2ad37/src/main/scala/beam/physsim/jdeqsim/RelaxationExperiment.scala#L424

Instead of that let's allow for the user to provide an array of the percentages of people to _simulate_ at that iteration. The number of iterations will be derived from the length of that array. 
This will allow getting rid of `Experiment_5_1` and using this approach instead.

Parameter name: `beam.physsim.relaxation.experiment5_0.percentToSimulate = []`

Once this parameter is implemented, it has to be passed to ApproxPhyssim: https://github.com/LBNL-UCB-STI/beam/blob/b3716666f526afe70c54cf0adda643c0fbd2ad37/src/main/scala/beam/physsim/jdeqsim/RelaxationExperiment.scala#L424

Notes:
- If the sum is not 100% than re-normalize assuming it is weights
- It is expected that the sum of percents should be 100
- The total number of iterations is equal to the length of the array

crixal:
Closes #2553 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2623)


crixal:
Test!

crixal:
optional annotation is used because default value for list type is not available. I create issue with this problem.

crixal:
Test!

JustinPihony:
Test!

PR #2652, opened 2020-05-14
Move jfree usage to GraphUtils, part1

crixal:
Move some jfree usage to GraphUtils.
Examples
1. Old version of `passengerPerTripRideHail.png` is
0 passengerPerTripRideHail

Fixed version is
0 passengerPerTripRideHail

2. Old version of `physsimFreeFlowSpeedDistribution.png` is
0 physsimFreeFlowSpeedDistribution
Fixed version is
0 physsimFreeFlowSpeedDistribution
Pngs looks same, bug in csv file.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2652)

### DIFF
--- a/src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedDistributionStats.java
+++ b/src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedDistributionStats.java
@@ -1,30 +1,22 @@
 package beam.analysis.physsim;
 
-import beam.analysis.plots.GraphsStatsAgentSimEventsListener;
-import beam.sim.OutputDataDescription;
+import beam.analysis.plots.GraphUtils;
 import beam.sim.config.BeamConfig;
-import beam.utils.OutputDataDescriptor;
-import org.jfree.chart.ChartFactory;
-import org.jfree.chart.ChartUtilities;
+import com.google.common.base.Suppliers;
 import org.jfree.chart.JFreeChart;
-import org.jfree.chart.plot.PlotOrientation;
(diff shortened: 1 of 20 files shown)


src/main/java/beam/analysis/physsim/PhyssimCalcLinkSpeedDistributionStats.java
```diff
@@ -104,25 +103,22 @@ private CategoryDataset generateLinkEfficienciesDataSet(Map<Double, Integer> gen
...
             String completeHeading = heading + ",numberOfLinks\n";
             bw.write(completeHeading);
-            double[] data = dataMatrix[0];
-            IntStream.range(0,data.length)
+            IntStream.range(0, data.length)
                     .forEach( i -> {
                         try {
                             bw.write(i + "," + data[i] + "\n");
                         } catch (IOException e) {
                             e.printStackTrace();
```
REASY, line 119: It was like this before, but could you, please, add logging in all the places where `e.printStackTrace` is used?, e.g use 
```java
private final Logger log = LoggerFactory.getLogger(PhyssimCalcLinkSpeedDistributionStats.class);
```
to log the exception so we will be able to find it in the logs.

crixal: Ok, I fixed it in part 2. I suggest merge it then I can create a new one PR.

JustinPihony:
Since this is Part 1, I assume this does NOT close 2618?

PR #2654, opened 2020-05-15
We need to have JDK installed in order to get javac (java compiler)

REASY:
Env: Ubuntu 18.04, only JRE is installed (openjdk-8-jre-headless). Got an error during compilation of beam:
```
FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':compileScala'.
> java.io.IOException: Cannot run program "/usr/lib/jvm/java-8-openjdk-amd64/bin/javac": error=2, No such file or directory
```

image

As we can see it needs `javac`, Java compiler. After installing JDK (`sudo apt-get install openjdk-8-jdk`) it worked:

image

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2654)

### DIFF
--- a/docs/users.rst
+++ b/docs/users.rst
@@ -13,9 +13,18 @@ System Requirements
 
 * At least 8GB RAM
 * Windows, Mac OSX, Linux
-* Java Runtime Environment or Java Development Kit 1.8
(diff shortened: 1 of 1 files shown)


REASY:
test!

PR #2661, opened 2020-05-18
Ak/#2540 java scala

### LINKED ISSUE #2540: Java code in Scala
Hello! I looked through the sources and found a lot of code pieces where we use plain Java approaches to get data from legacy or other steps. I have my doubts on that style because it could lead us to working with exceptions and other things which is not good for Scala with functional approaches. So my suggestion is to wrap the Java interfaces and classes with scala classes and provide a resulting types from Scala type system.

Here is a few examples:

1) Population.scala:74. Using null instead of Option
```
     if (scenario.getHouseholds.getHouseholdAttributes
            .getAttribute(household.getId.toString, "homecoordx") == null) {
        log.error(
          s"Cannot find homeCoordX for household ${household.getId} which will be interpreted at 0.0"
        )
      }
```

and then

```
val homeCoord = new Coord(
        scenario.getHouseholds.getHouseholdAttributes
          .getAttribute(household.getId.toString, "homecoordx")
          .asInstanceOf[Double]
```

I would suggest to wrap `scenario.getHouseholds.getHouseholdAttributes.getAttribute(household.getId.toString, "homecoordx")` with Option and work with this functor.

2) Throwing plain exception 

In PersonAgent.scala: 497 
```
assert(currentActivity(data).getLinkId != null)
```

In DrivesVehicle.scala: 228
```
 val currentLeg = data.passengerSchedule.schedule.keys.view
        .drop(data.currentLegPassengerScheduleIndex)
        .headOption
        .getOrElse(throw new RuntimeException("Current Leg is not available."))
```
instead of building logic arround Option context we just throwing an exception.

Do we just crashing actors on Exception and this is ok for us or it would be better to stop them gracefully without any kind of Exception? In this context Exception is just a warning from the logic that saying 'ok, something went wrong'. In the functional approach we have Either and Validated from  [cats validated](https://typelevel.org/cats/datatypes/validated.html) which could fits our needs better.

3) Force extract value from Option which could lead to getting an exception as well.

In BeamVehicle.scala: 86 we have
```
  def stall: Option[ParkingStall] = stallRWLock.read { stallInternal }
```

Meantime in DrivesVehicle.scala: 895
```
vehicle.stall.get
```

or

BeamVehicle.scala: 77
```
  private val driver: AtomicReference[Option[ActorRef]] = new AtomicReference(None)
  def getDriver: Option[ActorRef] = driver.get()
```

and then 

HouseHoldActor.scala:374
```
 cavAndSchedule._1.getDriver.get,
```

Again it would be better to go with Option context and not throwing an Exception.

Summary: If it's proper behavior on the project it is ok, but in my understanding (and by experience from projects I worked on) that could lead us to some problems with messy code in the future. Thanks!

YOxan:
Resolves #2540 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2661)


YOxan:
Test!

JustinPihony:
test!

JustinPihony:
Failure:

> 22:41:41 /home/ubuntu/workspace/beam-forced-prs@2/src/main/scala/beam/utils/csv/writers/PopulationCsvWriter.scala:60: not found: value readAge
> 22:41:41       val personAge = readAge(

PR #2670, opened 2020-05-21
new script to automatically generate nhts inputs

zneedell:
allows for a python script to be called with an argument determining which NHTS urban areas to include in producing work end time / duration distribution and activity end times and intercepts.

The first argument defines the urban areas and can take several formats:
String: (e.g. '12420') just gives the distribution for the urban area defined by the FIPS code in the string
List of strings: (e.g. '[12420,12060]') returns a single distribution averaged across all regions, weighted by sample size
Dict: (e.g.  '{12420:0.25,12060:0.25,-1:0.5}'), returns a single distribution weighted by the dictionary values. 

In all cases, a negative input defaults averaging across all CBSAs

The second argument is the path to the output.

Still needs a way of handling the NHTS input data. Right now it's just assumed to be local. Do we want to push it to lfs, or put it on s3 somewhere?

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2670)

### DIFF
--- a/.lfsconfig
+++ b/.lfsconfig
@@ -1,2 +1,1 @@
-[lfs]
-  url = http://52.15.173.229:8080/LBNL-UCB-STI/beam.git
+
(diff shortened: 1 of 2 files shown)


zneedell:
Still needs to be integrated with the ctpp pipeline, but now loads NHTS data from s3

PR #2674, opened 2020-05-22
write VM heap class histogram at iteration end and/or start

nikolayilyin:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2674)


src/main/scala/beam/utils/VMInfoCollector.scala
```diff
@@ -0,0 +1,57 @@
...
+  private val server: MBeanServer = ManagementFactory.getPlatformMBeanServer
+
+  def gcRun(): Unit = invoke("gcRun")
+
+  def gcClassHistogram: String = {
+    gcRun()
+    invoke("gcClassHistogram")
+  }
+
+  //  commands examples:
```
REASY, line 40: I'd suggest to use Scala type system to help with possible params. It can be done by using ADT, for example,
```scala
sealed trait JFRCommand {
 def asStr: String
}
object JFRCommand {
  case object GcRun extends JFRCommand  {
      def asStr: String = "gcRun"
  }
}
```

src/main/scala/beam/utils/VMInfoCollector.scala
```diff
@@ -0,0 +1,57 @@
...
+object VMInfoCollector {
+  private val DIAGNOSTIC_COMMAND_MBEAN_OBJECT_NAME = "com.sun.management:type=DiagnosticCommand"
+
+  def apply(): VMInfoCollector = {
+    try {
+      val objectName = new ObjectName(DIAGNOSTIC_COMMAND_MBEAN_OBJECT_NAME)
+      new VMInfoCollector(objectName)
+    } catch {
+      case _: MalformedObjectNameException =>
+        throw new RuntimeException(
```
REASY, line 23: Let's not throw away that exception, but instead, pass it to `RuntimeException` so there will be more information what went wrong.

src/main/scala/beam/utils/VMInfoCollector.scala
nikolayilyin: ok, looks cool, thanks!

src/main/scala/beam/utils/VMInfoCollector.scala
nikolayilyin: done

nikolayilyin:
test!

nikolayilyin:
test!

nikolayilyin:
test!

nikolayilyin:
test!

nikolayilyin:
test!

PR #2684, opened 2020-05-27
CsvSkimReader: Minor improvements

REASY:
- Reduce memory pressure by using `scala.collection.Map`
- Stop parsing in `initHeaders` and `readAggregatedSkims` in any case

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2684)

### DIFF
--- a/src/main/scala/beam/router/skim/CsvSkimReader.scala
+++ b/src/main/scala/beam/router/skim/CsvSkimReader.scala
@@ -3,13 +3,14 @@ package beam.router.skim
 import java.io.File
 import java.util
 
-import com.typesafe.scalalogging.{Logger}
+import com.typesafe.scalalogging.Logger
 import com.univocity.parsers.common.record.Record
 import com.univocity.parsers.csv.{CsvParser, CsvParserSettings}
 import org.matsim.core.utils.io.IOUtils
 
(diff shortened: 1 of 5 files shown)


wrashid:
Test!

src/main/scala/beam/router/skim/CsvSkimReader.scala
```diff
@@ -21,18 +22,18 @@ import scala.util.control.NonFatal
...
+  fromCsv: scala.collection.Map[String, String] => (AbstractSkimmerKey, AbstractSkimmerInternal),
   logger: Logger
 ) {
 
-  val header: Array[String] = initHeader
+  val header: Array[String] = initHeader()
 
   def readAggregatedSkims: immutable.Map[AbstractSkimmerKey, AbstractSkimmerInternal] = {
     var res = Map.empty[AbstractSkimmerKey, AbstractSkimmerInternal]
     val csvParser: CsvParser = getCsvParser
```
crixal, line 33: csvReader is created twice. May be move code from `initHeader()` here?

REASY: @crixal as per discussion, let's repair it in the separate PR

PR #2718, opened 2020-06-05
Art/#2701 skims in parallel 4ci

REASY:
High-level simplified diagram of the solution:
image

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2718)


nikolayilyin:
test!

alex-vv:
Test!

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -534,7 +581,11 @@ class BeamSim @Inject()(
     currentIteration == firstIteration
   }
 
+  import scala.concurrent.duration._
```
nikolayilyin, line 584: I think it would be better to put import in the beginning of the file

nikolayilyin (changes requested):
I am one of the authors, so, mostly it looks good to me

src/main/scala/beam/router/skim/ActivitySimSkimmer.scala
```diff
@@ -0,0 +1,273 @@
...
+    val EAHours = (3 until 6).toList
+    val AMHours = (6 until 10).toList
+    val MDHours = (10 until 15).toList
+    val PMHours = (15 until 19).toList
+    val EVHours = (0 until 3).toList ++ (19 until 24).toList
+
+    val pathTypes = ActivitySimPathType.allPathTypes
+    var writer: BufferedWriter = null
+    try {
+      writer = org.matsim.core.utils.io.IOUtils.getBufferedWriter(filePath)
```
dimaopen, line 93: Why don't we use CsvWriter from BEAM here?

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -740,4 +785,49 @@ class BeamSim @Inject()(
...
+          finalSkimmer.writeToDisk(
+            new IterationEndsEvent(beamServices.matsimServices, beamServices.matsimServices.getIterationNumber)
+          )
+        } catch {
+          case NonFatal(ex) =>
+            logger.error(
+              s"Can't get the result from background skims creator or write the result to the disk: ${ex.getMessage}",
+              ex
+            )
+            None
```
dimaopen, line 827: We could remove this None

dimaopen (approved):
LGTM

src/main/scala/beam/router/skim/ActivitySimSkimmer.scala
alex-vv: I suppose this is because existing AbstractSkimmer and ODSkimmer use this way for writing CSV. I also checked there are a few other places where it is used, probably we can create a ticket to migrate them all to use CsvWriter?

src/main/scala/beam/sim/BeamSim.scala
alex-vv: done

src/main/scala/beam/router/skim/ActivitySimSkimmer.scala
dimaopen: I agree. Let's do this within another ticket.

nikolayilyin:
test!

nikolayilyin (approved):
still lgtm

PR #2732, opened 2020-06-09
Remove timestamp from config file

rajnikantsh:


build.gradle
```diff
@@ -393,12 +393,26 @@ task generateConfig {
                     "--scala",
                     "--pn", "beam.sim.config",
                     "--cn", "BeamConfig",
```
REASY, line 395: nitpick: Why don't you provide a temporary class name and then, later on, replace it by the fixed on in `copyConfig` and then delete temp one in clean-up?

rajnikantsh: If we provide remporary class name while generating class from tscfg library then its generate inside classes in that file with temporary class name. So later if we change filename in `copyConfig` then its only change filename but generated inside classes still remain with same name. 

some snippet from code if we generate a temporary class name 

 `case class BeamConfigTmp(
  beam: BeamConfigTmp.Beam,
  matsim: BeamConfigTmp.Matsim
)

object BeamConfigTmp {
`

REASY: I see, then it is the way to go, thanks!

rajnikantsh:
test!

PR #2745, opened 2020-06-11
BugFix: CCH on Windows machine fails

REASY:
Use Unix path for all paths which are passed to Docker container

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2745)


REASY:
test!

PR #2760, opened 2020-06-17
Fix agent sim for plans with one activity

crixal:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2760)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
+++ b/src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
@@ -14,11 +14,7 @@ import beam.agentsim.agents.modalbehaviors.ChoosesMode.{CavTripLegsRequest, CavT
 import beam.agentsim.agents.modalbehaviors.DrivesVehicle.VehicleOrToken
 import beam.agentsim.agents.modalbehaviors.{ChoosesMode, ModeChoiceCalculator}
 import beam.agentsim.agents.planning.BeamPlan
-import beam.agentsim.agents.ridehail.RideHailAgent.{
-  ModifyPassengerSchedule,
-  ModifyPassengerScheduleAck,
-  ModifyPassengerScheduleAcks
-}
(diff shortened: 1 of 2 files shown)


src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -268,6 +265,16 @@ object HouseholdActor {
         household.members.foreach { person =>
           val attributes = person.getCustomAttributes.get("beam-attributes").asInstanceOf[AttributesOfIndividual]
           val modeChoiceCalculator = modeChoiceCalculatorFactory(attributes)
+          val selectedPlan = person.getSelectedPlan
+          // Set zero endTime for plans with one activity. In other case agent sim will be started
+          // before all InitializeTrigger's are completed
+          if (selectedPlan.getPlanElements.size() == 1) {
+            selectedPlan.getPlanElements.get(0) match {
+              case elem: Activity => if(Time.isUndefinedTime(elem.getEndTime)) elem.setEndTime(0.0)
```
zneedell, line 273: Conceptually, should the end time of the activity be the end of the simulation? If that causes things not to work this seems like a fine solution to me.

zneedell: Got it, thanks--so when endTime is 0.0 the activityEndTrigger is processed in the initialization but then nothing else is scheduled for that agent? If that doesn't cause any problems later in the simulation that is fine with me.

REASY:
test!

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
crixal: Yes, it's work fine with`0.0` endTime.
I also found big with plan attributes. You can find it in changes.

zneedell: Good catch, thanks!

PR #2770, opened 2020-06-21
Fixed nullPointer exception and removed try-catch

carloscaldas:


carloscaldas:
Building is passing
image

PR #2801, opened 2020-07-06
more from NHTS

zneedell:
To use for NYC, look for the CBSA code 35620 (San Francisco bay area is 41860)


JustinPihony:
Is this only for testing? The values changed seem to be specific for this scenario - maybe stick to this branch only?

JustinPihony:
@zneedell Ping on my question?

PR #2829, opened 2020-07-12
Complete beam leg ordering

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2829)


REASY:
test!

src/main/scala/beam/router/model/BeamLeg.scala
```diff
@@ -30,7 +30,8 @@ case class BeamLeg(startTime: Int, mode: BeamMode, duration: Int, travelPath: Be
   }
 
   def scaleToNewDuration(newDuration: Int): BeamLeg = {
-    val newTravelPath = this.travelPath.scaleTravelTimes(newDuration.toDouble / this.duration.toDouble)
+    val ratio = if (newDuration == 0 || this.duration == 0) 0 else newDuration.toDouble / this.duration.toDouble
```
REASY, line 33: Otherwise if we get `Double.NaN`
image

PR #2847, opened 2020-07-16
remove activities/person for certain industries

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2847)


rajnikantsh:
test!

JustinPihony:
Technically it looks good, but it is failing on multiple tests for breaking the expected flow

src/main/resources/beam-template.conf
```diff
@@ -268,6 +268,8 @@ beam.agentsim.agents.vehicles.sharedFleets = [
 beam.agentsim.collectEvents=false
 
 beam.agentsim.agents.population.useVehicleSampling = "boolean | false"
+beam.agentsim.agents.population.industryRemovalProbabilty.inputFilePath = ""
+beam.agentsim.agents.population.industryRemovalProbabilty.removeAgent = "boolean | false"
```
REASY, line 272: Please, add some description. I can see that if `removeAgent ` is false, then it will remove all work plans.

REASY: As per discussion, let's add another parameter to control this functionality: `beam.agentsim.agents.population.industryRemovalProbabilty.enabled  = "boolean | false"`

src/main/java/beam/utils/WeightedRandomBag.java
```diff
@@ -0,0 +1,35 @@
+package beam.utils;
```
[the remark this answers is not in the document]
rajnikantsh: Removed WeightedRandomBag and use EnumeratedDistribution class

src/main/resources/beam-template.conf
rajnikantsh: add comment on usability of removeAgent and add control param for removeAgent

rajnikantsh:
test!

src/main/scala/beam/sim/population/PopulationScaling.scala
```diff
@@ -195,6 +199,76 @@ trait PopulationScaling extends LazyLogging {
...
+    try {
+      val industrialProbability = iter.map(value => value("industry") -> value("removal_probability").toDouble).toMap
+      val persons = scenario.getPopulation.getPersons.values().asScala
+
+      //Build persons probability
+      val personDistributionPair = persons
+        .map(person => {
+          val industry = person.getAttributes.getAttribute("industry").toString
+          val probability = industrialProbability.getOrElse(industry, 0.0)
+          new PPair[Person, java.lang.Double](person, probability)
```
REASY, line 229: [EnumeratedDistribution](https://commons.apache.org/proper/commons-math/javadocs/api-3.3/org/apache/commons/math3/distribution/EnumeratedDistribution.html#EnumeratedDistribution(java.util.List)) will throw MathArithmeticException if all of the probabilities are 0. Please, let's make sure that it won't happen.

src/main/scala/beam/sim/population/PopulationScaling.scala
```diff
@@ -195,6 +199,76 @@ trait PopulationScaling extends LazyLogging {
...
+        })
+        .toVector
+
+      val enumeratedDistribution = new EnumeratedDistribution[Person](rng, personDistributionPair.asJava)
+
+      //Calculate number of persons to be removed from population
+      val removalItemSize = persons
+        .groupBy(_.getAttributes.getAttribute("industry").toString)
```
REASY, line 237: If there is no attribute `industry`, there will be NullPointerException. Let's handle that case

src/main/scala/beam/sim/population/PopulationScaling.scala
```diff
@@ -195,6 +199,76 @@ trait PopulationScaling extends LazyLogging {
...
+      toClose.close()
+    }
+  }
+
+  def removeWorkPlan(scenario: MutableScenario): Unit = {
+    scenario.getPopulation.getPersons.values().asScala.foreach { person: Person =>
+      val originalPlan = person.getSelectedPlan
+      val newPlan = PopulationUtils.createPlan(originalPlan.getPerson)
+      val planElements = originalPlan.getPlanElements
+      planElements.removeIf(isWorkActivity)
```
REASY, line 255: What about the legs which are left? Shouldn't they be removed too?

src/main/scala/beam/utils/scenario/BeamScenarioLoader.scala
```diff
@@ -150,7 +150,7 @@ class BeamScenarioLoader(
       personAttributes.putAttribute(personId, "sex", sexChar)
       person.getAttributes.putAttribute("sex", sexChar)
       person.getAttributes.putAttribute("age", personInfo.age)
-
+      person.getAttributes.putAttribute("industry", personInfo.industry)
```
REASY, line 153: `personInfo.industry` is an option, please use `getOrElse` otherwise the saved value will be `Some("IndustryType")`

src/main/scala/beam/sim/population/PopulationScaling.scala
rajnikantsh: make sure atleast 1 probability is greater then 0

src/main/scala/beam/sim/population/PopulationScaling.scala
rajnikantsh: handle NullPointerException

src/main/scala/beam/sim/population/PopulationScaling.scala
rajnikantsh: keep only first activity as per discussion

src/main/scala/beam/utils/scenario/BeamScenarioLoader.scala
rajnikantsh: applied changes as per suggestion

PR #2853, opened 2020-07-19
Austin: The latest scenario

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2853)


REASY:
test!

REASY:
test!

REASY:
test!

REASY:
test!

src/main/scala/beam/utils/scenario/urbansim/censusblock/UrbansimReaderV2.scala
```diff
@@ -21,15 +21,15 @@ class UrbansimReaderV2(
   private val logger = LoggerFactory.getLogger(getClass)
 
   private val inputHouseHoldMap: Map[String, InputHousehold] = {
-    logger.debug("Start reading of households info...")
+    logger.info("Start reading of households info...")
```
JustinPihony, line 24: Do you want this for the final merge - or was it only for easier debugging?

src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/InputPlanElement.scala
```diff
@@ -16,8 +16,8 @@ case class InputPlanElement(
 
 object InputPlanElement extends EntityTransformer[InputPlanElement] {
   override def transform(m: util.Map[String, String]): InputPlanElement = {
-    val personId = getIfNotNull(m, "person_id")
-    val planElementIndex = getIfNotNull(m, "PlanElementIndex").toDouble.toInt
+    val personId = getIfNotNull(m, "person_id").split("\\.").apply(0)
+    val planElementIndex = getIfNotNull(m, "PlanElementIndex").toInt
```
JustinPihony, line 20: Are you sure the `toDouble` is not needed? This will throw on values such as "3.0"

src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/InputPlanElement.scala
```diff
@@ -16,8 +16,8 @@ case class InputPlanElement(
 
 object InputPlanElement extends EntityTransformer[InputPlanElement] {
   override def transform(m: util.Map[String, String]): InputPlanElement = {
-    val personId = getIfNotNull(m, "person_id")
-    val planElementIndex = getIfNotNull(m, "PlanElementIndex").toDouble.toInt
+    val personId = getIfNotNull(m, "person_id").split("\\.").apply(0)
```
JustinPihony, line 19: What about the case where it is less than 0 without a leading 0?

REASY: Here is how input `plan.csv` looks like:
image
So person_id is represented as float, but agree, splitting is not the best way to deal

src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/InputPlanElement.scala
REASY: The input is the following:
image

As you can see `PlanElementIndex` is an integer

src/main/scala/beam/utils/scenario/urbansim/censusblock/UrbansimReaderV2.scala
REASY: I changed it in order to have better visibility because of some faced issue. Would prefer to leave it on info level in case something will go wrong.

REASY:
@nikolayilyin 
as per discussion we might need to extract all Austin data out from develop branch, but for now I'm merging it to `develop` to make sure nothing is left for Austin

PR #2860, opened 2020-07-21
Fix issue in secondary activities

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2860)


src/main/scala/beam/replanning/AddSupplementaryTrips.scala
```diff
@@ -57,14 +57,17 @@ class AddSupplementaryTrips @Inject()(beamConfig: BeamConfig) extends PlansStrat
         listOfAct.lastOption match {
           case Some(lastAct) =>
             if (lastAct.getType == currentAct.getType) {
-              listOfAct.last.setEndTime(currentAct.getEndTime)
-              listOfAct
+              val lastActivity = PopulationUtils.createActivity(listOfAct.last)
```
zneedell, line 60: Should this just be `PopulationUtils.createActivity(lastAct)` or am I missing something

rajnikantsh: this will also work. applied changes as per suggestion

PR #2861, opened 2020-07-22
change name to mitigate confusion of plugin

### LINKED ISSUE #2781: is there a reason for having two tasks checkScalafmt and checkScalaFmt when we look at the IntelliJ gradle plugin?
image

JustinPihony:
Resolves #2781 

This will NOT pass CI because we are removing the old name. I thought about aliasing it, but then you have the same problem as brought up in #2781 - so I think it'll be better to temporarily break branches until they merge this change. It's such a minor one anyway that it could be done manually if needed.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2861)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -556,7 +556,7 @@ tasks.withType(ScalaCompile) {
 }
 
 task fmt(dependsOn: scalafmtAll)
-task checkScalaFmt() {
+task verifyScalaFmtHasBeenRun() {
     doLast {
         try {
             def workingDir = new File("${project.projectDir}")


JustinPihony:
Hrmmm, I take it back - it seems if the task is not there then it does not throw an exception - just ignores it. I'm unsure how I feel about that, but if nobody sees it as a problem I won't dig into that new item.

JustinPihony:
Dagnabbit - this from an older branch:

> org.gradle.execution.TaskSelectionException: Task 'verifyScalaFmtHasBeenRun' not found in root project 'beam

@wrashid Are we willing to accept failures in CI and have users change their branches as needed? I think the original didn't fail because it isn't case sensitive and picks up the built in one.

PR #2865, opened 2020-07-24
Beautify population scaler

SHaaD94:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2865)


SHaaD94:
Test!

src/main/scala/beam/sim/population/PopulationScaling.scala
```diff
@@ -38,90 +39,72 @@ class PopulationScaling extends LazyLogging {
...
+        .map {
+          case (houseHoldId, houseHold) =>
+            // proceed only if the required population is not yet reached
+            // get the members in the current house hold and duplicate them with new ids
+            val members =
+              houseHold.getMemberIds.asScala
+                .flatMap(m => Option(scenario.getPopulation.getPersons.get(m)))
                 // proceed only if the required population is not yet reached
-                if (populationCounter <= totalPopulationRequired) {
+                .take((totalPopulationRequired - populationCounter + 1).toInt)
```
JustinPihony, line 55: This isn't the same logic - it will result in 1 more than the original code. For example, the original code:

```
totalPopReq = 4
popCtr = 1 //Starting size
```

Now, when it first hits it increments (2) and passes the if (2<=4), so run = 1. Then on the next run it increments (3) and passes the if (3<=4), so run=2. Then on the next run it increments (4) and passes the if (4<=4), so run=3. Then on the next run it increments (5) and fails the if (5<=4), so run stops at 3.

So, 3 runs with the above starting point. 

In the new code we would expect the take to be 3, but it is 4 (4-1+1=4). 

Thus the +1 should be dropped, unless you say the original code was the bug?

SHaaD94:
Test!

src/main/scala/beam/sim/population/PopulationScaling.scala
SHaaD94: Indeed, thank you, Justin. 
Missed this stuff last night :D
Fixed, also compared the results with the old ones again, also looks like old code was generating some households without people.

SHaaD94:
Test!

SHaaD94:
Test!

PR #2867, opened 2020-07-28
Fix of activity warnings

dimaopen:
Fixed "didn't get nextActivity" and "performing Activity at end of simulation" warnings by completing the day in case of no next activity.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2867)


dimaopen:
This appears that the logs are not fully relevant. If someone doesn't have the next activity is fine. Then we can close this PR without merging.

PR #2874, opened 2020-07-29
add duration to walk egress leg after parking in mode choice

### LINKED ISSUE #2870: Are parking walk egress times not showing up in mode choice calculations?
General problem is that we're not seeing expected changes in mode choice when we limit parking capacity. My hypothesis is that the parking choice process is returning stalls that are cheap but have long walk access legs, and that the walk access leg is not being used in the mode choice process

Example is from here https://github.com/LBNL-UCB-STI/beam/blob/f7e48f22c796639dd886e345aa0c6f0056392ab4/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L917-L921 in mode choice process, where the parking search returned an emergency stall really far away, but the duration of the end walk leg in the car mode itinerary is still 0:

Screen Shot 2020-07-29 at 9 37 18 AM

zneedell:
Seems to work in SF-light-1k. Before the fix, if we run a scenario with very limited parking we get 3727 car mode choices. With the fix, we now get 2160 car mode choices, implying that considering walk egress time makes the car mode less attractive. As reference, unlimited parking gives 4796 car mode choices, implying that before the fix some people were still being driven away from choosing the car mode because of the cost associated with the scarce parking (which checks out with expectations).

Should fix #2870

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2874)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -801,7 +801,7 @@ trait ChoosesMode {
                   parkingResponse.stall.locationUTM
                 )
                 val travelTime: Int = (dist / ZonalParkingManager.AveragePersonWalkingSpeed).toInt
-                leg.copy(beamLeg = leg.beamLeg.copy(duration = travelTime))
+                leg.copy(beamLeg = leg.beamLeg.scaleToNewDuration(travelTime))
               } else {
                 leg
               }
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -796,8 +796,6 @@ trait ChoosesMode {
               if (i == 2) {
                 leg.copy(cost = leg.cost + parkingResponse.stall.costInDollars)
               } else if (i == 3) {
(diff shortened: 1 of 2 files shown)


zneedell:
Test now!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -795,6 +795,13 @@ trait ChoosesMode {
             case (leg, i) =>
               if (i == 2) {
                 leg.copy(cost = leg.cost + parkingResponse.stall.costInDollars)
+              } else if (i == 3) {
+                val dist = geo.distUTMInMeters(
+                  geo.wgs2Utm(leg.beamLeg.travelPath.endPoint.loc),
+                  parkingResponse.stall.locationUTM
+                )
+                val travelTime: Int = (dist / ZonalParkingManager.AveragePersonWalkingSpeed).toInt
+                leg.copy(beamLeg = leg.beamLeg.copy(duration = travelTime))
```
REASY, line 804: Should it be scaled via `BeamLeg.scaleToNewDuration` in order to get `travelPath.linkTravelTime` updated too?

zneedell:
Test!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
zneedell: Ah yeah good point

zneedell:
Test!

PR #2895, opened 2020-08-10
It/#2845 charging network manager

### LINKED ISSUE #2845: To Implement the ChargingNetworkManager
Implement the ChargingNetworkManager that undertakes a list of tasks every X time interval:
1) Estimate the required power in the next Y time interval > X time interval
2) if BEAM connected to grid: Communicate via HELICS/BeamFederate the estimated power and wait for Physical Bounds from the grid (min, max Power and price)
3) Otherwise: Use default Physical Bounds: Min = Max = RequiredPower & price = 0
4) Update PricingAndPhysicalBounds in the class SitePowerManager
5) Call getChargingPlanPerVehicle in SitePowerManager, which returns a map of vehicles with amount of energyInJoules to charge the corresponding vehicle
6) Loop over vehicles, and create a copy of the vehicle as a dummy vehicle and add energyInJoules.
7) If vehicle is fully charged before end of the time interval X, then delete dummy vehicle from memory (for now, don't send message protocol to the Driver via the scheduler)

Screen Shot 2020-07-15 at 11 08 39 AM

jctim:
Resolves #2845

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2895)


test/input/beamville/beam.conf
```diff
@@ -110,6 +110,11 @@ beam.router.skim = {
     fileBaseName = "skimsTAZ"
   }
 }
+#helics
+beam.cosim.helics = {
```
jctim, line 114: Should it be added?

jctim:
Test!

test/input/beamville/beam.conf
jctim: Yes I keep it. 
Enabling of ChargingNetworkManager is configured by a separate property

jctim:
Test!

jctim:
Test!

jctim:
Test!

jctim:
Test!

jctim:
Test!

jctim:
Test!

jctim:
Test!

jctim:
Test!

jctim:
Test!

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,249 @@
...
+  }
+
+  private def endRefuelSessionTriggerMaybe(
+    vehicle: BeamVehicle,
+    tick: Int,
+    maxChargeDuration: Long,
+    currentSession: ChargingSession,
+    totalSession: ChargingSession
+  ): Option[EndRefuelSessionTrigger] = {
+    if (vehicle.primaryFuelLevelInJoules >= vehicle.beamVehicleType.primaryFuelCapacityInJoule) {
```
JustinPihony, line 178: I would assume if it is more than the capacity then this is a bug and should result in an error and reset back to capacity?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,249 @@
...
+      )
+      Some(
+        EndRefuelSessionTrigger(
+          tick + currentSession.duration.toInt,
+          vehicle.getChargerConnectedTick(),
+          totalSession.energy,
+          vehicle
+        )
+      )
+    } else if (currentSession.duration < maxChargeDuration) {
```
JustinPihony, line 193: Shouldn't this be the other way around? If the duration is more than the max duration then it should be ended, otherwise it is under max duration and can continue?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,249 @@
...
+        EndRefuelSessionTrigger(
+          tick + currentSession.duration.toInt,
+          vehicle.getChargerConnectedTick(),
+          totalSession.energy,
+          vehicle
+        )
+      )
+    } else {
+      log.debug(
+        "Ending refuel cycle for vehicle {}. Provided {} J. during {}",
```
JustinPihony, line 211: Is this log message correct? We are not ending the refuel since we are sending a None, correct?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,249 @@
...
+    log.info("postStop")
+    if (gridConnectionEnabled) {
+      powerController.close()
+    }
+    super.postStop()
+  }
+}
+
+object ChargingNetworkManager {
+  final case class PlanningTimeOutTrigger(tick: Int) extends Trigger
```
JustinPihony, line 230: Should we call this something more explicit to what is being done - unless I misread something? My suggestion is ReevaluateChargingVehicles(Statuses?) or something like that.

src/main/scala/beam/agentsim/infrastructure/power/PowerController.scala
```diff
@@ -0,0 +1,85 @@
...
+  /**
+    * Publishes required power to the grid
+    *
+    * @param requiredPower power (in joules) over planning horizon
+    */
+  def publishPowerOverPlanningHorizon(requiredPower: Double, currentTime: Int): Unit = {
+    logger.debug("Sending power over planning horizon {} to the grid at time {}...", requiredPower, currentTime)
+    beamFederateMaybe
+      .map(_.publishPowerOverPlanningHorizon(requiredPower))
+      .getOrElse {
```
JustinPihony, line 36: This should be a match on None, the get is misleading since it does nothing.

src/main/scala/beam/agentsim/infrastructure/power/PowerController.scala
```diff
@@ -0,0 +1,85 @@
...
+        }
+    logger.debug("Obtained power from the grid {}...", bounds)
+    (bounds, nextTime)
+  }
+
+  def close(): Unit = {
+    logger.debug("Release PowerController resources...")
+    beamFederateMaybe
+      .map(_.close())
+      .getOrElse {
```
JustinPihony, line 68: Same here - it should be a match on None - or use isEmpty

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -360,27 +364,12 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with Stash {
...
-                        currentBeamVehicle
-                      ),
-                      self
-                    )
+                  if (ChargingPointType
+                        .getChargingPointInstalledPowerInKw(currentBeamVehicle.stall.get.chargingPointType.get) > 20.0) {
+
+                    log.debug("Sending ChargingPlugRequest to chargingNetworkManager at {}", tick)
+                    chargingNetworkManager ! ChargingPlugRequest(currentBeamVehicle, self)
+                    handleStartCharging(tick, currentBeamVehicle)
```
JustinPihony, line 372: Why was this moved inside the if - was it a bug to begin with?

JustinPihony (commented):
I did not review the tests as this was a large PR to review to begin - but others can review as needed.

src/main/python/helics/GridBroker.py
```diff
@@ -0,0 +1,102 @@
...
+    timebin = 300
+    currenttime = -1
+    rec_value = 0.0
+    time_sleep = 0
+
+    for t in range(0, timebin*360+1, timebin):
+        while currenttime < t:
+            currenttime = h.helicsFederateRequestTime(cfed, t)
+
+        if time_sleep > 0:
```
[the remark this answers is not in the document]
jctim: it was added for testing purposes. It was being modified during testing runs
Probably a comment would be needed

JustinPihony:
Also, I noticed that CI is timing out - so something has changed that kills the performance or breaks some flow.

src/main/scala/beam/agentsim/infrastructure/power/models.scala
```diff
@@ -0,0 +1,7 @@
+package beam.agentsim.infrastructure.power
+
+case class PhysicalBounds(minPower: Double, maxPower: Double, price: Double)
+
+object PhysicalBounds {
+  def default(power: Double): PhysicalBounds = PhysicalBounds(power, power, 0)
```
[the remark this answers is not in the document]
jctim: Turned to val, moved this class and companion object to PowerController as it is a part of model of it

src/main/scala/beam/sim/LoggingEventsManager.scala
```diff
@@ -34,21 +34,19 @@ class LoggingEventsManager @Inject()(config: Config) extends EventsManager with
   private val stacktraceToException: collection.mutable.HashMap[StackTraceElement, Exception] =
     collection.mutable.HashMap()
 
-  private var eventsAccumulator: Option[ActorRef] = None
+  private var eventsAccumulatorMaybe: Option[ActorRef] = None
```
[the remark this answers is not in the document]
jctim: I don' mind, changed to `xxxOption`

src/main/python/helics/GridBroker.py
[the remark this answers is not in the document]
jctim: added comment about that

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
jctim: Yes, this is correct

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
jctim: This name was taken from the sequence diagram from task description - https://github.com/LBNL-UCB-STI/beam/issues/2845

src/main/scala/beam/agentsim/infrastructure/power/PowerController.scala
jctim: Sounds reasonable, replaced on `option.fold()()`

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
jctim: It wasn't moved, the if was there before:

image

src/main/scala/beam/agentsim/infrastructure/power/PowerController.scala
jctim: replaced on `fold`

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,249 @@
...
+  private val beamConfig: BeamConfig = beamScenario.beamConfig
+  private val vehiclesToCharge: TrieMap[Id[BeamVehicle], ChargingVehicle] = new TrieMap()
+  private def vehicles: Map[Id[BeamVehicle], BeamVehicle] = vehiclesToCharge.mapValues(_.vehicle).toMap
+
+  private val sitePowerManager = new SitePowerManager()
+  private val powerController = new PowerController(beamServices, beamConfig)
+  private val endOfSimulationTime: Int = DateUtils.getEndOfTime(beamConfig)
+
+  private val gridConnectionEnabled = beamConfig.beam.agentsim.chargingNetworkManager.gridConnectionEnabled
+  log.info("ChargingNetworkManager should be connected to grid: {}", gridConnectionEnabled)
```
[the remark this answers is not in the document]
jctim: A bit vice versa:

On the top level we have a configuration whether `gridConnectionEnabled` should be enabled or not.

This means if it's `true` then we will try to connect to federate grid if it's possible. But this is a responsibility of `powerController` to try to establish connection with Helics and connect to the grid. That's why we are calling their method `isConnectedToGrid` ( - probably better solution is to rename this method on somethign `initConnection`)

But if `gridConnectionEnabled == false` then we even don't want to force `powerController` to establish connection with Helics and that's why we call its method `powerController.defaultPowerPhysicalBounds()` which is safe to call. I will add a comment to code about that

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,249 @@
...
+      vehiclesToCharge
+        .remove(vehicle.id)
+        .map { cv =>
+          val restChargeDuration = (tick % beamConfig.beam.cosim.helics.timeStep)
+          val totalEnergyAdjustedByTick =
+            if (cv.lastChargingSession.duration == 0) {
+              val (_, energy) = vehicle.refuelingSessionDurationAndEnergyInJoules(Some(restChargeDuration))
+              energy
+            } else {
+              val energyAdjustedByLastTick = cv.lastChargingSession.energy / cv.lastChargingSession.duration * restChargeDuration
```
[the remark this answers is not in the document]
jctim: Not sure what do you mean

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,249 @@
...
+    case ChargingUnplugRequest(vehicle, tick) =>
+      log.info("ChargingUnplugRequest for vehicle {} at {}", vehicle, tick)
+
+      vehiclesToCharge
+        .remove(vehicle.id)
+        .map { cv =>
+          val restChargeDuration = (tick % beamConfig.beam.cosim.helics.timeStep)
+          val totalEnergyAdjustedByTick =
+            if (cv.lastChargingSession.duration == 0) {
+              val (_, energy) = vehicle.refuelingSessionDurationAndEnergyInJoules(Some(restChargeDuration))
```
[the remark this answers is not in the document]
jctim: A have added comments with explanations to the code

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
jctim: Good point!

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
jctim: it could not be more, because we limit it by `maxChargeDuration` here:
```
val (chargingDuration, providedEnergy) = vehicle.refuelingSessionDurationAndEnergyInJoules(Some(maxChargeDuration))
```

jctim:
Test!

jctim:
Hey, @JustinPihony could you please take a look at my comments and changes which were applied?

src/main/python/helics/GridBroker.py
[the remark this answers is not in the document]
JustinPihony: I would comment it all out so it is known from the comment it is only useful if uncommented, but that leads to potential missed uncommenting. So, 2 things - is the random.uniform needed to be uncommented to be useful in this test, in which case you should uncomment that -- and the comment should maybe be more explicit that you have to set it:

> if time_sleep is changed to be > 0 then simulation ... in the loop (useful for testing purposes)

That way it's VERY explicit :)

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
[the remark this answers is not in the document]
JustinPihony: I agree the name change helps, but I wonder if the config should be flipped - `enableGridConnection` or `allowGridConnection` or `attemptGridConnection` - or something of that sort? It would probably fully close the loop more concretely.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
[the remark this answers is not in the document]
JustinPihony: No, there is a difference between  `(chargingEnergy / lastEnergy) * restDuration` and chargingEnergy / (lastEnergy * restDuration) - so the parentheses should be put in place explicitly

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
JustinPihony: I think part of my confusion was the usage of the word `over` - I had been in the mindset that it meant is it overcharged. Maybe rename the method to `isChargingCompleted`? Also, I saw your change from >= to == in the initial if. Maybe it's the worrier in me, but I would put a log or exception if it is > since it should never be...maybe a test to catch if we break that guarantee (if one isn't already there) - or both?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
JustinPihony: Thanks for the change to a boolean - makes more sense now :)

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
JustinPihony: @haitamlaarabi Thoughts on the name change - or the domain follows this structure more?

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
JustinPihony: Not the `if`, the `handleStartCharging` invocation. It was before the `if` and now it is inside the `if`

JustinPihony (commented):
Mostly fine now with some additional suggestions - but especially curious about the handleStartCharging.

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
jctim: Now I see. And yes, it was moved for a reason the charging can start only when the condition is `true`.
And if to leave the method outside of `if` then when we start handling charging - we emit the matsim event `ChargingPlugInEvent` but no any message will be sent to ChargingNetworkManager, that means charging actually not happens

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
test!

haitamlaarabi:
Test!

haitamlaarabi:
test now!

haitamlaarabi:
test now!

haitamlaarabi:
test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
test now!

haitamlaarabi:
@JustinPihony @REASY @wrashid Please don't merge this branch until we merge the patch.

haitamlaarabi:
Test!

JustinPihony (commented):
A bunch of comments - I still have a lot of the core logic to review, but wanted to get this out there since I may not complete before I get back.

src/main/scala/beam/sim/BeamMobsim.scala
```diff
@@ -506,15 +506,11 @@ class BeamMobsimIteration(
...
-      if (eventsAccumulator.isDefined) {
-        eventsAccumulator.get ! Finish
-        context.stop(eventsAccumulator.get)
-      }
+      chargingNetworkManager ! Finish
       context.stop(scheduler)
       context.stop(errorListener)
       context.stop(parkingManager)
       sharedVehicleFleets.foreach(context.stop)
-      context.stop(tazSkimmer)
```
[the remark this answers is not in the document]
haitamlaarabi: the stopping is happening within ChargingNetworkManager when it receives Finish

src/test/scala/beam/router/skim/TransitCrowdingSkimsTest.scala
```diff
@@ -7,6 +7,7 @@ import beam.agentsim.events.SpaceTime
 import beam.router.Modes.BeamMode.WALK_TRANSIT
 import beam.router.model.RoutingModel.TransitStopsInfo
 import beam.router.model.{BeamLeg, BeamPath, EmbodiedBeamLeg, EmbodiedBeamTrip}
+import beam.router.skim.core.TransitCrowdingSkimmer
```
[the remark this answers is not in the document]
haitamlaarabi: yes, I encapsulated certain parts of the skimmer inside sub packages

src/main/scala/beam/analysis/TransitOccupancyByStopAnalysis.scala
```diff
@@ -33,7 +34,13 @@ class TransitOccupancyByStopAnalysis extends BasicEventHandler with IterationEnd
       pathTraversalEvents.toList.groupBy(_.vehicleId.toString).map {
         case (k, v) => k -> v.map(_.numberOfPassengers.toString)
       }
-    val max = analysis.values.map(_.size).max
+
+    val max = if (analysis.nonEmpty) {
+      analysis.values.map(_.size).max
+    } else {
+      logger.debug("No Transit PathTraversal event was captured. Either transit is deactivated or something is broken")
```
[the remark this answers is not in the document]
haitamlaarabi: I changed it to warning

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

src/main/scala/beam/router/skim/readonly/ODSkims.scala
```diff
@@ -1,4 +1,4 @@
-package beam.router.skim
+package beam.router.skim.readonly
```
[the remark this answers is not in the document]
JustinPihony: I don't recall resolving this one, so re-opening

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -1066,6 +1056,10 @@ class PersonAgent(
   }
 
   val myUnhandled: StateFunction = {
+    case Event(BeamAgentSchedulerTimer, _) =>
+      // Put a breakpoitn here to see an internal state of the actor
+      log.info(s"Received message from ${sender()}")
```
JustinPihony, line 1061: Should this remain or be put behind a config at least?

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -360,28 +365,27 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with Stash {
...
-                        tick + sessionDuration.toInt,
-                        tick,
-                        energyDelivered,
-                        currentBeamVehicle
-                      ),
-                      self
-                    )
+                  log.debug("Sending ChargingPlugRequest to chargingNetworkManager at {}", tick)
+                  try {
+                    Await
```
JustinPihony, line 370: Why `Await`? That will ultimately bottleneck the performance - you could just as easily handle the logging debug/error messages from an `onComplete` or the like

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -360,28 +365,27 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with Stash {
...
+                      .result(
+                        chargingNetworkManager ? ChargingPlugRequest(
+                          tick,
+                          currentBeamVehicle,
+                          ChargingNetworkManager.defaultVehicleManager
```
JustinPihony, line 375: What is the purpose of sending a vehicle network manager? Wouldn't the manager being sent to be the one to use?

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -843,10 +846,13 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with Stash {
...
-    case Event(TriggerWithId(EndRefuelSessionTrigger(tick, _, _, vehicle), triggerId), _) =>
+    case Event(
+        TriggerWithId(EndRefuelSessionTrigger(tick, sessionDuration, fuelAddedInJoule, vehicle), triggerId),
+        _
+        ) =>
+      log.info(s"DrivesVehicle: EndRefuelSessionTrigger. tick: $tick, vehicle: $vehicle")
       if (vehicle.isConnectedToChargingPoint()) {
-        handleEndCharging(tick, vehicle)
-        vehicle.unsetParkingStall()
+        handleEndCharging(tick, vehicle, sessionDuration, fuelAddedInJoule)
```
JustinPihony, line 855: Should this even be doing this any longer? Or should it be delegated to the CNM? The two handling of end refueling seem very disjointed...unless I missed something. They both have an end method, but they don't seem tied together really - so the two can happen separately and alone, possibly leaving the state in a weird spot?

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -870,7 +876,14 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with Stash {
       0.0
   }
 
+  /**
+    * process the event ChargingPlugInEvent
```
JustinPihony, line 880: As mentioned above, should the start/end still be here or be in CNM? Is that what this comment is about? When _would_ be the time to delete it if not during this PR?

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -886,19 +899,27 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with Stash {
...
+    * @param chargingDuration the duration of charging session
+    * @param energyInJoules the energy in joules
     */
-  def handleEndCharging(currentTick: Int, vehicle: BeamVehicle): Unit = {
-
-    val (chargingDuration, energyInJoules) =
-      vehicle.refuelingSessionDurationAndEnergyInJoules(Some(currentTick - vehicle.getChargerConnectedTick()))
-
-    log.debug("Ending refuel session for {} in tick {}. Provided {} J.", vehicle.id, currentTick, energyInJoules)
-    vehicle.addFuel(energyInJoules)
```
JustinPihony, line 899: Same as the other comments in this file - should this be deleted - it has some changes already...so maybe just finish removing it?

JustinPihony (commented):
Still reviewing, but wanted to provide my review as I go. I'm about 2/3 through.

haitamlaarabi:
Test!

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -291,7 +293,7 @@ class PersonAgent(
 
   def resetFuelConsumed(): Unit = curFuelConsumed = FuelConsumed(0.0, 0.0)
 
-  override def logDepth: Int = 30
+  override def logDepth: Int = 3000
```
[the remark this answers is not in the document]
haitamlaarabi: Yes, it should be

src/main/scala/beam/agentsim/agents/PersonAgent.scala
haitamlaarabi: switched it to a log.debug instead

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
haitamlaarabi: it is not going to be a bottleneck. It's just a way to synchronize the the workflow and avoid juggling between states.

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
haitamlaarabi: there are different networks within the ChargingNetworkManager. But we will be handling this differently using traits.

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
haitamlaarabi: This is mostly for Ridehail CAVs. This is going to be the next step, that is having Ridehail CAVs communicating with the CNM

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
haitamlaarabi: That will happen once we refractor Ridehail as well. We are waiting for a Colin's patch to merge first.

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
haitamlaarabi: Same response

haitamlaarabi:
Test!

src/main/scala/beam/cosim/helics/BeamHelicsInterface.scala
```diff
@@ -0,0 +1,270 @@
...
+import com.github.beam.HelicsLoader
+import com.java.helics._
+import com.java.helics.helicsJNI._
+import com.typesafe.scalalogging.StrictLogging
+import spray.json.DefaultJsonProtocol.{listFormat, mapFormat, JsValueFormat, StringJsonFormat}
+import spray.json.{JsNumber, JsString, JsValue, _}
+
+object BeamHelicsInterface {
+  // Lazy makes sure that it is initialized only once
+  lazy val loadHelics: Unit = HelicsLoader.load()
```
JustinPihony, line 13: This is confusing from a naming convention. Admittedly it is hard to name - possibly `loadHelicsIfNotAlreadyLoaded`? I'm a fan of readability over anything else. Otherwise it does beg the question as to whether it should be called as many times as it is (more an architectural question).

src/test/scala/beam/integration/CarSharingSpec.scala
```diff
@@ -149,117 +168,150 @@ class CarSharingSpec extends FlatSpec with Matchers with BeamHelper {
...
-    }
-    beamScenario.privateVehicles.clear()
-
-    DefaultPopulationAdjustment(services).update(scenario)
-    services.controler.run()
-
-    assume(carsharingTripsIt0 == 0, "no agent is supposed to be driving in iteration 1")
-    assume(carsharingTripsIt1 > 0, "at least one agent has to be driving in iteration 2")
-  }
+  // REPOSITION IS NOT WORKING TODO IT NEEDS TO BE FIXED
```
JustinPihony, line 171: Is this supposed to be fixed before this is pushed through? If not then this should be an issue, lest it be forgotten.

JustinPihony (commented):
Down to the last 6 large files - sorry it has taken so long to complete the review. I'll complete fully tomorrow.

src/main/scala/beam/agentsim/infrastructure/power/SitePowerManager.scala
```diff
@@ -0,0 +1,153 @@
...
+
+  /**
+    * Get required power for electrical vehicles
+    *
+    * @param tick current time
+    * @return power (in Kilo Watt) over planning horizon
+    */
+  def requiredPowerInKWOverNextPlanningHorizon(tick: Int): Map[ChargingStation, PowerInKW] = {
+    lazy val planningFuture = Future
+      .sequence(allChargingStations.map { station =>
```
JustinPihony, line 34: Since you call `sequence` this is mostly a stylistic change than anything else, but you could change from Futures to parallel collections and it would likely reduce the code footprint. The biggest thing is that parallel collections do not have a simple mechanism for timeouts. So, if that is needed then simply resolve this and move on ;)

JustinPihony (commented):
Down to 3 files needed to review.

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+          case Some(cycle: ChargingCycle) if cycle.startTime == startTime && cycle.duration > duration =>
+            chargingSessionInternal.remove(chargingSessionInternal.size - 1)
+            val cycle = ChargingCycle(startTime, energy, duration)
+            chargingSessionInternal.append(cycle)
+            Some(cycle)
+          case Some(cycle: ChargingCycle) if cycle.startTime == startTime =>
+            // keep existing charging cycle
+            logger.debug(s"the new charging cycle of vehicle $vehicle ends after the end of the last cycle")
+            None
+          case Some(cycle: ChargingCycle) if cycle.startTime > startTime =>
```
JustinPihony, line 226: I would put this as `startTime < cycle.startTime` strictly from a readability perspective. That way it reads as the error points out: "If the new start time is before the stored start time.."

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+      this.synchronized {
+        chargingSessionInternal.lastOption match {
+          case Some(cycle: ChargingCycle) if cycle.startTime == startTime && cycle.duration > duration =>
+            chargingSessionInternal.remove(chargingSessionInternal.size - 1)
+            val cycle = ChargingCycle(startTime, energy, duration)
+            chargingSessionInternal.append(cycle)
+            Some(cycle)
+          case Some(cycle: ChargingCycle) if cycle.startTime == startTime =>
+            // keep existing charging cycle
+            logger.debug(s"the new charging cycle of vehicle $vehicle ends after the end of the last cycle")
```
JustinPihony, line 224: This debug is not true - it would be that the cycles are equal. Or is it because we checked the duration in the last match and we know this duration is longer than the previous for the same start time?

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+      * adding a new charging cycle to the charging session
+      * @param startTime start time of the charging cycle
+      * @param energy energy delivered
+      * @param duration duration of charging
+      * @return boolean value expressing if the charging cycle has been added
+      */
+    def processChargingCycle(startTime: Int, energy: Double, duration: Long): Option[ChargingCycle] =
+      this.synchronized {
+        chargingSessionInternal.lastOption match {
+          case Some(cycle: ChargingCycle) if cycle.startTime == startTime && cycle.duration > duration =>
```
JustinPihony, line 217: Should this even happen at all? And why do we opt to choose the lesser duration - what makes that more correct than the original?

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+            None
+          case _ =>
+            val cycle = ChargingCycle(startTime, energy, duration)
+            chargingSessionInternal.append(cycle)
+            Some(cycle)
+        }
+      }
+
+    def computeSessionStartTime: Int = this.synchronized {
+      chargingSessionInternal.headOption.map(_.startTime).getOrElse(arrivalTime)
```
JustinPihony, line 238: I assume this is different because arrival time is the arrival at the station, where they could be put into waiting?

JustinPihony (commented):
I still have the ChargingNetworkManager and PowerController, then I will have reviewed it all. Admittedly I am still unsure about all the mutable state that is not encapsulated inside an Actor (as you can tell by my comments).

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+              case (_, chargingVehicle @ ChargingVehicle(vehicle, vehicleManager, _, _, _, _)) =>
+                Future {
+                  // Refuel
+                  handleRefueling(chargingVehicle)
+                  // Calculate the energy to charge each vehicle connected to the a charging station
+                  val (chargingDuration, energyToCharge) =
+                    dispatchEnergy(cnmConfig.timeStepInSeconds, chargingVehicle, physicalBounds)
+                  // update charging vehicle with dispatched energy and schedule ChargingTimeOutScheduleTrigger
+                  chargingVehicle.processChargingCycle(timeBin, energyToCharge, chargingDuration).flatMap {
+                    case cycle if cycle.duration == 0 =>
```
JustinPihony, line 95: This should be changed to a `def` that names what this `if` means? I still am not 100%....following the logic it seems that the current impl is that the duration is really just the `cnmConfig.timeStepInSeconds` (dispatch returns the duration as the timestep, then process passes the duration back out (unless it is the same starttime))..

JustinPihony: Actually...even the further cycles are confusing me...Will ping offline to discuss the final logic a bit more.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+    log.debug(s"ChargingPlugInEvent: $chargingPlugInEvent")
+    beamServices.matsimServices.getEvents.processEvent(chargingPlugInEvent)
+  }
+
+  /**
+    * Calculates the duration of the refuel session, the provided energy and throws corresponding events
+    * @param currentTick current time
+    * @param chargingVehicle vehicle charging information
+    */
+  def handleEndChargingHelper(currentTick: Int, chargingVehicle: ChargingVehicle): Unit = {
```
JustinPihony, line 308: This needs renamed - I believe it used to be something that was for a more functional style of code where the name made a bit more sense, but now it doesn't. In fact maybe it should just become part of handleEndCharging?

JustinPihony: Although for readability - maybe renaming is better - processEndChargingEvents?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+    beamServices.matsimServices.getEvents.processEvent(refuelSessionEvent)
+    val chargingPlugOutEvent: ChargingPlugOutEvent = new ChargingPlugOutEvent(
+      currentTick,
+      stall.copy(locationUTM = geo.utm2Wgs(stall.locationUTM)),
+      vehicle.id,
+      vehicle.primaryFuelLevelInJoules,
+      Some(vehicle.secondaryFuelLevelInJoules)
+    )
+    log.debug(s"ChargingPlugOutEvent: $chargingPlugOutEvent")
+    beamServices.matsimServices.getEvents.processEvent(chargingPlugOutEvent)
```
JustinPihony, line 336: Should this be done here? Shouldn't it be done alongside the actual plugout?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+    val chargingNetwork = chargingNetworkMap(vehicleManager)
+    if (chargingNetwork.disconnectVehicle(vehicle)) {
+      handleRefueling(chargingVehicle)
+      handleEndChargingHelper(tick, chargingVehicle)
+      vehicle.disconnectFromChargingPoint()
+      vehicle.stall match {
+        case Some(stall) =>
+          parkingManager ! ReleaseParkingStall(stall.parkingZoneId, stall.tazId)
+          vehicle.unsetParkingStall()
+        case None =>
```
JustinPihony, line 261: As above - if this happens then the disconnectVehicle will throw and it will never reach this point anyway.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
JustinPihony: Same questions for the starthelper btw.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+    * @param chargingVehicle charging vehicle information
+    */
+  private def handleStartCharging(tick: Int, chargingVehicle: ChargingVehicle): Unit = {
+    val ChargingVehicle(vehicle, vehicleManager, _, _, _, theSender) = chargingVehicle
+    log.debug(s"Starting charging for vehicle $vehicle at $tick")
+    val physicalBounds = obtainPowerPhysicalBounds(tick, None)
+    vehicle.connectToChargingPoint(tick)
+    theSender ! StartingRefuelSession(tick, vehicle.id)
+    handleStartChargingHelper(tick, chargingVehicle)
+    val (chargeDurationAtTick, energyToChargeAtTick) =
```
JustinPihony, line 232: Everything after this seems a bit duplicated - shouldn't this be added and then let the PlanningTimeOutTrigger handle the below? Unless this is to handle the scenario from the time it starts UNTIL the next planning - which seems to be more forward calculation than reverse calculation? Maybe that is why you do the handleRefueling at the start of the planning - to handle this scenario? If so, then this is rather confusing and I still argue that you run a high risk of double refueling for 0 duration. This whole thing should be put into the same logic flow so it can only do the fuel dance in one place at one time. It seems kind of like doing a while loop where you have to prime it - correct. In this actor logic it is too disjointed and needs rewritten for clarity IMHO - lmk if that makes sense - or if I am totally off base.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+        )
+      }
+
+    case ChargingUnplugRequest(tick, vehicle, vehicleManager) =>
+      log.debug(s"ChargingUnplugRequest received for vehicle $vehicle from plug ${vehicle.stall} at $tick")
+      val physicalBounds = obtainPowerPhysicalBounds(tick, None)
+      val chargingNetwork = chargingNetworkMap(vehicleManager)
+      chargingNetwork.lookupVehicle(vehicle.id) match {
+        case Some(chargingVehicle) =>
+          val prevStartTime = chargingVehicle.latestChargingCycle.get.startTime
```
JustinPihony, line 200: My head is a bit turned around, so this is just a question to make sure - Does the min/max logic make it so that this handles all scenarios: plugged in but unplugged before the first planning AND plugged in, planned, and then unplugged? I assume it handles the latter, but does it the former?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+          val (chargeDurationAtTick, energyToChargeAtTick) = dispatchEnergy(duration, chargingVehicle, physicalBounds)
+          chargingVehicle.processChargingCycle(startTime, energyToChargeAtTick, chargeDurationAtTick)
+          handleEndCharging(tick, chargingVehicle, Some(sender))
+        case _ =>
+          log.debug(s"Vehicle $vehicle is already disconnected at $tick")
+          sender ! UnhandledVehicle(tick, vehicle.id)
+      }
+
+    case Finish =>
+      chargingNetworkMap.foreach(_._2.clear())
```
JustinPihony, line 213: Should this do a final pass of the vehicle usage - I know we have `// if we still have a BEV/PHEV that is connected to a charging point,`, but want to see if it's worth a final pass through?

JustinPihony (commented):
There are some definite questions where I think there is either logic I misread or possible gaps in edge cases. I know I put a ton of comments overall. LMK if it would be better for me to make some PRs, pair on it, review my comments overall, or something else? Some comments came from my order of reviewing the code, so may be obsolete once I understood better. Feel free to say I am wrong and resolve if so - I'll yell if I still disagree ;)

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -69,36 +70,36 @@ trait ChoosesParking extends {
...
       stash()
       stay using data
     case Event(StateTimeout, data: BasePersonData) =>
       val (tick, _) = releaseTickAndTriggerId()
-
       if (currentBeamVehicle.isConnectedToChargingPoint()) {
-        handleEndCharging(tick, currentBeamVehicle)
+        log.debug("Sending ChargingUnplugRequest to ChargingNetworkManager at {}", tick)
+        try {
+          Await
```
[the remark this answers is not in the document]
haitamlaarabi: Okay, I have changed that to a state gymnastic

src/main/scala/beam/cosim/helics/BeamHelicsInterface.scala
haitamlaarabi: good point. name changed.

src/test/scala/beam/integration/CarSharingSpec.scala
haitamlaarabi: It has been fixed

src/main/scala/beam/cosim/helics/BeamHelicsInterface.scala
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+      * close HELICS library
+      */
+    def close(): Unit = {
+      if (helics.helicsFederateIsValid(fedComb) == 1) {
+        helics.helicsFederateFinalize(fedComb)
+        helics.helicsFederateFree(fedComb)
+        helics.helicsCloseLibrary()
+        logger.debug(s"closing BeamFederate")
```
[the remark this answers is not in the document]
haitamlaarabi: OK

src/main/scala/beam/cosim/helics/BeamHelicsInterface.scala
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  /**
+    * Create a Federate Instance
+    * @param fedName FEDERATE_NAME
+    * @param dataOutStreamPointMaybe "PUBLICATION_NAME"
+    * @param dataInStreamPointAndBufferSizeMaybe ("FEDERATE_NAME/SUBSCRIPTION_NAME" , "BUFFER_SIZE")
+    * @return
+    */
+  def getFederateInstance(
```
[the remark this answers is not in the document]
haitamlaarabi: OK

src/main/scala/beam/agentsim/infrastructure/power/SitePowerManager.scala
```diff
@@ -0,0 +1,153 @@
...
+import scala.concurrent.duration._
+import scala.concurrent.{Await, Future, TimeoutException}
+
+class SitePowerManager(chargingNetworkMap: TrieMap[String, ChargingNetwork], beamServices: BeamServices)
+    extends LazyLogging {
+  import SitePowerManager._
+
+  private val cnmConfig = beamServices.beamConfig.beam.agentsim.chargingNetworkManager
+  private val tazSkimmer = beamServices.skims.taz_skimmer
+  private val allChargingStations = chargingNetworkMap.flatMap(_._2.lookupStations).toList.distinct
```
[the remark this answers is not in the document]
haitamlaarabi: lazy is fine

src/main/scala/beam/agentsim/infrastructure/power/SitePowerManager.scala
```diff
@@ -0,0 +1,153 @@
...
+    lazy val planningFuture = Future
+      .sequence(allChargingStations.map { station =>
+        Future {
+          station -> observedPowerDemandInKW(tick, station.zone).getOrElse(estimatePowerDemandInKW(tick, station.zone))
+        }
+      })
+      .map(_.toMap)
+      .recover {
+        case e =>
+          logger.debug(s"Charging Replan did not produce allocations: $e")
```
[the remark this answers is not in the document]
haitamlaarabi: yep

src/main/scala/beam/agentsim/infrastructure/power/SitePowerManager.scala
```diff
@@ -0,0 +1,153 @@
...
+  private val unlimitedPhysicalBounds = getUnlimitedPhysicalBounds(allChargingStations).value
+
+  /**
+    * Get required power for electrical vehicles
+    *
+    * @param tick current time
+    * @return power (in Kilo Watt) over planning horizon
+    */
+  def requiredPowerInKWOverNextPlanningHorizon(tick: Int): Map[ChargingStation, PowerInKW] = {
+    lazy val planningFuture = Future
```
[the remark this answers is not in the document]
haitamlaarabi: yep

src/main/scala/beam/agentsim/infrastructure/power/SitePowerManager.scala
haitamlaarabi: I think it is a good idea. thanks

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+    * @return charging vehicle
+    */
+  def lookupVehicle(vehicleId: Id[BeamVehicle]): Option[ChargingVehicle] =
+    lookupVehicles.get(vehicleId)
+
+  /**
+    * get all stations
+    * @return list of station
+    */
+  def lookupStations: List[ChargingStation] =
```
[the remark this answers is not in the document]
haitamlaarabi: yep

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+import scala.collection.JavaConverters._
+import scala.collection.concurrent.TrieMap
+import scala.collection.mutable
+import scala.collection.mutable.ListBuffer
+
+/**
+  * Created by haitamlaarabi
+  */
+
+class ChargingNetwork(vehicleManagerName: VehicleManager, chargingStationsQTree: QuadTree[ChargingZone])
```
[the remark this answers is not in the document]
haitamlaarabi: it has been changed to managerId

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+    * @param tick current time
+    * @param vehicle vehicle to charge
+    * @return a tuple of the status of the charging vehicle and the connection status
+    */
+  def connectVehicle(tick: Int, vehicle: BeamVehicle, theSender: ActorRef): ChargingVehicle = {
+    if (vehicle.stall.isEmpty) {
+      throw new RuntimeException(s"Vehicle $vehicle doesn't have a stall!")
+    } else {
+      val stall = vehicle.stall.get
+      lookupStation(stall).map(station => station.connect(tick, vehicle, stall, theSender)).get
```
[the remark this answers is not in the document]
haitamlaarabi: named changed

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
haitamlaarabi: yep

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
haitamlaarabi: yes

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
haitamlaarabi: the lesser duration is always correct, because it means that vehicle received and a disconnect signal before end of the cycle

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
haitamlaarabi: you are right. I changed the code to cater for waiting in line

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+  parkingManager: ActorRef,
+  scheduler: ActorRef
+) extends Actor
+    with ActorLogging {
+  import ChargingNetworkManager._
+  import ConnectionStatus._
+  import beamServices._
+
+  private val beamConfig: BeamConfig = beamScenario.beamConfig
+  private val cnmConfig = beamConfig.beam.agentsim.chargingNetworkManager
```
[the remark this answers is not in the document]
haitamlaarabi: Also having a long name make the code less readable.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+  private val chargingNetworkMap = TrieMap.empty[String, ChargingNetwork]
+  private val chargingZoneList: QuadTree[ChargingZone] = loadChargingZones(beamServices)
+  private lazy val sitePowerManager = new SitePowerManager(chargingNetworkMap, beamServices)
+  private lazy val powerController = new PowerController(chargingNetworkMap, beamConfig)
+  private val endOfSimulationTime: Int = DateUtils.getEndOfTime(beamConfig)
+
+  import powerController._
+  import sitePowerManager._
+
+  private def currentTimeBin(tick: Int): Int = cnmConfig.timeStepInSeconds * (tick / cnmConfig.timeStepInSeconds)
```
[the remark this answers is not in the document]
haitamlaarabi: Not really. Since they are Integers it gets round up and then multiplied to obtain the start of the time interval

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+  private def nextTimeBin(tick: Int): Int = currentTimeBin(tick) + cnmConfig.timeStepInSeconds
```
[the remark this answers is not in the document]
haitamlaarabi: same

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+
+  override def receive: Receive = {
+    case TriggerWithId(InitializeTrigger(_), triggerId) =>
+      import scala.concurrent.{ExecutionContext, Future}
+      implicit val timeout: Timeout = Timeout(50000, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
haitamlaarabi: yep

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+      import scala.concurrent.{Await, Future, TimeoutException}
+      val estimatedLoad = requiredPowerInKWOverNextPlanningHorizon(timeBin)
+      log.debug("Total Load estimated is {} at tick {}", estimatedLoad.values.sum, timeBin)
+
+      // obtaining physical bounds
+      val physicalBounds = obtainPowerPhysicalBounds(timeBin, Some(estimatedLoad))
+
+      val futures = Future
+        .sequence(chargingNetworkMap.flatMap {
+          case (_, chargingNetwork) =>
```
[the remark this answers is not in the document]
haitamlaarabi: The goal is to process all managers here

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: good idea

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+  private val endOfSimulationTime: Int = DateUtils.getEndOfTime(beamConfig)
+
+  import powerController._
+  import sitePowerManager._
+
+  private def currentTimeBin(tick: Int): Int = cnmConfig.timeStepInSeconds * (tick / cnmConfig.timeStepInSeconds)
+  private def nextTimeBin(tick: Int): Int = currentTimeBin(tick) + cnmConfig.timeStepInSeconds
+
+  override def receive: Receive = {
+    case TriggerWithId(InitializeTrigger(_), triggerId) =>
```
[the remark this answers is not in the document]
haitamlaarabi: maybe yes.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+      vehicle.primaryFuelLevelInJoules,
+      Some(vehicle.secondaryFuelLevelInJoules)
+    )
+    log.debug(s"ChargingPlugOutEvent: $chargingPlugOutEvent")
+    beamServices.matsimServices.getEvents.processEvent(chargingPlugOutEvent)
+  }
+}
+
+object ChargingNetworkManager {
+  case class PlanningTimeOutTrigger(tick: Int) extends Trigger
```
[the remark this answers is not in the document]
haitamlaarabi: Ok

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+  /**
+    * Refuel the vehicle using last charging session and collect the corresponding load
+    * @param chargingVehicle vehicle charging information
+    */
+  private def handleRefueling(chargingVehicle: ChargingVehicle): Unit = {
+    chargingVehicle.latestChargingCycle.foreach { cycle =>
+      val vehicle = chargingVehicle.vehicle
+      vehicle.addFuel(cycle.energy)
+      collectObservedLoadInKW(chargingVehicle, cycle)
+      log.debug(s"Charging vehicle $vehicle. Stall ${vehicle.stall}. Provided energy of = ${cycle.energy} J")
```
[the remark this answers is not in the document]
haitamlaarabi: ok

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,477 @@
...
+      val physicalBounds = obtainPowerPhysicalBounds(timeBin, Some(estimatedLoad))
+
+      val futures = Future
+        .sequence(chargingNetworkMap.flatMap {
+          case (_, chargingNetwork) =>
+            chargingNetwork.lookupConnectedVehicles.map {
+              case (_, chargingVehicle @ ChargingVehicle(vehicle, vehicleManager, _, _, _, _)) =>
+                Future {
+                  // Refuel
+                  handleRefueling(chargingVehicle)
```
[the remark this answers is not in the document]
haitamlaarabi: that refueling concerns the previous charging cycle and not the current one. that's how I'm avoiding double refueling.

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: it is

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: yes. But I remember there was some concurrency issue around it. I will remove that and see what happens

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: Yes it is convoluted to avoid double refueling or not refueling vehicles. The code is set with fallbacks. I tried to make it easier, but I didn't find any easier way to do so

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: yes it handles both with this
```
          val startTime = Math.min(tick, prevStartTime)
          val endTime = Math.max(tick, prevStartTime)
```

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: I do a final pass here:
    ```
  // rescheduling the PlanningTimeOutTrigger
      sender ! CompletionNotice(
        triggerId,
        triggers ++ (if (nextTimeBin(timeBin) < endOfSimulationTime) {
                       Vector(ScheduleTrigger(PlanEnergyDispatchTrigger(nextTimeBin(timeBin)), self))
                     } else {
                       // if we still have a BEV/PHEV that is connected to a charging point,
                       // we assume that they will charge until the end of the simulation and throwing events accordingly
                       chargingNetworkMap
                         .flatMap(_._2.vehicles)
                         .map {
                           case (_, chargingVehicle @ ChargingVehicle(vehicle, stall, _, _, _, _)) =>
                             if (chargingVehicle.status == Connected) {
                               val (duration, energy) = dispatchEnergy(Int.MaxValue, chargingVehicle, physicalBounds)
                               chargingVehicle.processChargingCycle(timeBin, energy, duration)
                             }
                             ScheduleTrigger(
                               ChargingTimeOutTrigger(nextTimeBin(timeBin) - 1, vehicle.id, stall.managerId),
                               self
                             )
                         }
                         .toVector
                     })
      )
```

File: src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala [hunk 58]
```diff
@@ -0,0 +1,246 @@
...
+  /**
+    * get name of the vehicle manager
+    * @return VehicleManager
+    */
+  def vehicleManager: VehicleManager = vehicleManagerName
+
+  /**
+    * clear charging vehicle map
+    */
+  def clear(): Unit = chargingStationMap.foreach(_._2.clear())
```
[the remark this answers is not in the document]
haitamlaarabi: ok

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+
+    /**
+      * adding a new charging cycle to the charging session
+      * @param startTime start time of the charging cycle
+      * @param energy energy delivered
+      * @param duration duration of charging
+      * @return boolean value expressing if the charging cycle has been added
+      */
+    def processChargingCycle(startTime: Int, energy: Double, duration: Long): Option[ChargingCycle] =
+      this.synchronized {
```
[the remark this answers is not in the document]
haitamlaarabi: It's a good point. I will keep it this way for now and I will need to think about possible turning them into Actors

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -0,0 +1,246 @@
...
+
+/**
+  * Created by haitamlaarabi
+  */
+
+class ChargingNetwork(vehicleManagerName: VehicleManager, chargingStationsQTree: QuadTree[ChargingZone])
+    extends LazyLogging {
+  import ChargingNetwork._
+
+  private val chargingStationMap: Map[ChargingZone, ChargingStation] =
```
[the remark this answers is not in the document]
haitamlaarabi: Well for now we are assuming that stations are giant zones. This might change in the future

haitamlaarabi:
Test!

haitamlaarabi:
Test!

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -0,0 +1,447 @@
...
+        })
+        .map(_.toVector)
+        .recover {
+          case e =>
+            log.debug(s"No energy was dispatched at time $timeBin because: $e")
+            Vector.empty[Option[ScheduleTrigger]]
+        }
+
+      val triggers = (try {
+        Await.result(futures, atMost = 1.minutes)
```
REASY, line 128: nitpick: instead of blocking you can use `pipeTo` to send the result back to yourself
https://doc.akka.io/docs/akka/2.5/futures.html#use-the-pipe-pattern

REASY (approved):

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: this is much better. I replaced it with:
`        .map { triggers =>
          val nextStepPlanning = if (!isEndOfSimulation(timeBin)) {
            Vector(ScheduleTrigger(PlanEnergyDispatchTrigger(nextTimeBin(timeBin)), self))
          } else {
            completeChargingAndTheDisconnectionOfAllConnectedVehiclesAtEndOfSimulation(timeBin, physicalBounds)
          }
          CompletionNotice(triggerId, triggers.flatten.toVector ++ nextStepPlanning)
        }
        .pipeTo(sender())`

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test now!

haitamlaarabi (approved):

PR #2898, opened 2020-08-10
GoogleSpeedApi: specify exact iteration number for different kind of google speed requests

nikolayilyin:
because requests cost money and we want to do very specific '3am' requests at iteration 0 and requests for regular route time at last iteration.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2898)


nikolayilyin:
there is a problem - the result file should contain regular route time even if there were only 3am requests made.

PR #2904, opened 2020-08-13
Google Routes DB

aaltergot:


src/main/python/google_routes_db/google_routed_db_ddl.sql
```diff
@@ -0,0 +1,42 @@
+CREATE TABLE gr_route (
+  id BIGSERIAL PRIMARY KEY,
+  bound_northeast geography(POINT),
```
SHaaD94, line 3: > None of the GEOS functions support the geography type. As a workaround one can convert back and forth between geometry and geography types.
https://postgis.net/docs/using_postgis_dbmanagement.html#PostGIS_Geography

Hence, **geometry** type is better

SHaaD94:
Could you please specify the queries you are going to use with this table?

REASY:
@aaltergot couple of things:
1) Please, take a look to the closed PR from @SHaaD94 which contains Scala code to work with PostgreSQL (https://github.com/LBNL-UCB-STI/beam/pull/2357/files). I'd prefer to have this as Scala/JVM app rather than Python
2) As a part of this task could you, please, collect all the JSON files from all the runs and create a database from it?

src/main/scala/beam/utils/google_routes_db/ReqRespMapping.scala
```diff
@@ -0,0 +1,210 @@
...
+  private implicit val execCtx: ExecutionContext = system.dispatcher
+  private implicit val jsonStreamingSupport: JsonEntityStreamingSupport = EntityStreamingSupport.json()
+
+  private var csvWriter: CsvWriter = _
+
+  //noinspection DuplicatedCode
+  def main(args: Array[String]): Unit = try {
+    val (_, cfg) = prepareConfig(args, isConfigArgRequired = true)
+    val config = GoogleRoutesDBConfig(cfg)
+    val geoUtils = SimpleGeoUtils()
```
REASY, line 41: In order to get a good precision there should be a way to provide CRS (localCRS) instead of using default in that class which is `epsg:26910`

src/main/scala/beam/utils/google_routes_db/GoogleRoutesDB.scala
```diff
@@ -0,0 +1,178 @@
...
+        batchUpdateResult.itemsProcessed,
+        batchUpdateResult.rowsUpdated
+      )}
+      _ ← doneFuture
+    } yield Done
+
+    allDoneFuture.onComplete {
+      case Success(_) ⇒
+      case Failure(e) ⇒
+        logger.error("An error occurred: {}", e.getMessage)
```
REASY, line 105: nitpick: pass an exception to error method

src/main/scala/beam/utils/google_routes_db/ReqRespMapping.scala
```diff
@@ -0,0 +1,210 @@
...
+    csvWriter = new CsvWriter(s"ReqRespMapping.csv", header)
+
+    val doneFuture: Future[Done] = sourceGoogleapiFiles(config)
+      .flatMapConcat { googleapiFiles ⇒
+        Source(epsilons).map(epsilon ⇒ (googleapiFiles, epsilon))
+      }
+      .toMat(Sink.foreach { case (googleapiFiles, epsilon) ⇒
+        val requests =
+          req.csv.parseGoogleTravelTimeEstimationCsv(
+            googleapiFiles.googleTravelTimeEstimationCsvText
```
REASY, line 66: Will it download the same files `epsilons.length` times? If yes, should it be cached?

REASY (changes requested):

src/main/scala/beam/utils/google_routes_db/ReqRespMapping.scala
aaltergot: I don't think so. `sourceGoogleapiFiles` yields downloaded contents in _xxxText_ fields, i.e _googleTravelTimeEstimationCsvText_, which is parsed here.

src/main/scala/beam/utils/google_routes_db/ReqRespMapping.scala
aaltergot: Added a config value

src/main/scala/beam/utils/google_routes_db/GoogleRoutesDB.scala
```diff
@@ -0,0 +1,178 @@
...
+ * Run with Gradle:
+ * ./gradlew :execute \
+ *   -PmainClass=beam.utils.google_routes_db.GoogleRoutesDB \
+ *   -PappArgs="['--config','src/main/scala/beam/utils/google_routes_db/config/google_routes_db.conf']" \
+ *   -PlogbackCfg=logback.xml
+ */
+object GoogleRoutesDB extends BeamHelper {
+
+  private implicit val system: ActorSystem = ActorSystem("google-routes-db")
+  private implicit val execCtx: ExecutionContext = system.dispatcher
```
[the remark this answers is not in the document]
aaltergot: Replaced with cached thread pool.

aaltergot:
test now!

src/main/scala/beam/agentsim/events/handling/GoogleTravelTimeEstimationEntry.scala
```diff
@@ -0,0 +1,150 @@
...
+    )
+
+    def writeEntries(entries: Seq[GoogleTravelTimeEstimationEntry], filePath: String): Unit = {
+      using(new CsvWriter(filePath, headers)) { csvWriter =>
+        entries.map(GoogleTravelTimeEstimationEntry.toRow).foreach(csvWriter.writeRow)
+      }
+    }
+
+    def parseEntries(text: String): Seq[GoogleTravelTimeEstimationEntry] = {
+      val csvReader = new CsvMapReader(new StringReader(text), CsvPreference.STANDARD_PREFERENCE)
```
JustinPihony, line 136: This should be in a using (maybe the string reader also, but the csvmapreader probably handles it in it's close - or should ;) )

src/main/scala/beam/utils/google_routes_db/GoogleRoutesDB.scala
```diff
@@ -0,0 +1,120 @@
...
+  ): Map[PathTraversalEvent, Seq[GoogleTravelTimeEstimationEntry]] = {
+    using(con.prepareStatement(sql.Select.QueryGoogleRoutes.sql)) { ps =>
+      val result = mutable.Map.empty[PathTraversalEvent, Seq[GoogleTravelTimeEstimationEntry]]
+
+      ptes.foreach { pte =>
+        import sql.Select.QueryGoogleRoutes._
+
+        val arg = makeArg(pte, departureDateTime)
+        argPsMapping.mapPrepared(arg, ps)
+        val items = using(ps.executeQuery()) { rs => readItems(rs) }
```
JustinPihony, line 37: Should we batch these queries? It should be profiled on being hit by multiple PTE queries versus sending as a single transaction.

src/main/scala/beam/utils/google_routes_db/GoogleRoutesDB.scala
```diff
@@ -0,0 +1,120 @@
...
+  def insertGoogleRoutesAndLegs(
+    grrSeq: Seq[GoogleRoutesResponse],
+    requestIdToDepartureTime: Map[String, Int],
+    googleapiResponsesJsonFileUri: String,
+    timestamp: Instant,
+    con: Connection
+  ): Unit = {
+    grrSeq.foreach { grr: GoogleRoutesResponse =>
+
+      val routeItemToRoute: Map[Update.GoogleRouteItem, DirectionsRoute] =
```
JustinPihony, line 80: Should this be wrapped in a transaction? What are the ramifications of a create with a failed insert? Ultimately it should be a tx or self healing.

src/main/scala/beam/utils/google_routes_db/doc/README.md
```diff
@@ -0,0 +1,29 @@
+# TravelTimeGoogleStatistic: Build database of the computed routes
```
JustinPihony, line 1: Should you document how to update the db config - or add something to gradle to generate it when the base beam config is updated?

src/main/scala/beam/utils/google_routes_db/sql/package.scala
```diff
@@ -0,0 +1,21 @@
+package beam.utils.google_routes_db
+
+import com.google.maps.model.LatLng
+
+package object sql {
+
+  type GeometryPoint = String
+  type GeometryLinestring = String
+
+  val projection: Int = 4326
```
JustinPihony, line 10: What is this really? Can we make the naming more descriptive?

JustinPihony (changes requested):
Some leaks - also curious if you contemplated using something like Quill (https://github.com/getquill/quill) to avoid writing our own SQL directly?

aaltergot:
test now!

src/main/scala/beam/utils/google_routes_db/GoogleRoutesDB.scala
aaltergot: Transaction is managed externally to this method by the creator of the connection. Effectively, the whole method is in one transaction. Added try-catch on the client side (TravelTimeGoogleStatistic) also.

aaltergot:
test now!

aaltergot:
test now!

src/main/scala/beam/utils/google_routes_db/sql/package.scala
JustinPihony: I'm still not quite sure the new name is much better, but maybe I am not familiar enough with the underlying names of this domain?

PR #2934, opened 2020-08-23
Fix flaky tests

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2934)

### DIFF
--- a/src/test/scala/beam/cosim/helics/BeamFederateSpec.scala
+++ b/src/test/scala/beam/cosim/helics/BeamFederateSpec.scala
@@ -54,14 +54,8 @@ class BeamFederateSpec extends FlatSpec with Matchers with BeamHelper with Befor
     } yield (f1Result, f2Result)
     try {
       Await.result(aggregatedFuture, 5.minutes)
-      assert(
-        chargingPlugInEvents.get() > 0,
-        s"Charging Plugin Event expected to be greater than 0 but is ${chargingPlugInEvents.get}"
-      )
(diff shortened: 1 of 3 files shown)


src/test/scala/beam/cosim/helics/BeamFederateSpec.scala
```diff
@@ -53,8 +54,14 @@ class BeamFederateSpec extends FlatSpec with Matchers with BeamHelper with Befor
...
-      require(chargingPlugInEvents.get() > 0)
-      require(chargingPlugOutEvents.get() > 0)
+      assert(
+        chargingPlugInEvents.get() > 0,
+        s"Charging Plugin Event expected to be greater than 0 but is ${chargingPlugInEvents.get}"
+      )
+      assert(
+        chargingPlugOutEvents.get() > 0,
+        s"Chaging Plugout Event expected to be greater than 0 but is ${chargingPlugOutEvents.get}"
+      )
```
carloscaldas, line 64: I believe the lines `57-64` could be simplified with Matchers:
```scala
chargingPlugInEvents.get() should be > 0
chargingPlugOutEvents.get() should be > 0
```
This way description is not needed

JustinPihony: fair enough - then the line number should cover which event

JustinPihony:
test now!

PR #2936, opened 2020-08-25
fix bug in secondary activities

zneedell:
Hopefully this fixes the problem with disappearing work activities

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2936)


nikolayilyin (approved):
the fix works well

PR #2938, opened 2020-08-25
added tests to ModeChoiceMultinomialLogit

carloscaldas:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2938)

### DIFF
--- a/src/test/scala/beam/agentsim/agents/choice/mode/ModeChoiceMultinomialLogitSpec.scala
+++ b/src/test/scala/beam/agentsim/agents/choice/mode/ModeChoiceMultinomialLogitSpec.scala
@@ -19,22 +19,27 @@ class ModeChoiceMultinomialLogitSpec extends WordSpec {
         val value1 = 11.4D
         val value2 = 12.3D
         val path = randomPath(travelTimes = 10D, value1, value2)
-        val data = BikeLanesData(scaleFactorFromConfig = 1D, bikeLanesLinkIds = randomIntList.toSet)
+        val bikeLaneLinkIds = randomIntList.toSet
+        val data = BikeLanesData(scaleFactorFromConfig = 1D, bikeLanesLinkIds = bikeLaneLinkIds)
         val adjustment = new BikeLanesAdjustment(data)
 
         val result: Double = ModeChoiceMultinomialLogit.pathScaledForWaiting(path, adjustment)
 
-        assert(result === value1 + value2)
+        val clue = s"Input path: [$path]; bikeLaneLinkIds: [$bikeLaneLinkIds]"
+        assert(result === value1 + value2, clue)
       }
 
       "be zero when path has zero or only one element" in {
         val path = if (Random.nextBoolean()) randomPath() else randomPath(travelTimes = 10D)
-        val data = BikeLanesData(scaleFactorFromConfig = Random.nextDouble(), bikeLanesLinkIds = randomIntList.toSet)
+        val scaleFactor = Random.nextDouble()
+        val bikeLaneLinkIds = randomIntList.toSet
+        val data = BikeLanesData(scaleFactorFromConfig = scaleFactor, bikeLanesLinkIds = bikeLaneLinkIds)
         val adjustment = new BikeLanesAdjustment(data)
 
         val result: Double = ModeChoiceMultinomialLogit.pathScaledForWaiting(path, adjustment)
 
-        assert(result === 0)
+        val clue = s"Input path: [$path]; scaleFactor: [$scaleFactor]; bikeLaneLinkIds: [$bikeLaneLinkIds]"
+        assert(result === 0, clue)
       }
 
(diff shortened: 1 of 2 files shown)


carloscaldas:
Test now!

carloscaldas:
Test!

carloscaldas:
Test now!

carloscaldas:
Test!

src/test/scala/beam/agentsim/agents/choice/mode/ModeChoiceMultinomialLogitSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+
+        assert(result === value1 * 1 / data.scaleFactorFromConfig + value2)
+      }
+
+    }
+  }
+
+}
+
+object ModeChoiceMultinomialLogitSpec {
```
REASY, line 63: nitpick: no seed number for the test which means that every time I run it the input will change. Isn't it against the general TDD concept? I'm fine to use them in Fuzzy testing, but it is not the case

carloscaldas: TDD is the process (is not against randomization). Introducing limited randomization help to figure corner cases and increase the readability since makes clear what we are testing. Random data means "I don't care about those values" or "can be any value that belongs to this domain"

REASY: Without providing seed number (which is the only thing which will help you to find a corner cases) your test is not reproducible - if it fails I have no idea what was an input which causes it to fail. And I don't see currently any extra code to say what was the whole input.

Regarding your input, once in a while it will fail because it will generate random number which can be Int.MaxValue, and the sum of two Int.MaxValue is -2 in Scala.

carloscaldas: > Regarding your provided example with anyPositiveNumber, once in a while it will fail because it will generate a random number which can be Int.MaxValue, and the sum of two Int.MaxValue is -2 in Scala.

image

This is the benefit I talked about

REASY: @carloscaldas but how will I figure out if something is wrong, I'll just get an assertion and have no idea of the whole input. TDD is traditionally used with deterministic algorithms when there is a known input and one expected result. Good unit tests should be reproducible and independent from external factors such as the environment (current time, random number generator and so on) or running order. Introducing randomness doesn't help me to understand whether the input is wrong or the code. Again, if it is fuzzy testing, I'm fine, but I don't think the current implementation is fuzzy testing.

JustinPihony: So many thoughts:

* With regards to reproducibility - I agree that this should be reproducible. And the way to do that is by outputting the random values so that it **can** then be reproduced. Often you only need to output those random values on failure, though. 
* With regards to randomness - that should not be part of the tests that are run every commit. The randomness can make flaky tests and shakiness with the tests overall. It can be debated whether they should be part of the final PR review is a good question, though. The two options would be to have it as a daily standalone test to make sure develop hasn't regressed and/or find new bugs -- or to have it as a second requirement to a PR which would only need run when you feel the PR is ready for final review.
* If this is a one off, then that's fine, but we could also look into using Scalacheck - which has built in utilities for really good randomness - https://medium.com/se-notes-by-alexey-novakov/generate-random-data-with-scalacheck-a89ba699fe3d (and it outputs the random factors by default iirc)

Ultimately, bullet number 1 needs done for this code imho.

src/test/scala/beam/agentsim/agents/choice/mode/ModeChoiceMultinomialLogitSpec.scala
```diff
@@ -0,0 +1,121 @@
...
+        val path = randomPath(travelTimes = 10D, value1, value2)
+        val data = BikeLanesData(scaleFactorFromConfig = 1D, bikeLanesLinkIds = randomIntList.toSet)
+        val adjustment = new BikeLanesAdjustment(data)
+
+        val result: Double = ModeChoiceMultinomialLogit.pathScaledForWaiting(path, adjustment)
+
+        assert(result === value1 + value2)
+      }
+
+      "be zero when path has zero or only one element" in {
```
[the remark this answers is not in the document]
JustinPihony: This test is also not reproducible and needs to print the factor - as well as should be split into two tests as already mentioned (again for reproducibility)

REASY:
test!

carloscaldas:
Test now!

JustinPihony (approved):

PR #2955, opened 2020-08-29
Beautify mode chooses

SHaaD94:


SHaaD94:
Test!

SHaaD94:
Test!

SHaaD94:
Test!

SHaaD94:
Test!

SHaaD94:
Test now!

SHaaD94:
test now!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -57,121 +57,42 @@ trait ChoosesMode {
...
-                  listOfResponses
-                    .collect {
-                      case MobilityStatusResponse(vehicles) =>
-                        vehicles
-                    }
-                    .flatten
-                    .toVector
+            .map { listOfResponses =>
+              MobilityStatusResponse(
+                listOfResponses.flatMap { case MobilityStatusResponse(vehicles) => vehicles }.toVector
```
REASY, line 93: I don't think this is equal transformation. Imagine you have the following code:
```scala
val seq: Seq[Any] = List[Any]("Hello", 1, 4.0)
```
Then if we use collect, it will work as expected:
```scala
seq.collect { case s: String => s }
```

But if I use map/flatMap, it will throw `scala.MatchError: 1 (of class java.lang.Integer)`:
```scala
seq.map { case s: String => s }
```
image

REASY (changes requested):

SHaaD94:
test now!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -601,8 +514,7 @@ trait ChoosesMode {
     driveTransitTrip: Option[EmbodiedBeamTrip],
     rideHail2TransitResult: Option[RideHailResponse]
   ): Boolean = {
-    driveTransitTrip.isDefined && driveTransitTrip.get.legs
-      .exists(leg => beamScenario.rideHailTransitModes.contains(leg.beamLeg.mode)) &&
+    driveTransitTrip.exists(_.legs.exists(leg => beamScenario.rideHailTransitModes.contains(leg.beamLeg.mode))) &&
```
JustinPihony, line 517: I do not believe this to be better performance nor does it add much to readability. I guess it doesn't matter in the long run here, but I'd personally prefer to not change this one.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -767,9 +678,8 @@ trait ChoosesMode {
 
     val remainingTripData: Option[ParkingMNL.RemainingTripData] =
       stateData match {
-        case _: BasePersonData =>
-          calculateRemainingTripData(stateData.asInstanceOf[BasePersonData])
-        case _ => None
+        case basePersonData: BasePersonData => calculateRemainingTripData(basePersonData)
```
JustinPihony, line 681: Without running this I'm unsure if erasure plays into this and will result in a warning? But we should have warnings as errors still...so it's probably fine.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
SHaaD94: fixed

JustinPihony:
@SHaaD94 Where does this stand?

PR #2978, opened 2020-09-08
#2814 RealizedModeChoiceWriter

### LINKED ISSUE #2814: New York: Create realized mode choices event file
In order to make an analysis easier, instead of reading events.csv.gz (which is pretty large and we don't write it on every iteration), let's create another listener which will write only realized mode choice events.

aaltergot:
Resolves #2814 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2978)

### DIFF
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -663,6 +663,8 @@ beam.debug {
 
     vmInformation.createGCClassHistogram = "boolean | false"
     writeModeChoiceAlternatives = "boolean | false"
+
+    writeRealizedModeChoiceFile = "boolean | false"
 }
 
 beam.logger.keepConsoleAppenderOn = "boolean | true"
--- a/src/main/scala/beam/analysis/RealizedModeChoiceWriter.scala
+++ b/src/main/scala/beam/analysis/RealizedModeChoiceWriter.scala
@@ -75,10 +75,10 @@ class RealizedModeChoiceWriter(beamServices: BeamServices)
 
   override def notifyIterationEnds(event: IterationEndsEvent): Unit = {
     personIdPrevMCE.values.foreach {
-        case Some(modeChoiceEvent @ _) =>
-          csvWriter.handleEvent(modeChoiceEvent)
+      case Some(modeChoiceEvent @ _) =>
+        csvWriter.handleEvent(modeChoiceEvent)
 
-        case _ =>
+      case _ =>
     }
 
     csvWriter.closeFile()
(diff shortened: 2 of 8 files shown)


aaltergot:
test now!

src/main/scala/beam/analysis/RealizedModeChoiceWriter.scala
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  override def notifyIterationStarts(event: IterationStartsEvent): Unit = {
+    csvFilePath = beamServices.matsimServices.getControlerIO.getIterationFilename(
+      event.getIteration,
+      "realizedModeChoice.csv.gz"
+    )
+
+    // writeHeader is invoked in constructor
+    csvWriter = new BeamEventsWriterCSV(
```
rajnikantsh, line 64: possibility  for resource leak , close csv writer for any exceptions

rajnikantsh (changes requested):

src/main/scala/beam/analysis/RealizedModeChoiceWriter.scala
JustinPihony: This is the last thing done in this method - and closing it on exceptions outside of this won't make a difference. So, the only place needed would be `notifyIterationEnds` as far as I can tell

src/main/resources/beam-template.conf
```diff
@@ -663,6 +663,8 @@ beam.debug {
 
     vmInformation.createGCClassHistogram = "boolean | false"
     writeModeChoiceAlternatives = "boolean | false"
+
+    writeRealizedModeChoiceFile = "boolean | false"
```
JustinPihony, line 667: Should there be anything added to the docs so that this isn't a hidden feature?

JustinPihony (approved):

src/main/resources/beam-template.conf
aaltergot: I haven't found any mentions of "beam.debug" config in docs. Did I miss it somewhere? (https://github.com/LBNL-UCB-STI/beam/blob/develop/docs/inputs.rst] looks like a correct place. Shall we create a task to document the whole section?

aaltergot: `beam-template.conf` is also a doc, comments may be added for details. Let's consider `writeRealizedModeChoiceFile ` naming descriptive enough.

JustinPihony (approved):

JustinPihony:
@rajnikantsh Is this good with you now?

rajnikantsh:
> @rajnikantsh Is this good with you now?

Yep, LGTM

rajnikantsh (approved):

REASY (approved):

PR #2984, opened 2020-09-10
Bug: ODSkimmer writes full skim with incorrect order of columns

REASY:
- `generalizedTime` was written twice (the second time it should have been `generalizedCost`), fixed it. Now we write `generalizedCost`
- Fix the order of columns `count` and `energy`
- Added the value for `iterations` column

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2984)


REASY:
test!

nikolayilyin (approved):

PR #2992, opened 2020-09-17
Fix TryGet and EitherGet error

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/2992)

### DIFF
--- a/src/main/scala/beam/router/BeamRouter.scala
+++ b/src/main/scala/beam/router/BeamRouter.scala
@@ -396,13 +396,7 @@ class BeamRouter(
   }
 
   private def resolveAddressBlocking(addr: Address, d: FiniteDuration = 60.seconds): Option[ActorRef] = {
-    Try(Await.result(resolveAddress(addr, d), d))
-      .recover {
-        case t: Throwable =>
-          log.error(t, "resolveAddressBlocking failed to resolve '{}' in {}: {}", addr, d, t.getMessage)
-          None
-      }
-      .getOrElse(None)
+    Await.result(resolveAddress(addr, d), d)
   }
 
   private def resolveAddress(addr: Address, duration: FiniteDuration = 60.seconds): Future[Option[ActorRef]] = {
--- a/src/main/scala/beam/router/BeamRouter.scala
+++ b/src/main/scala/beam/router/BeamRouter.scala
@@ -396,11 +396,13 @@ class BeamRouter(
   }
 
   private def resolveAddressBlocking(addr: Address, d: FiniteDuration = 60.seconds): Option[ActorRef] = {
-    Try(Await.result(resolveAddress(addr, d), d)).recover {
-      case t: Throwable =>
-        log.error(t, "resolveAddressBlocking failed to resolve '{}' in {}: {}", addr, d, t.getMessage)
-        None
-    }.get
+    Try(Await.result(resolveAddress(addr, d), d))
+      .recover {
+        case t: Throwable =>
+          log.error(t, "resolveAddressBlocking failed to resolve '{}' in {}: {}", addr, d, t.getMessage)
+          None
+      }
+      .getOrElse(None)
   }
 
   private def resolveAddress(addr: Address, duration: FiniteDuration = 60.seconds): Future[Option[ActorRef]] = {
--- a/src/main/scala/beam/router/r5/NetworkCoordinator.scala
+++ b/src/main/scala/beam/router/r5/NetworkCoordinator.scala
@@ -101,7 +101,6 @@ trait NetworkCoordinator extends LazyLogging {
         // Needed because R5 closes DB on write
         transportNetwork = KryoNetworkSerializer.read(path.toFile)
       }
-      .get
   }
 
   def overwriteLinkParams(
--- a/src/main/scala/beam/router/r5/NetworkCoordinator.scala
+++ b/src/main/scala/beam/router/r5/NetworkCoordinator.scala
@@ -65,42 +65,46 @@ trait NetworkCoordinator extends LazyLogging {
   def loadNetwork(): Unit = {
     val GRAPH_FILE = "/network.dat"
     val graphPath = Paths.get(beamConfig.beam.routing.r5.directory, GRAPH_FILE)
-    FileUtils
-      .readOrCreateFile(graphPath) { path =>
-        logger.info(
-          s"Initializing router by reading network from: ${path.toAbsolutePath}"
-        )
-        transportNetwork = KryoNetworkSerializer.read(path.toFile)
+    try {
+      FileUtils
+        .readOrCreateFile(graphPath) { path =>
+          logger.info(
+            s"Initializing router by reading network from: ${path.toAbsolutePath}"
+          )
+          transportNetwork = KryoNetworkSerializer.read(path.toFile)
 
-        network = FileUtils
-          .readOrCreateFile(Paths.get(beamConfig.matsim.modules.network.inputNetworkFile)) { _ =>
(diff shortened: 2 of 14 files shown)


src/main/scala/beam/router/BeamRouter.scala
```diff
@@ -396,11 +396,13 @@ class BeamRouter(
   }
 
   private def resolveAddressBlocking(addr: Address, d: FiniteDuration = 60.seconds): Option[ActorRef] = {
-    Try(Await.result(resolveAddress(addr, d), d)).recover {
-      case t: Throwable =>
-        log.error(t, "resolveAddressBlocking failed to resolve '{}' in {}: {}", addr, d, t.getMessage)
-        None
-    }.get
+    Try(Await.result(resolveAddress(addr, d), d))
```
JustinPihony, line 399: Should this one even be wrapped in a Try? It already has a recover, and any errors that come out of this are likely bad enough that they should propagate out.

src/main/scala/beam/router/r5/NetworkCoordinator.scala
```diff
@@ -101,7 +101,6 @@ trait NetworkCoordinator extends LazyLogging {
         // Needed because R5 closes DB on write
         transportNetwork = KryoNetworkSerializer.read(path.toFile)
       }
-      .get
```
JustinPihony, line 104: Rather than removing, should we check and log/handle the failure scenario?

src/main/scala/beam/router/r5/R5Wrapper.scala
```diff
@@ -449,6 +449,7 @@ class R5Wrapper(workerParams: R5Parameters, travelTime: TravelTime, travelTimeNo
       }
       streetRouter.profileRequest = profileRequest
       streetRouter.streetMode = toR5StreetMode(vehicle.mode)
+      val legMode = vehicle.mode.r5Mode.get.left.getOrElse(LegMode.valueOf(""))
```
JustinPihony, line 452: Should we map over the `r5Mode` if you are trying to handle `get`'s better?

```suggestion
      val legMode = vehicle.mode.r5Mode.flatmap(_.left.toOption).getOrElse(LegMode.valueOf(""))
```

src/main/scala/beam/router/r5/R5Wrapper.scala
```diff
@@ -752,7 +754,8 @@ class R5Wrapper(workerParams: R5Parameters, travelTime: TravelTime, travelTimeNo
 
           if (itinerary.connection.egress != null) {
             val egress = option.egress.get(itinerary.connection.egress)
-            val vehicle = egressVehicles.find(v => v.mode.r5Mode.get.left.get == egress.mode).get
+            val vehicle =
```
JustinPihony, line 757: Same code suggestion here - also this ends in a `get` - does that need handled?

src/main/scala/beam/router/skim/CsvSkimReader.scala
```diff
@@ -39,17 +39,19 @@ class CsvSkimReader(
...
-        logger.warn(s"Could not load warmStart skim from '$aggregatedSkimsFilePath'", ex)
-        Map.empty[AbstractSkimmerKey, AbstractSkimmerInternal]
-    }.get
+    tryReadSkims(reader)
+      .recover {
+        case ex: Throwable =>
+          logger.warn(s"Could not load warmStart skim from '$aggregatedSkimsFilePath'", ex)
+          Map.empty[AbstractSkimmerKey, AbstractSkimmerInternal]
+      }
+      .getOrElse(Map.empty[AbstractSkimmerKey, AbstractSkimmerInternal])
```
JustinPihony, line 54: Same as the other spot, should we let this propagate out since it is already being primarily handled?

JustinPihony (changes requested):
I want to put this out there that technically yes we mitigate a `None.get`, but may hide problems in doing so. Admittedly if the default of the type is not acceptable then it should maybe fail at the point of return from whatever method it is generate. But pragmatism at play - we need to make sure to not add getOrElse just to fix a compiler issue - we need to make sure business logic is considered.

The above was commented before doing the full review - marking as request for change as there are some items that seemed like they needed a bit more change - as commented.

src/main/scala/beam/router/BeamRouter.scala
rajnikantsh: Yes, We dont need Try to Wrap this because its already has a recover

src/main/scala/beam/router/r5/NetworkCoordinator.scala
rajnikantsh: use try/catch for handle/log error

src/main/scala/beam/router/r5/R5Wrapper.scala
rajnikantsh: applied this fix

src/main/scala/beam/router/r5/R5Wrapper.scala
rajnikantsh: last get is coming from Option.get but in this PR we are handing EItherGet and TryGet. will take a look how we can handle in next PR based on Option.get

src/main/scala/beam/router/r5/R5Wrapper.scala
```diff
@@ -305,8 +305,8 @@ class R5Wrapper(workerParams: R5Parameters, travelTime: TravelTime, travelTimeNo
         geo.utm2Wgs(request.destinationUTM),
         10E3
       )
-      val directMode = vehicle.mode.r5Mode.get.left.get
-      val accessMode = vehicle.mode.r5Mode.get.left.get
+      val directMode = vehicle.mode.r5Mode.flatMap(_.left.toOption).getOrElse(LegMode.valueOf(""))
```
JustinPihony, line 308: I just realized that these are exactly the same - why have two vals that are the same thing?

rajnikantsh: this was just for better readability. we are passing directMode and accessMode in RoutingRequest so here we are preparing these. Now make changes in this and reduce to one variable and use parameter in RoutingRequest for identifying directMode and accessMode

JustinPihony: I like that much better - thanks

src/main/scala/beam/router/skim/CsvSkimReader.scala
```diff
@@ -39,9 +39,11 @@ class CsvSkimReader(
             logger.warn(s"Could not load warmStart skim from '$aggregatedSkimsFilePath': ${ex.getMessage}")
             Map.empty[AbstractSkimmerKey, AbstractSkimmerInternal]
         }
-        .getOrElse(Map.empty[AbstractSkimmerKey, AbstractSkimmerInternal])
+        .getOrElse({
```
JustinPihony, line 42: Now it ends up with *really* duplicate code. Here is what I was suggesting earlier with trapping the main part in a variable

```
val skimMapTry = Try{
  IOUtils.getBufferedReader(aggregatedSkimsFilePath)
}.flatMap(tryReadSkims)
skimMapTry match {
  case Success(skimMap) => skimMap
  case Failure(ex) =>
    logger.warn(s"Could not load warmStart skim from '$aggregatedSkimsFilePath': ${ex.getMessage}")
    Map.empty[AbstractSkimmerKey, AbstractSkimmerInternal]
}
```

rajnikantsh: thanks for snipppet . Applied this change

JustinPihony (approved):

PR #3002, opened 2020-09-22
Fix OptionGet issues : Iteration-1

rajnikantsh:


src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceMultinomialLogit.scala
```diff
@@ -169,19 +169,19 @@ class ModeChoiceMultinomialLogit(
...
+                alternatives,
+                altCostTimeTransfer,
+                altUtility,
+                chosenModeCostTime.head.index
+              )
             )
-          )
-        } else {
-          None
+          case None => None
```
carloscaldas, line 184: consider replacing pattern match with `time.map{ t => ... }`

src/main/scala/beam/agentsim/agents/household/FastHouseholdCAVScheduling.scala
```diff
@@ -433,8 +433,13 @@ object HouseholdTripsHelper {
...
                 )
-              if (firstPickupOfTheDay.isEmpty || firstPickupOfTheDay.get.baselineNonPooledTime > pickup.baselineNonPooledTime)
-                firstPickupOfTheDay = Some(pickup)
+              firstPickupOfTheDay match {
+                case Some(firstPickup) if firstPickup.baselineNonPooledTime > pickup.baselineNonPooledTime =>
+                  firstPickupOfTheDay = Some(pickup)
+                case None =>
+                  firstPickupOfTheDay = Some(pickup)
+                case _ =>
+              }
```
carloscaldas, line 442: Just a suggestion. No need to change
```Scala
firstPickupOfTheDay = firstPickupOfTheDay match {
  case Some(firstPickup) if firstPickup.baselineNonPooledTime > pickup.baselineNonPooledTime =>
    Some(pickup)
  case None =>
    Some(pickup)
  case _ =>
    firstPickupOfTheDay
}
```

src/main/scala/beam/cosim/helics/BeamFederate.scala
```diff
@@ -23,11 +23,14 @@ object BeamFederate {
   var beamFed = Option.empty[BeamFederate]
 
   def getInstance(beamServices: BeamServices): BeamFederate = this.synchronized {
-    if (beamFed.isEmpty) {
-      loadHelics
-      beamFed = Some(BeamFederate(beamServices))
+    beamFed match {
+      case Some(beamFederate) => beamFederate
+      case None =>
+        loadHelics
```
carloscaldas, line 29: is `loadHelics` a side-effect method? is case of yes consider adding parenthesis to it. Looks like unused variable

src/main/scala/beam/cosim/helics/BeamFederate.scala
```diff
@@ -23,11 +23,14 @@ object BeamFederate {
   var beamFed = Option.empty[BeamFederate]
 
   def getInstance(beamServices: BeamServices): BeamFederate = this.synchronized {
-    if (beamFed.isEmpty) {
-      loadHelics
-      beamFed = Some(BeamFederate(beamServices))
+    beamFed match {
+      case Some(beamFederate) => beamFederate
```
carloscaldas, line 27: this method looks like buggy. pls double check because `beamFed` is changed only in the second case

carloscaldas (changes requested):

src/main/scala/beam/cosim/helics/BeamFederate.scala
rajnikantsh: its not a method. Its a lazy val which is called only one time.

src/main/scala/beam/cosim/helics/BeamFederate.scala
rajnikantsh: its working like this: If beamFed is initialized then return its instance otherwise initialize it then reurn its instance. So second case is basically handle its initialization

src/main/scala/beam/agentsim/agents/choice/mode/ModeChoiceMultinomialLogit.scala
rajnikantsh: replaced pattern matching with map

src/main/scala/beam/utils/traveltime/FeatureEventHandler.scala
```diff
@@ -57,7 +57,7 @@ class FeatureEventHandler(
 
   def handleEvent(event: Event): Unit = {
     val attrib = event.getAttributes
-    val linkId = Option(attrib.get("link")).map(_.toInt).get
+    val linkId = attrib.get("link").toInt
```
JustinPihony, line 60: This breaks the logic. IIRC `get` can return a null. While at it I'm also wondering if the `toInt` should be wrapped in a Try? But I'm guessing we want an exception if a non-numeric is encountered?

JustinPihony (changes requested):
One nit and one more of a logic change than not

PR #3003, opened 2020-09-23
[Issue-2995] Removed 90 UnsafeTraversableMethods error msgs

### LINKED ISSUE #2995: Scapegoat - remove UnsafeTraversableMethods error messages
Related to #2985

to test: `/gradlew compileScala -Preport`

carloscaldas:
Closes #2995 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3003)


carloscaldas:
Test now!

carloscaldas:
Test now!

carloscaldas:
Test!

carloscaldas:
Test now!

carloscaldas:
Test!

src/main/scala/beam/router/r5/R5Wrapper.scala
```diff
@@ -305,18 +307,20 @@ class R5Wrapper(workerParams: R5Parameters, travelTime: TravelTime, travelTimeNo
         geo.utm2Wgs(request.destinationUTM),
         10E3
       )
-      val vehicleLegMode = vehicle.mode.r5Mode.flatMap(_.left.toOption).getOrElse(LegMode.valueOf(""))
+      val directMode = vehicle.mode.r5Mode.get.left.get
```
rajnikantsh, line 310: This change was done in last PR. We have to change becuase of TryGet error so now if we revert this again as original then we will encountered again TryGet error

src/main/scala/beam/router/r5/R5Wrapper.scala
```diff
@@ -376,9 +380,15 @@ class R5Wrapper(workerParams: R5Parameters, travelTime: TravelTime, travelTimeNo
...
+    val vehicleToInt = (streetVehicle: StreetVehicle) =>
+      maybeWalkToVehicle(streetVehicle)
+        .map(leg => leg.beamLeg.duration)
+        .getOrElse(0)
+
+    @SuppressWarnings(Array("UnsafeTraversableMethods"))
     val bestAccessVehiclesByR5Mode: Map[LegMode, StreetVehicle] = accessVehicles
-      .groupBy(_.mode.r5Mode.flatMap(_.left.toOption).getOrElse(LegMode.valueOf("")))
-      .mapValues(vehicles => vehicles.minBy(maybeWalkToVehicle(_).map(leg => leg.beamLeg.duration).getOrElse(0)))
+      .groupBy(_.mode.r5Mode.get.left.get)
```
rajnikantsh, line 390: Same with this. We will introduce TryGet error

rajnikantsh (changes requested):
Except changes in one class all looks good

src/main/scala/beam/router/r5/R5Wrapper.scala
carloscaldas: reverted

src/main/scala/beam/replanning/utilitybased/ChainBasedTourVehicleAllocator.scala
```diff
@@ -202,8 +206,10 @@ object ChainBasedTourVehicleAllocator {
     def apply(possibleVehicles: Vector[VehicleRecord], subtour: Subtour): SubtourRecord = {
       val trips = JavaConverters.collectionAsScalaIterable(subtour.getTrips)
       val startTime = Try {
-        trips.head.getOriginActivity.getStartTime
-      }.getOrElse(throw new RuntimeException(s"No endTime in ${trips.head.getOriginActivity}"))
+        trips.headOption.map(_.getOriginActivity.getStartTime).get
```
JustinPihony, line 209: Doesn't this trigger an OptionGet warning? How about:

```
val tripOriginActivityOption = trips.headOption.map(_.getOriginActivity)
val startTime = (
  Try{ tripOriginActivityOption.map(_.getStartTime) }
).toOption.flatten
.getOrElse(throw new RuntimeException(s"No startTime in $tripOriginActivityOption"))
```

This resolves all of the possible safety issues. I'm only unsure if `getOriginActivity` needs wrapped in the `Try`?

src/main/scala/beam/router/BeamRouter.scala
```diff
@@ -384,6 +384,9 @@ class BeamRouter(
       case None => //No matching id. No need to log since this is more for analysis
     }
 
+  // TODO: availableWorkers is a SET (not sortedSet).
```
JustinPihony, line 387: This seems to be a bug that should be issue-ized?

src/main/scala/beam/router/model/BeamLeg.scala
```diff
@@ -126,20 +126,20 @@ object BeamLeg {
...
-      var latestEndTime = legs.head.startTime - 1
-      var newLeg = legs.head
-      for (leg <- legs) yield {
-        if (leg.startTime < latestEndTime) {
-          newLeg = leg.updateStartTime(latestEndTime)
-        } else {
-          newLeg = leg
+    legs.headOption match {
+      case None => legs
+      case Some(legHead) =>
```
JustinPihony, line 131: Do we need to care it is the head - if so then I'd say we should go with headLeg or topLeg, firstLeg

src/main/scala/beam/router/model/BeamPath.scala
```diff
@@ -60,17 +60,18 @@ case class BeamPath(
     )
   }
 
+  @SuppressWarnings(Array("UnsafeTraversableMethods"))
   def linkAtTime(tick: Int): Int = {
     tick - startPoint.time match {
       case secondsAlongPath if secondsAlongPath <= 0 || linkIds.size <= 1 =>
+        // TODO: there is a likely bug here because linkIds.size can be 0(zero)
```
JustinPihony, line 67: Issue-ize? Although I don't see the issue per se

src/main/scala/beam/utils/gtfs/GTFSToShape.scala
```diff
@@ -53,9 +53,7 @@ object GTFSToShape extends LazyLogging {
 
       try {
         val agencies = GTFSReader.readAgencies(zipFile, zipFileName)
-        val defaultAgency =
-          if (agencies.size == 1) Some(agencies.head)
-          else None
+        val defaultAgency = if (agencies.size == 1) agencies.headOption else None
```
JustinPihony, line 56: This seems like a bug - should the test be > 0? In which case it is just headOption

src/main/scala/beam/utils/mapsapi/googleapi/GoogleAdapter.scala
```diff
@@ -91,9 +91,11 @@ class GoogleAdapter(apiKey: String, outputResponseToFile: Option[Path] = None, a
   }
 
   private def parseRoutes(jsRoutes: Seq[JsValue]): Seq[Route] = {
-    jsRoutes.map { route =>
-      val firstAndUniqueLeg = (route \ "legs").as[JsArray].value.head
-      parseRoute(firstAndUniqueLeg.as[JsObject])
+    jsRoutes.flatMap { route =>
+      (route \ "legs").as[JsArray].value.headOption match {
```
JustinPihony, line 95: Wouldn't a map be better here?

JustinPihony (approved):
Some minor items - although most are suppressions - which begs the question if we should use this setting at all?

src/main/scala/beam/router/r5/R5Wrapper.scala
carloscaldas: reverted

src/main/scala/beam/router/BeamRouter.scala
carloscaldas: I cannot answer this question

REASY (changes requested):
Blocking it until the sync with forked repo is done

src/main/scala/beam/router/model/BeamLeg.scala
carloscaldas: variable renamed

src/main/scala/beam/router/BeamRouter.scala
REASY: @carloscaldas are you suggesting instead to use some data structure which keeps the order, like `TreeSet` so there is expected behavior of getting the first element? Agree, in that case it will be clear how it works, otherwise someone needs to go check that Set is implemented as HashSet which uses, and `Set.head` is actually accessing iterator.

carloscaldas: @REASY there are 2 possibilities:
- the first element can be any element. in this case we just need to rename the method to make the behavior clear
- it really needs to be the first element. in this case we must make sure the datastructure is sortable

I don't know what is the expected behavior. <-- this is required to propose some solution

JustinPihony:
This is done, however same as the other PR you reverted I'm unclear if your change was pushed through or not - can you point to what we need to update against? @REASY

src/main/scala/beam/utils/mapsapi/googleapi/GoogleAdapter.scala
carloscaldas: Agree. I did the change as you suggested

src/main/scala/beam/utils/gtfs/GTFSToShape.scala
carloscaldas: In this case does not sounds like a bug to me. basically states that a `defaultAgency` will exists only in case there is only one agency. in case of 2 or more there is no default.

src/main/scala/beam/router/model/BeamPath.scala
carloscaldas: @JustinPihony the following is the whole condition:
```Scala
case secondsAlongPath if secondsAlongPath <= 0 || linkIds.size <= 1 =>
  linkIds.head
```
if `linkIds.size == 0` then `linkIds.head` will throw an exception. That is why looks like a bug to me

src/main/scala/beam/replanning/utilitybased/ChainBasedTourVehicleAllocator.scala
carloscaldas: The `Try` is redundant. After changes became as follows:
```Scala
val tripOriginActivityOption = trips.headOption.map(_.getOriginActivity)
val startTime = tripOriginActivityOption.map(_.getStartTime) 
  .getOrElse(throw new RuntimeException(s"No startTime in $tripOriginActivityOption"))
```

carloscaldas:
Test now!

carloscaldas:
Test!

JustinPihony:
@REASY https://github.com/LBNL-UCB-STI/beam/pull/3084 is the one that needs merged first, correct?

src/main/scala/beam/router/model/BeamPath.scala
JustinPihony: Sorry - I honestly didn't read the TODO itself - went into auto-pilot to ask a TODO to be issue-ized. That said, I looked at the blame and found [this](https://github.com/LBNL-UCB-STI/beam/commit/49899e973407383f50be779e1621aa3902197469#diff-161b4a7c3d656317ee6511bbf2212cfa249d2fb042e2e087b5af63d9de206f82). Upon review I am questioning a number of those changes. This definitely should be raised as a new issue as there are a lot of flaws - and should have tests to catch.

JustinPihony:
This was blocked by Rajnikant and Art still, I'll override as Art has said it is fine in Slack, and Rajnikant's issues have been addressed.

PR #3020, opened 2020-09-27
Refactored Tour class to encapsulate property and avoid IndexOfBounds…

### LINKED ISSUE #3000: Refactor Tour class
The class `Tour` has a lot of methods that assumes the variable `trips` is not empty (like using `head` operation). However it is not required at the constructor level. Is it possible to make some assumption? This should be double checked and the class refactored to avoid warnings and work in a safe mode

carloscaldas:
Closes #3000 

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3020)


carloscaldas:
Test!

carloscaldas:
Test now!

carloscaldas:
Test now!

src/main/scala/beam/agentsim/agents/planning/Tour.scala
```diff
@@ -1,28 +1,38 @@
...
-          if trip._1 == planElement || trip._1.activity == planElement || (trip._1.leg.isDefined && trip._1.leg.get == planElement))
-      yield trip._2).head
+    val indexes = for {
+      tripsWithIndex <- tripsInternal.zipWithIndex
+      if isTripOrActivityPlanElement(tripsWithIndex._1, planElement) || isTripLegPlanElement(
+        tripsWithIndex._1,
+        planElement
+      )
+    } yield tripsWithIndex._2
+    indexes.headOption.getOrElse(-1)
```
REASY, line 23: This is behavioral change. If before it would fail right there, now it will give -1 and then try to understand where it came from.

carloscaldas: @REASY changed as discussed

carloscaldas:
Test now!

REASY (approved):
LGTM, but please, postpone a while with the merge.

src/main/scala/beam/agentsim/agents/planning/Tour.scala
```diff
@@ -1,28 +1,37 @@
...
+class Tour(private var tripsInternal: Vector[Trip] = Vector()) extends PlanElement {
 
-  def originActivity(): Activity = {
-    trips.head.activity
-  }
-
-  def destActivity(): Activity = {
-    trips.reverse.head.activity
-  }
+  def trips: Seq[Trip] = tripsInternal
```
JustinPihony, line 8: Why was this done? Ultimately if this was done to control mutation outside of this then we may want to deep copy?

JustinPihony (approved):
LGTM with one question

src/main/scala/beam/agentsim/agents/planning/Tour.scala
carloscaldas: the idea was to make the class immutable

JustinPihony: I guess what I am saying is that it is not immutable due to this since the items in the Seq are changable:

```
scala> class Tour(var trips: Vector[Int])
defined class Tour

scala> case class Trip(x:Int, fOpt: Option[String], parentTour: Tour)
defined class Trip

scala> val test = Seq(Trip(1,None,new Tour(Vector(1))))
test: Seq[Trip] = List(Trip(1,None,Tour@3a388769))

scala> test.head.parentTour.trips
res3: Vector[Int] = Vector(1)

scala> test.head.parentTour.trips = Vector(0)
test.head.parentTour.trips: Vector[Int] = Vector(0)

scala> test.head.parentTour.trips
res4: Vector[Int] = Vector(0)
```

JustinPihony:
Can this be merged - my question is not a blocker, so in the vein of forward progress I ask the question of merging :)

src/main/scala/beam/agentsim/agents/planning/Tour.scala
carloscaldas: @JustinPihony I used not the most accurate word. I meant is not exposed as reassignable variable to class clients:

PR #3028, opened 2020-10-07
added ability to have multiple different runners

carloscaldas:


src/main/scala/beam/sim/BeamRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+import com.google.inject.Module
+import org.matsim.core.config.{Config => MatsimConfig}
+
+class BeamRunner(
+  val matsimConfig: MatsimConfig,
+  val outputDirectory: String,
+  val beamServices: BeamServices,
+  val injector: inject.Injector,
+  runMethod: () => Unit,
+  postRunner: BeamRunner => Any = _ => ()
```
JustinPihony, line 15: Nit: Would it be more informative to make this an Option?

src/main/scala/beam/sim/BeamRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+    postRunner = func
+  )
+
+  def run(): BeamRunner = {
+    runMethod()
+    postRunner(this)
+    this
+  }
+
+  def addModules(module: Module, others: Module*): BeamRunner = {
```
JustinPihony, line 33: Is this used? Or is it from a previous iteration.

src/main/scala/beam/sim/BeamRunner.scala
```diff
@@ -0,0 +1,37 @@
...
+    outputDirectory = outputDirectory,
+    beamServices = beamServices,
+    injector = injector,
+    runMethod = runMethod,
+    postRunner = func
+  )
+
+  def run(): BeamRunner = {
+    runMethod()
+    postRunner(this)
```
JustinPihony, line 29: This seems very cyclic. Is there a post run hook or could we make the runMethod go off of a Future?

src/main/scala/beam/sim/RunBeamVersion2.scala
```diff
@@ -0,0 +1,20 @@
...
+object RunBeamVersion2 extends BeamHelper {
+  val logbackConfigFile: Option[String] = Option(System.getProperty(ContextInitializer.CONFIG_FILE_PROPERTY))
+  if (logbackConfigFile.isEmpty)
+    System.setProperty(ContextInitializer.CONFIG_FILE_PROPERTY, "logback.xml")
+
+  def main(args: Array[String]): Unit = {
+
+    print(beamAsciiArt)
+    val runner = buildRunner(args)
+    //    runner.addListener(myListener)
```
JustinPihony, line 14: We should be able to remove this.

JustinPihony (commented):
A couple questions I'd like addressed (not worth blocking, but also not approving)

PR #3030, opened 2020-10-09
AWS bucket listing

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3030)

### DIFF
--- a/aws/src/main/python/s3BucketList/lambda_function.py
+++ b/aws/src/main/python/s3BucketList/lambda_function.py
@@ -1,0 +1,73 @@
+import os
+import json
+import boto3
+import datetime
+
+initscript = (('''#cloud-config
+runcmd:
+  - echo "-------------------checking bucket size----------------------"
+  - sudo dpkg --configure -a
+  - sudo dpkg --remove --force-remove-reinstreq unattended-upgrades
+  - sudo apt-get install unattended-upgrades
+  - sudo dpkg --configure -a
+  - sudo apt update
+  - buckets=$(aws s3 ls | awk '{print $3}')
+  - for bucket in $buckets; do
+  - lines=$(aws s3 ls --region=$REGION s3://$bucket --recursive --human-readable --summarize)
+  - for line in $lines; do
+  - echo $line | grep -qE "/$"
+  - if [ $? = 0 ]; then 
+  - token_count=$(echo $line | awk -F'/' '{ print NF }')
+  - if [ $token_count -eq 2 -o $token_count -eq 3 ]; then
+  - result=$(aws s3 ls --region=$REGION s3://$bucket/$line --recursive | awk 'BEGIN {total=0}{total+=$3}END{print total/1024/1024}')
+  - echo $bucket/$line, $result >> "result.csv"
+  - fi
+  - fi 
+  - done
+  - done
+  - sort -o s3-result-out.csv -n -t , -k 2 -r result.csv
+  - echo "folder, size(MB)" > output.csv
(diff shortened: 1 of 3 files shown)


aws/src/main/python/s3BucketList/lambda_function.py
```diff
@@ -0,0 +1,73 @@
...
+                            } ])
+    return res['Instances'][0]['InstanceId']
+
+def lambda_handler(event, context):
+    region = event.get('region', DEFAULT_REGION)
+    size = event.get('size', 100)
+    shutdown_behaviour = 'terminate'
+    shutdown_wait = "10"
+    init_ec2(region)
+    script = initscript.replace("$N", size).replace('$REGION', region).replace('$SHUTDOWN_WAIT', shutdown_wait)
```
JustinPihony, line 70: You have to turn the size into a string explicitly and use single quotes I found:

```
script = initscript.replace('$N', str(size)).replace('$REGION', region).replace('$SHUTDOWN_WAIT', shutdown_wait)
```

JustinPihony (commented):
A couple of fixes needed before it can run (already applied to the lambda since that is how I was testing). That said, any idea of how long this should be expected to take? I had it running for an hour with no output.

JustinPihony:
My attempt to run this ended in:

> [Errno 32] Broken pipe
> Exception ignored in: <_io.TextIOWrapper name='<stdout>' mode='w' encoding='UTF-8'>

So, I'm wondering if this needs to be optimized and run in a different way?

aws/src/main/python/s3BucketList/lambda_function.py
```diff
@@ -0,0 +1,73 @@
...
+import datetime
+
+initscript = (('''#cloud-config
+runcmd:
+  - echo "-------------------checking bucket size----------------------"
+  - sudo dpkg --configure -a
+  - sudo dpkg --remove --force-remove-reinstreq unattended-upgrades
+  - sudo apt-get install unattended-upgrades
+  - sudo dpkg --configure -a
+  - sudo apt update
```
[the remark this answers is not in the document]
rajnikantsh: applied this change

aws/src/main/python/s3BucketList/lambda_function.py
rajnikantsh: applied this change

rajnikantsh:
> A couple of fixes needed before it can run (already applied to the lambda since that is how I was testing). That said, any idea of how long this should be expected to take? I had it running for an hour with no output.

@JustinPihony It will take around 5 hours to complete. Meanwhile you can check internediatory results in `/result.csv` . That other issue related to broken pipe has been fixed

JustinPihony (approved):
Looks good, but has it been tested? Also, has https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-inventory.html been looked into?

rajnikantsh:
> Looks good, but has it been tested? Also, has https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-inventory.html been looked into?

yes its tested one time and results is on location : https://beam-outputs.s3.us-east-2.amazonaws.com/s3-buckets-output.csv 

also I have reviewed https://docs.aws.amazon.com/AmazonS3/latest/dev/storage-inventory.html . It can be used in our case if we can identify that these bucket/folder we have to monitor. In current solution we dont need to know bucket/folder name. It will list top N folder by size

JustinPihony:
test!

PR #3060, opened 2020-10-22
Fix others and transit_walk in test

rajnikantsh:
While creating new plan elements we were using matsim mode. Now we have change this with beam mode

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3060)

### DIFF
--- a/src/main/scala/beam/replanning/ReplanningUtil.scala
+++ b/src/main/scala/beam/replanning/ReplanningUtil.scala
@@ -103,7 +103,7 @@ object ReplanningUtil {
     for (i <- 0 until originalPlan.getPlanElements.size() - 1) {
       newPlan.getPlanElements.add(originalPlan.getPlanElements.get(i))
       if (tripsLength > i) {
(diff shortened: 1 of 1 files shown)


REASY:
test!

src/main/scala/beam/replanning/ReplanningUtil.scala
```diff
@@ -103,7 +103,7 @@ object ReplanningUtil {
     for (i <- 0 until originalPlan.getPlanElements.size() - 1) {
       newPlan.getPlanElements.add(originalPlan.getPlanElements.get(i))
       if (tripsLength > i) {
-        val newLeg = PopulationUtils.createLeg(trips(i).tripClassifier.matsimMode)
+        val newLeg = PopulationUtils.createLeg(trips(i).tripClassifier.value)
```
REASY, line 106: According to the following code: https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala#L245 we create a non-matsim mode. So yes, it's good to go

REASY (approved):

PR #3077, opened 2020-11-13
pickups and dropoffs are affecting link travel time

nikolayilyin:
There are two logical parts.
Part I - collects all person enters\leaves vehicle events and path traversal events for CAV and RH vehicles
Part II - based on data from part I - changes link travel time for physSim if there are pickUps\dropOffs on a current link for simulation time not too far from pickUp/dropOff time.

require tests:

- [x] pickups\dropoffs increase average link travel time
- [x] CACC decrease impact of pickups\dropoffs
- [x] modified JDEQSim with pickups\dropoffs disabled and CACC disabled should return almost the same travel time as original JDEQSim (without CACC and pickups\dropoffs)
- [x] BPR and ParBPR should return the same link travel time with and without pickups/dropoffs

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3077)


nikolayilyin:
test!

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
```diff
@@ -0,0 +1,270 @@
...
+  }
+
+  override def handleEvent(event: Event): Unit = {
+    def notIgnored(vehicleId: Id[Vehicle]): Boolean = {
+      val isIgnored = vehiclesToIgnore.contains(vehicleId)
+      !isIgnored
+    }
+
+    event match {
+      case pte: PathTraversalEvent if pteIsSelected(pte) && notIgnored(pte.vehicleId) =>
```
JustinPihony, line 69: As far as I can tell, ignored is really just where pte is NOT selected? If so, then this additional check only adds a jump that is not needed. If you switched the order then it might save a call to pteIsSelected if we feel it is less performant (debatable). I would argue the notIgnored can be dropped altogether in this condition.

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
```diff
@@ -0,0 +1,270 @@
...
+
+  override def toString: String = s"PickUps: $pickUpTimeToLink DropOffs: $dropOffTimeToLink"
+}
+
+class TimeToValueCollection() {
+  val times: mutable.ListBuffer[Double] = mutable.ListBuffer.empty
+  val values: mutable.ListBuffer[Int] = mutable.ListBuffer.empty
+
+  // I assume time will be added already in sorted order
+  def addTimeToValue(time: Double, value: Int): Unit = {
```
JustinPihony, line 202: I'm not sure this can be tested for...can we really guarantee this over time? Should we use a sorted list to guarantee?

src/test/scala/beam/analysis/physsim/PickUpDropOffCollectorTest.scala
```diff
@@ -0,0 +1,219 @@
...
+        val linkPickUps: Int = pickUpDropOff.timeToPickUps.values.sum
+        val linkDropOffs: Int = pickUpDropOff.timeToDropOffs.values.sum
+
+        (pickUps + linkPickUps, dropOffs + linkDropOffs)
+    }
+
+    pickUps shouldBe (19 + 20)
+    dropOffs shouldBe (19 + 20)
+
+    /*
```
JustinPihony, line 62: Can this be dropped - is there a reason to keep commented out code?

JustinPihony (commented):
Mostly clarifying comments

src/test/scala/beam/analysis/physsim/PickUpDropOffCollectorTest.scala
nikolayilyin: this is a python code I used to calculate values from the events file, I am not sure if it will be better to put this piece of code elsewhere because it is only needed to obtain values to generate test code.

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
```diff
@@ -0,0 +1,270 @@
...
+import org.matsim.api.core.v01.events.{Event, PersonEntersVehicleEvent, PersonLeavesVehicleEvent}
+import org.matsim.core.controler.events.{IterationEndsEvent, IterationStartsEvent}
+import org.matsim.core.controler.listener.{IterationEndsListener, IterationStartsListener}
+import org.matsim.core.events.handler.BasicEventHandler
+import org.matsim.vehicles.Vehicle
+
+import scala.annotation.tailrec
+import scala.collection.mutable
+
+class PickUpDropOffCollector(vehicleTypes: Map[Id[BeamVehicleType], BeamVehicleType])
```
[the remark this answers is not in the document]
nikolayilyin: This is an interesting idea, could you, please, point me to something similar in beam codebase or elaborate more about how to change this collector?

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
nikolayilyin: there are two separate functions, one is to check if an event is what we are looking for, another is to check if a vehicle should be ignored. 
I have renamed them to not be confusing anymore.

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
nikolayilyin: ok, I wrote the code in case time is not added in sorted order

nikolayilyin:
test!

src/test/scala/beam/analysis/physsim/PickUpDropOffCollectorTest.scala
JustinPihony: I am fine with the comment given what you said

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
JustinPihony: The naming wasn't the issue, but instead the logic itself. It's more of an optimization on performance, so maybe readability is more important and I'll resolve  - feel free to change or not :)

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
```diff
@@ -0,0 +1,277 @@
...
+    }
+  }
+
+  override def handleEvent(event: Event): Unit = {
+    def vehicleNotIgnored(vehicleId: Id[Vehicle]): Boolean = {
+      val isIgnored = vehiclesToIgnore.contains(vehicleId)
+      !isIgnored
+    }
+
+    event match {
```
REASY, line 68: Repositioning of ride-hail fleet (controlled by `beam.agentsim.agents.rideHail.repositioningManager.name`) will cause them to be driving around the map which will create `PathTraversalEvent` every time when ride-hail vehicle repositions. Shall it be specifically handled because there will no no customer in that vehicle?

REASY (dismissed):

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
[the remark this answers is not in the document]
nikolayilyin: thanks! 
I will think about it

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
nikolayilyin: I do not think that empty RH vehicles should be handled here, at least for now.

JustinPihony:
Per Nikolay:

> there are still proper tests to write

alex-vv:
test!

alex-vv:
test!

src/main/scala/beam/physsim/PickUpDropOffCollector.scala
```diff
@@ -0,0 +1,277 @@
...
+  val times: mutable.ListBuffer[Double] = mutable.ListBuffer.empty
+  val values: mutable.ListBuffer[Int] = mutable.ListBuffer.empty
+
+  def addTimeToValue(time: Double, value: Int): Unit = {
+    if (times.isEmpty || times.last < time) {
+      times.append(time)
+      values.append(value)
+    } else {
+      var idx = times.length - 1
+      while (idx > 0 && times(idx) > time) { idx -= 1 }
```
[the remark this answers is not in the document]
alex-vv: it was changed to `ArrayBuffer` so resolving this

JustinPihony:
Tests were fixed, but no new tests were added - per my last comment from @nikolayilyin it sounded like new tests were needed. Otherwise I can approve

src/main/java/beam/physsim/jdeqsim/cacc/sim/Road.java
```diff
@@ -15,7 +16,9 @@
 
     private static final double INCREASE_TIMESTAMP = 0.0000000001D;
     private static RoadCapacityAdjustmentFunction roadCapacityAdjustmentFunction;
-    private final HashMap<Vehicle,Double> caccShareEncounteredByVehicle=new HashMap<>();
+    private static AdditionalLinkTravelTimeCalculationFunction additionalLinkTravelTimeCalculationFunction;
```
JustinPihony: This should be an `Optional` with the default being `Optional.empty`  and then use a `map` to calculate followed by `orElse` returning `0`

src/main/java/beam/physsim/AdditionalLinkTravelTimeCalculationFunctionMock.java
```diff
@@ -0,0 +1,10 @@
+package beam.physsim;
```
JustinPihony: This can be deleted given the Optional suggestion

src/main/java/beam/physsim/jdeqsim/cacc/roadcapacityadjustmentfunctions/RoadCapacityAdjustmentFunctionMock.java
```diff
@@ -0,0 +1,24 @@
+package beam.physsim.jdeqsim.cacc.roadcapacityadjustmentfunctions;
+
+import org.matsim.api.core.v01.network.Link;
+
+public class RoadCapacityAdjustmentFunctionMock implements RoadCapacityAdjustmentFunction {
```
JustinPihony: This should also be using an `Optional` over a mock

JustinPihony: Although this one has more functionality, so there is an argument for this, but I wouldn't call it a mock. Mocks are for testing, this is more of a Null Object. So if kept I would rename to `EmptyRoadCapacityAdjustmentFunction` or the like. And realistically it can be the default instead of having to be explicitly set.

src/main/scala/beam/agentsim/infrastructure/parking/TazToLinkLevelParkingApp.scala
```diff
@@ -32,6 +32,13 @@ object TazToLinkLevelParkingApp extends App with StrictLogging {
 
   val argsMap = parseArgs(args)
 
+  //  val argsMap = Map(
+  //    ("taz-parking", "test/input/sf-light/taz-parking.csv"),
+  //    ("network", "test/input/sf-light/r5/physsim-network.xml"),
+  //    ("taz-centers", "test/input/sf-light/taz-centers.csv"),
```
JustinPihony: This can be removed, correct?

JustinPihony (approved):
Couple of comments not worth blocking, but still suggested

nikolayilyin:
test!

nikolayilyin:
test!

nikolayilyin:
looks like test BeamWarmStartSpec works fine locally .. another try.

test!

nikolayilyin:
test!

PR #3080, opened 2020-11-30
NYC

REASY:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3080)


JustinPihony:
Is this the same as https://github.com/LBNL-UCB-STI/beam/pull/3109 or maybe since that is labeled as code only? Basically, is this still needed?

PR #3085, opened 2020-12-14
add secure download from s3

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3085)


JustinPihony (approved):
The logic is fine, but I question the naming. s3 is not inherently secure, so I think we should embrace the s3 portion, but no need to say it is secure.

rajnikantsh:
> The logic is fine, but I question the naming. s3 is not inherently secure, so I think we should embrace the s3 portion, but no need to say it is secure.

change secure to s3

JustinPihony:
@rajnikantsh Barring https://beam--team.slack.com/archives/CHFRVD3V1/p1609386673078200 can this be merged?

PR #3088, opened 2020-12-28
Tweak lambdas

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3088)

### DIFF
--- a/aws/src/main/python/cloudwatchAllInstances/lambda_function.py
+++ b/aws/src/main/python/cloudwatchAllInstances/lambda_function.py
@@ -121,7 +121,7 @@ def convert_to_ec2_instance_from(response_instance, region):
     logger.info('Convert to EC2Instance from ' + str(response_instance))
     email = None
     tags = response_instance.get('Tags')
-    if tags != None:
+    if tags:
         for tag in tags:
             if tag['Key'] == 'GitUserEmail':
                 email = tag['Value']
--- a/aws/src/main/python/cloudwatchAllInstances/lambda_function.py
+++ b/aws/src/main/python/cloudwatchAllInstances/lambda_function.py
@@ -6,9 +6,10 @@ import os
 import dateutil.relativedelta
 from datetime import datetime, timezone
 
-
 logger = logging.getLogger()
 logger.setLevel(logging.INFO)
+
+cloudtrail=boto3.client('cloudtrail')
 
 class EC2Instance:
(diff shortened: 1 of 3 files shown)


aws/src/main/python/cloudwatchAllInstances/lambda_function.py
```diff
@@ -103,39 +105,39 @@ def get_running_instance_for(region):
...
+    if tags != None:
         for tag in tags:
             if tag['Key'] == 'GitUserEmail':
                 email = tag['Value']
     instance_id = safe_get(response_instance, 'InstanceId')
-    #Email not found in tags, so now check instance in directly started instances
+    #Email not found in tags, so now check for directly started instances
     username=None
     if email is None:
-        cloudtrail = boto3.client('cloudtrail', region_name=region)
```
JustinPihony, line 124: @rajnikantsh This was a change from a copy in the server. I think we still need to do this from a region perspective, otherwise it uses what's in the environment - afaik. But this is from your change, so I wanted to check

rajnikantsh: yes @JustinPihony we need it region specific

JustinPihony:
Test!

JustinPihony:
Test!

JustinPihony:
I added a number of additional fixes that have been needed.

aws/src/main/python/cloudwatchAllInstances/lambda_function.py
```diff
@@ -113,29 +121,30 @@ def convert_to_ec2_instance_from(response_instance, region):
     logger.info('Convert to EC2Instance from ' + str(response_instance))
     email = None
     tags = response_instance.get('Tags')
-    if tags is not None:
+    if tags != None:
```
nikolayilyin, line 124: would it be better to write as 'if tags:'? 
to check also empty string and an empty list.

nikolayilyin (approved):
lgtm

aws/src/main/python/cloudwatchAllInstances/lambda_function.py
JustinPihony: Sure - although it would not make a difference in this case afaict

JustinPihony:
Test!

JustinPihony:
Test!

PR #3096, opened 2021-01-13
Hybrid Router

dimaopen:


dimaopen:
Test!

JustinPihony:
Test!

src/main/scala/beam/router/RoutingWorker.scala
```diff
@@ -107,8 +112,23 @@ class RoutingWorker(workerParams: R5Parameters) extends Actor with ActorLogging
     if (carRouter == "staticGH" || carRouter == "quasiDynamicGH") {
       new Directory(new File(graphHopperDir)).deleteRecursively()
       createWalkGraphHopper()
-      createCarGraphHoppers()
+      createCarGraphHoppers(carRouter)
       askForMoreWork()
+    } else if (carRouter == "hybridGH") {
+      createWalkGraphHopper()
```
JustinPihony, line 118: Should these methods return the object itself? Then you don't have to do any of the casting below..which is not compile safe.

src/main/scala/beam/router/HybridRouter.scala
```diff
@@ -0,0 +1,119 @@
...
+    if (request.withTransit) {
+      //find routes to the nearest stop via GH
+      val stop = stopQuadTree.getClosest(request.originUTM.getX, request.originUTM.getY)
+      val stopLocation = geo.wgs2Utm(new Coord(stop.stop_lon, stop.stop_lat))
+      val ghRequest = request.copy(destinationUTM = stopLocation, withTransit = false)
+      val ghResponses: IndexedSeq[BeamRouter.RoutingResponse] = calcRouteGH(ghRequest).map(_._2)
+      val resultItineraries: IndexedSeq[EmbodiedBeamTrip] = if (ghResponses.nonEmpty) {
+        //find routes from the nearest stop to the destination
+        val bodyStreetVehicle = request.streetVehicles
+          .filter(vehicle => vehicle.mode == WALK)
```
JustinPihony, line 54: `filter`...`map` should be a `collect`

src/main/scala/beam/router/graphhopper/BikeGraphHopperWrapper.scala
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  override def prepareRequest(request: GHRequest): Unit = {
+    request.setProfile(FastestProfileName)
+    request.setPathDetails(Seq(Parameters.Details.EDGE_ID, Parameters.Details.TIME).asJava)
+  }
+
+  override protected def getLinkTravelTimes(responsePath: ResponsePath, totalTravelTime: Int): IndexedSeq[Double] = {
+    responsePath.getPathDetails
+      .asScala(Parameters.Details.TIME)
```
JustinPihony, line 37: I don't have a compiler while checking, but why two `asScala`'s?

JustinPihony (approved):
Nothing blatent - please still review, but otherwise approved

src/main/scala/beam/router/RoutingWorker.scala
dimaopen: Probably you are right. I'll refactor these methods.

src/main/scala/beam/router/graphhopper/BikeGraphHopperWrapper.scala
dimaopen: The first one for converting the Map, the second one for converting the values of that map.

src/main/scala/beam/router/HybridRouter.scala
dimaopen: Fixed

src/main/scala/beam/router/RoutingWorker.scala
dimaopen: I cannot change createCarGraphHoppers to return the object because it also does some side effects. I resolved this issue by assigning the right subtypes to the fields.

JustinPihony:
Test!

dimaopen:
Do not merge this PR unless Art approved the approach.

PR #3100, opened 2021-02-01
add profiler

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3100)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,4 @@
-    import java.time.Instant
+import java.time.Instant
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 
--- a/build.gradle
+++ b/build.gradle
@@ -1,4 +1,4 @@
-import java.time.Instant
+    import java.time.Instant
 import java.time.ZoneOffset
 import java.time.format.DateTimeFormatter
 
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -1,4 +1,4 @@
-import java.time.Instant
+    import java.time.Instant
```
REASY, line 1: nitpick: formatting

REASY (commented):
Will it work like that without any change to aws/build.gradle? How that parameter will be propagated from someone's machine to AWS?

build.gradle
```diff
@@ -505,10 +505,21 @@ def jmx = ["-Dcom.sun.management.jmxremote", "-Dcom.sun.management.jmxremote.por
 
 def logbackConfig = project.findProperty('logbackCfg') ?: "logback_prod.xml"
 
+def profilerTypeConfig = project.findProperty('profiler_type') ?: null
+
+def profilerOption = ""
+if( profilerTypeConfig.toString().equalsIgnoreCase('cpu')) {
+    profilerOption = jfr
+}
+else if (profilerTypeConfig.toString().equalsIgnoreCase('cpumem')){
```
JustinPihony, line 514: I'm unsure of how groovy handles null, but should the config default to an empty string instead of null?

JustinPihony:
> Will it work like that without any change to aws/build.gradle? How that parameter will be propagated from someone's machine to AWS?

I do not believe this will propagate to AWS - and will need to be added explicitly in other ways. If that is wanted then we should add a means explicitly.

JustinPihony:
test!

JustinPihony (approved):

JustinPihony:
Approved, but still double back on the comments please. I just don't see a reason to block it, though.

build.gradle
rajnikantsh: apply changes for this

build.gradle
rajnikantsh: > > Will it work like that without any change to aws/build.gradle? How that parameter will be propagated from someone's machine to AWS?
> 
> I do not believe this will propagate to AWS - and will need to be added explicitly in other ways. If that is wanted then we should add a means explicitly.

yes it was not propagated in  AWS, But now its updated

REASY (approved):

PR #3156, opened 2021-03-29
Fix logging for excluded modes

crixal:


JustinPihony (approved):
Other than it using a mutable val instead of a var, what was the problem? It should work both ways.

JustinPihony:
NVM, I see the attached ticket - and agree with the change - which I already approved :)

crixal:
test!

PR #3159, opened 2021-04-04
Added option to convert Physsim network to pbf directly

### LINKED ISSUE #3153: PhyssymXmlToOsmConverter should be able to save in OSM.PB
Parent issue: https://github.com/LBNL-UCB-STI/beam/issues/3076

carloscaldas:
Closes #3153 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3159)

Changes:
- Fix bug: currently the Physym to OSM repeats the wayId. In the new version a unique wayId is extracted from linkId of Physsim network
- Allow to convert Physym network to PBF (in addition to OSM XML) 
- Small refactoring to improve log

### DIFF
--- a/src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
+++ b/src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
@@ -68,7 +68,6 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
               startElement.getAttributeByName(new QName("to")).getValue
             )
             wayInnerTags = buildLinkProps(startElement)
-            links += Way(wayId, source_destination._1, source_destination._2, wayInnerTags)
           case "attributes" if canProcessLink =>
             canProcessLinkAttributes = true
           case "attribute" if canProcessLinkAttributes && tagNameHasKind(startElement, "origid") =>
@@ -86,6 +85,7 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
         endElement.getName.getLocalPart match {
           case "nodes"                  => canProcessNode = false
           case "link" if canProcessLink =>
+            links += Way(wayId, source_destination._1, source_destination._2, wayInnerTags)
           case "links" if canProcessLink =>
             canProcessLink = false
           case "attributes" if canProcessLinkAttributes => canProcessLinkAttributes = false
--- a/src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
+++ b/src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
@@ -2,7 +2,7 @@ package beam.router
 
 import java.io.{File, FileInputStream, FileOutputStream}
 import java.nio.charset.StandardCharsets
-import java.nio.file.{Files, Path, Paths}
+import java.nio.file.{Files, Path}
 import java.time.{LocalDateTime, ZoneOffset}
 import java.time.format.DateTimeFormatter
 import javax.xml.namespace.QName
@@ -15,6 +15,10 @@ import scala.util.{Success, Try}
 
 import beam.agentsim.infrastructure.geozone.WgsCoordinate
 import beam.router.Modes.BeamMode
+import beam.router.PhyssymXmlToOsmConverterParams.OutputType
+import beam.router.PhyssymXmlToOsmConverterParams.OutputType.OutputType
+import beam.utils.FileUtils
+import com.conveyal.osmlib.{Node => ConveyalOsmNode, OSM => ConveyalOsm, Way => ConveyalOsmWay}
 import com.typesafe.scalalogging.StrictLogging
 import scopt.OParser
 
@@ -30,10 +34,12 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
     var canProcessLink = false
     var canProcessLinkAttributes = false
     var canProcessAttribute = false
+
     var isWayId = false
     var isHighway = false
+    var wayCounter = 1L
 
-    var wayId: String = null
+    var wayId: String = String.valueOf(wayCounter)
     var wayInnerTags: mutable.Buffer[(String, String)] = null
     var source_destination: (String, String) = null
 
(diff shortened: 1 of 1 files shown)


src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
```diff
@@ -30,10 +34,12 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
     var canProcessLink = false
     var canProcessLinkAttributes = false
     var canProcessAttribute = false
+
     var isWayId = false
     var isHighway = false
+    var wayCounter = 1L
 
-    var wayId: String = null
+    var wayId: String = String.valueOf(wayCounter)
```
JustinPihony, line 42: The links and wayCounter change logic, whereas the PR is listed as enabling PBF. The two changes should probably be in separate PRs - or at least document the changes being made. That way future debuggers can trace changes easier if needed.

JustinPihony (changes requested):

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
carloscaldas: @JustinPihony since was defined that two ways must be generated per link (one on each direction) the previous code was generating two ways with the same id. it is not only a "changing of behavior" - it is a bug fix. QGIS don't care about this bug, but the conveyal osm library keep only one (since its use a map on its data structure)
Since this code is relatively new and it was merged last week I believe this bug fix won't affect any client

carloscaldas:
Test now!

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
JustinPihony: OK, can you at least add a comment to the PR description then?

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
```diff
@@ -62,6 +68,7 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
               startElement.getAttributeByName(new QName("to")).getValue
             )
             wayInnerTags = buildLinkProps(startElement)
+            links += Way(wayId, source_destination._1, source_destination._2, wayInnerTags)
```
JustinPihony, line 71: Also, you added to the links here, but removed from the endElement section. Shouldn't it still be added there also?

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
carloscaldas: sure

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
carloscaldas: nice catch. this change should not have been done

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
```diff
@@ -61,48 +64,47 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
...
           case somethingElse =>
             logger.warn(s"Unidentified tag: [$somethingElse]")
         }
       } else if (nextEvent.isEndElement) {
         val endElement = nextEvent.asEndElement()
         endElement.getName.getLocalPart match {
-          case "nodes" => canProcessNode = false
-          case "link" if canProcessLink =>
-            links += Way(wayId, source_destination._1, source_destination._2, wayInnerTags)
+          case "network"                =>
```
JustinPihony, line 81: Why catch this and do nothing? To avoid the not predicted logging?

JustinPihony: Same with `"node" if canProcessNode`

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
```diff
@@ -61,48 +64,47 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
...
-          case "attribute" if canProcessLinkAttributes && isWayId =>
-            isWayId = false
-          case "attribute" if canProcessLinkAttributes && isHighway =>
-            isHighway = false
+          case "link" if canProcessLink =>
+            links += Way(wayId, source_destination._1, source_destination._2, wayInnerTags)
+          case "attributes" if canProcessLinkAttributes =>
+            canProcessLinkAttributes = false
+          case "attribute" if canProcessLinkAttributes =>
+            if (isHighway) {
```
JustinPihony, line 91: This changes the behavior so that it will no longer log attribute if it isn't `isHighway`

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
```diff
@@ -61,48 +64,47 @@ object PhyssymXmlToOsmConverter extends StrictLogging {
...
+            logger.warn(
+              s"Something not predicted. canProcessLinks: [$canProcessLink]; [canProcessLinkAttributes: [$canProcessLinkAttributes]; canProcessNode: [$canProcessNode]; somethingElse: [$somethingElse]"
+            )
         }
       } else if (nextEvent.isCharacters) {
         val characters: Characters = nextEvent.asCharacters()
         val charactersValue = characters.getData
-        if (!isBlank(charactersValue)) {
-          if (isWayId) {
-            wayId = charactersValue
```
JustinPihony, line 98: This was removed as part of the refactoring you did - was that on purpose? Or is that the bugfix you mention?

JustinPihony (approved):
One change that needs verified, but is probably fine - others are stylistic more than anything else

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
carloscaldas: is the bugfix

src/test/scala/beam/router/PhyssymXmlToOsmConverter.scala
carloscaldas: to avoid log warning. the warning should never happen

PR #3188, opened 2021-05-26
Fixing performance issues

dimaopen:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3188)


dimaopen:
Test!

dimaopen:
Test!

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -59,7 +59,8 @@ class ChargingNetwork(managerId: Id[VehicleManager], chargingStationsQTree: Quad
     * @param vehicleId vehicle Id
     * @return charging vehicle
     */
-  def lookupVehicle(vehicleId: Id[BeamVehicle]): Option[ChargingVehicle] = vehicles.get(vehicleId)
+  def lookupVehicle(vehicleId: Id[BeamVehicle]): Option[ChargingVehicle] =
+    chargingZoneKeyToChargingStationMap.values.view.flatMap(_.lookupVehicle(vehicleId)).headOption
```
JustinPihony, line 63: This seems less performant, and a logic change - was this meant to be in this PR?

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -129,13 +130,16 @@ object ChargingNetwork {
       mutable.PriorityQueue.empty[ChargingVehicle](Ordering.by((_: ChargingVehicle).arrivalTime).reverse)
 
     def numAvailableChargers: Int = zone.numChargers - connectedVehiclesInternal.size
-    def connectedVehicles: Map[Id[BeamVehicle], ChargingVehicle] = connectedVehiclesInternal.toMap
+    def connectedVehicles: collection.Map[Id[BeamVehicle], ChargingVehicle] = connectedVehiclesInternal
```
JustinPihony, line 133: The reason (as far as I can tell) this was passed as `toMap` was to make sure there would be no mutation of the underlying backing. I would not suggest this change..

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -147,7 +151,7 @@ object ChargingNetwork {
       vehicle: BeamVehicle,
       stall: ParkingStall,
       theSender: ActorRef
-    ): ChargingVehicle = this.synchronized {
+    ): ChargingVehicle = {
```
JustinPihony, line 154: These synchronizations were likely added to make sure the data did not get corrupted...we should leave these. Unless you are going to convert this into an actor or something else which can handle parallel reads/writes.

src/main/scala/beam/router/skim/core/AbstractSkimmer.scala
```diff
@@ -58,7 +58,7 @@ abstract class AbstractSkimmerReadOnly extends LazyLogging {
   def getCurrentSkimValue(key: AbstractSkimmerKey): Option[AbstractSkimmerInternal] =
     Option(currentSkimInternal.get(key))
   def aggregatedFromPastSkims: Map[AbstractSkimmerKey, AbstractSkimmerInternal] = aggregatedFromPastSkimsInternal
-  def pastSkims: Map[Int, collection.Map[AbstractSkimmerKey, AbstractSkimmerInternal]] = pastSkimsInternal.toMap
+  def pastSkims: collection.Map[Int, collection.Map[AbstractSkimmerKey, AbstractSkimmerInternal]] = pastSkimsInternal
```
JustinPihony, line 61: Same as above - we shouldn't pass the internal directly..

JustinPihony (changes requested):
Is there an issue to link this to with more details on what you are trying to solve? Overall it looks like the changes make the code more sensitive to parallelism issues.

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
dimaopen: You are right, though it's used only for reading. Though to avoid mutation in the future I restored that toMap and minimized the number of calls to this method.

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
dimaopen: You are right, this makes this class not thread-safe, but this is the price for a better performance. Though I think that there's no big difference in performance. So I just restored these synchronizations.

src/main/scala/beam/router/skim/core/AbstractSkimmer.scala
dimaopen: I restored that toMap and minimized the number of calls to this method.

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
dimaopen: This is the main point of the PR. Generally, it does the same as in the previous version. But instead of building a Map (this operation is O(n)) and get the vehicle by Id, it looks up the vehicle in the internal map and if not found it looks up the vehicle in the PriorityQueue.

dimaopen:
> Is there an issue to link this to with more details on what you are trying to solve? Overall it looks like the changes make the code more sensitive to parallelism issues.

No issue is presented. But there's a performance issue when a lot of electric cars presented in the simulation. I restored all the sync/toMap methods.

dimaopen:
Test!

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
JustinPihony: Oy, this got complicated the more I looked into it. It seems like it is immutable at first, and I was going to suggest making the map a val - however I see that it is not - so I understand the change better.

JustinPihony (approved):

dimaopen:
Test!

dimaopen:
Test!

PR #3190, opened 2021-05-28
Improving parking manager performance

dimaopen:
I'm removing undesired alternatives when looking for a parking stall just in place (without reiterating it). That gives about 20% of performance gain for ParkingManager.
We could also collapse the ParkingZones using beam.agentsim.infrastructure.HierarchicalParkingManager$#collapse method. It would give us about 50% performance improvement. But it would slightly change the parking search because the number of alternatives is decreased.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3190)


dimaopen:
Test!

nikolayilyin (approved):
LGTM

PR #3214, opened 2021-06-15
Validate type id for shared vehicles. Require it to start with "sharedVehicle" instead of adding it silently.

dimaopen:
This allows the users to have better understanding of the errors happens during usage of shared vehicle scenarios.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3214)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -48,6 +48,7 @@ sourceCompatibility = 1.8
 targetCompatibility = 1.8
 
 compileScala.options.encoding = 'UTF-8'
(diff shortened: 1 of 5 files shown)


dimaopen:
Test!

JustinPihony (approved):

JustinPihony:
CI is failing due to GC....but there are no changes in this PR that would cause that - I am looking into what is going on ...

PR #3216, opened 2021-06-24
change default listing differences to comments - remove warning

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3216)


JustinPihony:
The test is failing because I removed the default values. It seemed arbitrary since only the first was entered - thoughts on whether I should fix the tests or re-add a default?

nikolayilyin (approved):
lgtm

PR #3221, opened 2021-06-29
Urbansim mode mapping from config

### LINKED ISSUE #3218: Urbansim scenario reader: add smart mode mapping
Right now we have a hard-coded mapping of activitySim modes into beamModes, 
it would be better to move such mapping into the config file.

Current version of config reader does not support defoult values for lists, so, it in order to move mode mapping into the config and to have a default mapping at the same time one might need to upgrade version of config reader.

dimaopen:
One  must set beam.exchange.scenario.modeMap parameter when they use urbansim_v2 scenario

Resolves #3218

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3221)

### DIFF
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -357,6 +357,31 @@ beam.exchange.scenario {
   # Input file format for scenario loader can be "xml", "csv" or "parquet"
   fileFormat = "xml"
   folder = ""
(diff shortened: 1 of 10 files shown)


nikolayilyin (approved):
lgtm

PR #3230, opened 2021-07-07
Upgraded gradle wrapper version

### LINKED ISSUE #3229: Upgrade gradle to latest version (7.1)

carloscaldas:
Closes #3229

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3230)


build.gradle
```diff
@@ -118,197 +125,197 @@ configurations {
...
-    compile group: 'de.heikoseeberger', name: "akka-http-circe_${scalaBinaryVersion}", version: "1.25.2"
+    implementation group: 'com.typesafe.akka', name: "akka-http_${scalaBinaryVersion}", version: "10.1.12"
+    implementation group: 'de.heikoseeberger', name: "akka-http-circe_${scalaBinaryVersion}", version: "1.25.2"
 
     // TEST Akka //
-    testCompile group: 'com.typesafe.akka', name: "akka-testkit_${scalaBinaryVersion}", version: akkaBinaryVersion
+    testImplementation group: 'com.typesafe.akka', name: "akka-testkit_${scalaBinaryVersion}", version: akkaBinaryVersion
 
-    scoverage "org.scoverage:scalac-scoverage-plugin_${scalaBinaryVersion}:1.3.1", "org.scoverage:scalac-scoverage-runtime_${scalaBinaryVersion}:1.3.1"
+//    scoverage "org.scoverage:scalac-scoverage-plugin_${scalaBinaryVersion}:1.4.8", "org.scoverage:scalac-scoverage-runtime_${scalaBinaryVersion}:1.4.8"
```
JustinPihony, line 288: Was this commented for testing? Otherwise why remove?

build.gradle
```diff
@@ -331,7 +338,7 @@ configurations.all {
     exclude group: 'javax.media', module: 'jai_imageio'
 }
 
-//compileScala.dependsOn(scalafmtAll)
+compileScala.dependsOn(scalafmtAll)
```
JustinPihony, line 341: This should be reviewed in the standup as I vaguely recall the reason for this was that it messed with compile times - so we added the jenkins check to catch it before it was fully commited

JustinPihony (changes requested):
My comments should be addressed before moving forward

build.gradle
carloscaldas: @JustinPihony it was just testing

build.gradle
carloscaldas: agree. it basically change a lot of untouched files

alex-vv:
LGTM, but I have two non-blocking suggestions:
1. Currently there is a deprecation warning on the startup which may be an issue for the future upgrade to Gradle 8. You can check these issues using `./gradlew --warning-mode all`
2. New Gradle has a configuration cache feature which can probably improve a build time. Currently it cannot be used because `grgit` plugin doesn't support it. I think this plugin is not used anywhere and can be removed.

alex-vv (approved):

build.gradle
```diff
@@ -285,7 +285,8 @@ dependencies {
     // TEST Akka //
     testImplementation group: 'com.typesafe.akka', name: "akka-testkit_${scalaBinaryVersion}", version: akkaBinaryVersion
 
-//    scoverage "org.scoverage:scalac-scoverage-plugin_${scalaBinaryVersion}:1.4.8", "org.scoverage:scalac-scoverage-runtime_${scalaBinaryVersion}:1.4.8"
+    scoverage "org.scoverage:scalac-scoverage-plugin_2.12.12:1.4.8", "org.scoverage:scalac-scoverage-runtime_2.12:1.4.8"
```
JustinPihony, line 288: This should continue to use the `scalaBinaryVersion` so that we don't miss it in future upgrades

carloscaldas: we have two patterns here:
`scalac-scoverage-plugin_2.12.12` and `scalac-scoverage-runtime_2.12`
I can have two variables for that but there is no guarantee releases of component will keep same approach since currently is different

carloscaldas: @JustinPihony done

JustinPihony: That's fun..

JustinPihony (approved):

PR #3242, opened 2021-07-20
A script to filter points by shapefile

nikolayilyin:
This is a small tool that, probably, never won't be used as is.
I have caught myself writing almost the same script third time, so, here it is.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3242)

### DIFF
--- a/src/main/scala/beam/utils/shape/FilterPointsInShapeFile.scala
+++ b/src/main/scala/beam/utils/shape/FilterPointsInShapeFile.scala
@@ -1,0 +1,75 @@
+package beam.utils.shape
+
+import beam.utils.csv.CsvWriter
+import beam.utils.data.synthpop.GeoService.findShapeFile
+import beam.utils.map.ShapefileReader
+import beam.utils.scenario.urbansim.censusblock.entities.InputPlanElement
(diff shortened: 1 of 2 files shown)


nikolayilyin:
test!

src/main/scala/beam/utils/shape/FilterPointsInShapeFile.scala
```diff
@@ -0,0 +1,74 @@
...
+        case InputPlanElement(personId, _, _, _, Some(x), Some(y), _) => Some(personId, x, y)
+        case _                                                        => None
+      }
+      .map { case (personId, x, y) => (personId, gf.createPoint(new Coordinate(x, y))) }
+
+    points.toIndexedSeq
+  }
+
+  def main(args: Array[String]): Unit = {
+    val pathToShp = "/home/nikolay/Downloads/Oakland_Alamenda_TAZ/Oakland+Alameda+TAZ/Oakland+Alameda+TAZ"
```
dimaopen, line 46: I would suggest adding parameters as in ```BackgroundSkimsCreatorApp```.

dimaopen (approved):

PR #3257, opened 2021-08-12
[WIP] plan en-route charging if recharging needed

heyrutvik:
Status:

- [x] figure out if recharging needed

- [x] request charging stations filtered by charging types (>= 50Kw)

- [ ] WIP based on nearby stall provided by charging station manager, send router a request to re-plan the trip

- [ ] write tests

Question:
I used existing isRefuelNeeded method. I think it should be working with any vehicle, not just ride hail, correct?

Cc @nikolayilyin @haitamlaarabi

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3257)


src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -763,8 +767,30 @@ class PersonAgent(
...
-  onTransition { case _ -> _ =>
-    unstashAll()
+  onTransition {
+    // todo rrp
+    case _ -> PlanningEnRouteCharging =>
+      nextStateData match {
+        case data: BasePersonData =>
+          data.currentVehicle.headOption
+            .map(beamVehicleId => beamVehicles(beamVehicleId).vehicle)
+            .fold(log.error("PlanningEnRouteCharging: vehicle not found.")) { vehicle =>
```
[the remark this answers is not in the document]
heyrutvik: Thanks, @nikolayilyin. The code is updated as per your suggestion.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -833,6 +861,16 @@ class PersonAgent(
...
             currentVehicle
           }
+        // todo rrp
+        currentVehicleForNextState.headOption.foreach { beamVehicleId =>
+          val vehicle = beamVehicles(beamVehicleId).vehicle
+          val refuelNeeded = vehicle.isRefuelNeeded(
+            beamScenario.beamConfig.beam.agentsim.agents.vehicles.rechargeRequiredThresholdInMeters,
+            beamScenario.beamConfig.beam.agentsim.agents.vehicles.noRechargeThresholdInMeters
+          )
+          if ((vehicle.isBEV || vehicle.isPHEV) && refuelNeeded)
```
haitamlaarabi, line 871: We also should look at the distance to destination.. if distance is short then let it follow default workflow, if destination is far then we plan enroute charging.
Maybe you can do with: 

val newRechargeRequiredThresholdInMeters = rechargeRequiredThresholdInMeters + (1-rechargeRequiredThresholdInMeters/distanceToDestinationInMeters)*distanceToDestinationInMeters

something similar should be done to noRechargeThresholdInMeters

haitamlaarabi (changes requested):

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -763,8 +768,31 @@ class PersonAgent(
...
+            case None => log.error("PlanningEnRouteCharging: vehicle not found.")
+            case Some(vehicle) =>
+              chargingNetworkManager ! ParkingInquiry(
+                destinationUtm = vehicle.spaceTime,
+                activityType = ParkingActivityType.Charge,
+                vehicleManagerId = vehicle.vehicleManagerId,
+                beamVehicle = Some(vehicle),
+                valueOfTime = attributes.valueOfTime,
+                triggerId = getCurrentTriggerIdOrGenerate,
+                chargingPointTypes =
```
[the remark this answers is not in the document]
heyrutvik: @haitamlaarabi, I made the change as you have suggested. We will use `activityType: ParkingActivityType` of `ParkingInquiry` rather than adding new property `chargingPointTypes`. The "filter" logic is updated as well, so if activity type is `FastCharge`, look for >= 50 Kw power otherwise any will do (as long as we have charging point). There is a doubt, but I'll ask that once I push new changes.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
heyrutvik: Got it. Correct me if I'm wrong, we need to use `distanceToDestinationInMeters` to come up with new thresholds (at run time) and use them to compute whether recharging it required or not. 

@haitamlaarabi, Could you please provide equation for `noRechargeThresholdInMeters` as well? I can't figure it out how do that.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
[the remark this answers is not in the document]
heyrutvik: @haitamlaarabi, while skimming though charging point related code, I found we already have a threshold and predicate defined [here](https://github.com/LBNL-UCB-STI/beam/blob/rrp/calculate-battery-level-and-charging-point-stay-time/src/main/scala/beam/agentsim/infrastructure/charging/ChargingPointType.scala#L179) to check for "fast charger". Apparently, we use in `ifRideHailCurrentlyOnShiftThenFastChargingOnly` method, for the same purpose, to filter out charging stations. Should I make changes in existing thing or continue doing what I've done (you can check in newer commits)?

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -833,6 +859,29 @@ class PersonAgent(
...
+          val vehicle = beamVehicles(beamVehicleId).vehicle
+          if (vehicle.isBEV || vehicle.isPHEV) {
+            val distanceToDestinationInMeters = trip.foldLeft(0.0)(_ + _.beamLeg.travelPath.distanceInM)
+            val newRechargeRequiredThresholdInMeters = {
+              rechargeRequiredThresholdInMeters +
+              (1 - rechargeRequiredThresholdInMeters / distanceToDestinationInMeters) *
+              distanceToDestinationInMeters
+            }
+            // todo: compute new noRechargeThresholdInMeters using distanceToDestinationInMeters value.
+            val newNoRechargeThresholdInMeters = noRechargeThresholdInMeters
```
heyrutvik, line 878: @haitamlaarabi, Could you please provide equation for `noRechargeThresholdInMeters` as well? I can't figure it out how do that.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -833,6 +859,29 @@ class PersonAgent(
...
+        // todo rrp
+        currentVehicleForNextState.headOption.foreach { beamVehicleId =>
+          import beamScenario.beamConfig.beam.agentsim.agents.vehicles.{
+            noRechargeThresholdInMeters,
+            rechargeRequiredThresholdInMeters
+          }
+
+          val vehicle = beamVehicles(beamVehicleId).vehicle
+          if (vehicle.isBEV || vehicle.isPHEV) {
+            val distanceToDestinationInMeters = trip.foldLeft(0.0)(_ + _.beamLeg.travelPath.distanceInM)
```
dimaopen, line 871: I think we need to filter the trip so that leg.vehicle == vehicle. The rest of trip can include the walk and/or transit legs.

src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala
```diff
@@ -170,10 +171,21 @@ class ParkingFunctions[GEO: GeoLevel](
     preferredParkingTypes: Set[ParkingType]
   ): Boolean = {
 
+    val fiftyPowerInKw: Double = 50.0
```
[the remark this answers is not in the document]
heyrutvik: Yes, that makes sense. I was waiting for @haitamlaarabi's response on https://github.com/LBNL-UCB-STI/beam/pull/3257#discussion_r690311354. Then I'll make appropriate changes.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
heyrutvik: Ah, nice catch. Thanks! I'll add filter part. :+1:

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -833,6 +886,38 @@ class PersonAgent(
...
+              distanceToDestinationInMeters
+            }
+            // todo: compute new noRechargeThresholdInMeters using distanceToDestinationInMeters value.
+            val newNoRechargeThresholdInMeters = noRechargeThresholdInMeters
+            val refuelNeeded =
+              vehicle.isRefuelNeeded(newRechargeRequiredThresholdInMeters, newNoRechargeThresholdInMeters)
+
+            if (refuelNeeded) {
+              val enrouteCharging = EnrouteCharging(
+                destinationUTM = beamServices.geo.wgs2Utm(trip.last.beamLeg.travelPath.endPoint.loc)
```
heyrutvik, line 915: @dimaopen, is this the correct way to find destination of a trip? We want to remember this to re-route the vehicle once it done recharging.

heyrutvik: @dimaopen, made the change. See [here](https://github.com/LBNL-UCB-STI/beam/pull/3257/files#diff-940cd21d65d4ec3f2f9c49c5aaf9753e3b2b4d030fabd7a7d4ed861a8eda857fR898), we basically use it for two things.
1) to find distance 
2) to find end point

src/main/scala/beam/agentsim/agents/PersonAgent.scala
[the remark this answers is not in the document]
haitamlaarabi: Yes, feel free to use that one. You can also parameterized it so that we can change that 20kW from config file if we want to.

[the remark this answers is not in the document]
haitamlaarabi: You can also add a function similar to: `ifRideHailCurrentlyOnShiftThenFastChargingOnly`, or use the same by addition not only RideHail, but also private EVs who want to EnRoute charge. so instead of ParkingActivityType.FastCharge, create ParkingActivityType.EnRouteCharge

src/main/scala/beam/agentsim/agents/PersonAgent.scala
haitamlaarabi: yes, we need to know if the battery will run low while driving towards destination. If it is not running low then we can leave workflow as it is. If it is running low then we need to EnRoute charge

src/main/scala/beam/agentsim/agents/PersonAgent.scala
dimaopen: After charging we need generally do the same thing we do during Mode Choice (ChoosesMode trait).
Before we go to EnRoute charging workflow we need to save the final destination parking location (it's the end of the last (second) leg of this vehicle). And probably we need to save the last WALK leg (parking -> our Activity location). At some point we need to get the route from charging parking to the final parking. Then we need to split this leg for parking using `beam.agentsim.agents.modalbehaviors.ChoosesMode#splitLegForParking` and add the last WALK leg to the final destination.
This needs to be saved to the `currentTrip` field (and probably `restOfCurrentTrip`) of `PersonData`. And then we can go to `ProcessingNextLegOrStartActivity` state.

heyrutvik:
These are old changes. The work has been completed through https://github.com/LBNL-UCB-STI/beam/pull/3433 PR.

PR #3266, opened 2021-08-23
Util to compare values of two config files

nikolayilyin:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3266)

### DIFF
--- a/src/main/scala/beam/utils/ConfigValuesComparator.scala
+++ b/src/main/scala/beam/utils/ConfigValuesComparator.scala
@@ -5,16 +5,28 @@ import com.typesafe.scalalogging.LazyLogging
 
 import scala.collection.JavaConverters._
 
+/*
+Application expects to get two config files as input
+args(0) - config file #1
+args(1) - config file #2
+
+The output of application will contains the difference between values of two provided config files.
+Configs will be resolved, all included configs will be parsed too and all nested config values will be used in comparison.
+The order of config keys in the config files does not matter, but, if config entry contains list of values
+and the order in the list is different, then it will be considered as difference.
+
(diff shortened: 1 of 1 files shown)


nikolayilyin:
test!

src/main/scala/beam/utils/ConfigValuesComparator.scala
```diff
@@ -0,0 +1,55 @@
...
+object ConfigValuesComparator extends LazyLogging {
+
+  def getFullConfigMap(configFileLocation: String): Predef.Map[String, String] = {
+    val renderOptions = ConfigRenderOptions.concise()
+
+    val config = BeamConfigUtils.parseFileSubstitutingInputDirectory(configFileLocation)
+    config
+      .entrySet()
+      .asScala
+      .map(x => x.getKey -> x.getValue.render(renderOptions).split(',').last)
```
JustinPihony, line 17: Does this handle nested configurations? I'd be especially curious about nested configs that are mismatched on order.

JustinPihony (approved):
Basic question for clarity - the code itself is otherwise simple enough

rajnikantsh:
@nikolayilyin can you add small description in comments about usage of this script. What parameters is it expecting and what will be output ? otherwise LGTM

src/main/scala/beam/utils/ConfigValuesComparator.scala
nikolayilyin: It will handle nested configurations and if the order of included configs is different and included configs contains the same key, it will use the latest version of the value.

nikolayilyin:
test!

src/main/scala/beam/utils/ConfigValuesComparator.scala
JustinPihony: I think the newest change is what does that now. thanks!

nikolayilyin:
test!

PR #3272, opened 2021-08-26
Production data in separate repos

### LINKED ISSUE #3254: Separation of logic and data. Moving cities data into another git repos
Currently we store all data for scenarios in git repository together with code, it was fine while we had one city. 
Now we have three (Detroit, NYC, San Francisco) and keeping them together with code creates problems: size of repository is growing, we create many branches with different city data and then it is complicated to merge things back into develop or we have outdated code in these branches. 

One of promising approach - moving data for each city into separate git repository, one repository for each city.
The goals of this task:
- investigation of this approach
- checking if this approach helps to reduce the size of code repository, the problem is - git has history and even after moving city data from code branch we might have big size just because of history, but we use git-lfs for big files, so, we might not have this problem.
- investigation of details (how to use multiple repos? is it possible to use nested repos for cities?)

alex-vv:
Production data is extracted with these paths

`test/input/texas` -> https://github.com/LBNL-UCB-STI/beam-data-austin
`production/sfbay` @ branch `production-gemini-develop` ->  https://github.com/LBNL-UCB-STI/beam-data-sfbay
`test/input/newyork` @ branch `art/new-york-scenario` -> https://github.com/LBNL-UCB-STI/beam-data-newyork
`test/input/detroit` @ branch `inm/merge-urbansim-with-detroit` -> https://github.com/LBNL-UCB-STI/beam-data-detroit

I also added `production/common` directory which contains common configuration for production scenarios.

All production scenarios are mapped to `production` directory using git submodules:
SFbay: `production/sfbay`
New York: `production/newyork`
Detroit: `production/detroit`
Austin: `production/austin` 

I updated documentation with examples on how to use these scenarios locally.

Fixes #3254, #3275 

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3272)


alex-vv:
test!

nikolayilyin (approved):
LGTM

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -112,6 +156,19 @@
   - sudo git lfs pull
   - echo "git checkout -qf ..."
   - GIT_LFS_SKIP_SMUDGE=1 sudo git checkout -qf $COMMIT
+  
+  - SUBMODULES=$(git submodule | awk '{ print $2 }')
```
JustinPihony: Can you add the simulateBeam.yaml environment backup? And/Or should this be a passed in as a parameter rather than loading it all?

JustinPihony (changes requested):

aws/src/main/python/beam_lambda/lambda_function.py
alex-vv: Not sure what do you mean regarding backup? I now think it would be better to name this local variable in lower case to make it clear that it is not environment variable and make it consistent with other local variables. Will change now.

Regarding parameters the idea is that a user would pass one or several configs and then this script can automatically detect which submodules it needs to load, so we don't need to specify the same path twice, for a config and for a submodule.

alex-vv:
test!

docs/developers.rst
```diff
@@ -103,46 +103,51 @@ Sometimes it is possible to face a timeout issue when trying to push huge files.
 
 #. Just push the files as usual
 
-Keeping Production Data out of Master Branch
+Production Data And Git Submodules
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Production versus test data. Any branch beginning with "production" or "application" will contain data in the "production/" subfolder. This data should stay in that branch and not be merged into master. To keep the data out, the easiest practice is to simply keep merges one-way from master into the production branch and not vice versa.
+Production data is located in separate git repositories each scenario in its own repo (see `#3254 <https://github.com/LBNL-UCB-STI/beam/issues/3254>`_ for the reason)
```
nikolayilyin: I think the link to the issue is not what we need in documentation, instead I assume using the following explanation:
Separation of production data and code is needed to: 
1. reducing the git repository size for developers
2. easier addition or changing production data without merging back into develop code changes
3. ability to use any production data with any code branch/commit without creation of yet another git production branch

nikolayilyin (approved):

docs/developers.rst
alex-vv: done

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

aws/src/main/python/beam_lambda/lambda_function.py
JustinPihony: I misread this. I thought it was reading the variables from an environment variable in lambda.

JustinPihony (approved):

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

PR #3279, opened 2021-08-29
Scapegoat - Enabled lint for RepeatIfElseBody

### LINKED ISSUE #3278: Scapegoat - RepeatedIfElseBody
Related to #2985

Currently the lint for `RepeatedIfElseBody` is disabled. Ideally all lints should be enabled

carloscaldas:
Closes #3278 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3279)


heyrutvik (dismissed):
I would suggest to remove unnecessary (or commented out) code from the codebase. I know we have them at lot of places in the project. We rather replace them with short informative comment, if we think we need to implement it in the future. Thoughts? @JustinPihony @dimaopen 

Otherwise, actual changes LGTM, @carloscaldas .

dimaopen:
> I would suggest to remove unnecessary (or commented out) code from the codebase. I know we have them at lot of places in the project. We rather replace them with short informative comment, if we think we need to implement it in the future. Thoughts? @JustinPihony @dimaopen

I tend to agree. We also may put the code deletion in a separate commit with the comment describing what we're doing and why. This can help to find deleted code in the future if needed.

carloscaldas:
@heyrutvik and @dimaopen I agree with both of you. Since the code only runs when debug level is enabled I uncommented the code. please check again

heyrutvik:
> Since the code only runs when debug level is enabled I uncommented the code. please check again

There is a misunderstanding. We didn't mean it to [this particular](https://github.com/LBNL-UCB-STI/beam/pull/3279/commits/6f766767f3234fded4a223dcd8f1f5a3713c1e77#diff-24bb3ec617b1811f323c515caab5f0e72c7651a4a77d6695d054f60c7ff1d697R428) case. But if @dimaopen or @nikolayilyin is fine with the changes, I'm too. Let's wait till they review.

carloscaldas:
@heyrutvik in general commit commented code is a bad practice. What doesn't help gets in the way. For some cultural reason the project have a lot of this scenario (like you pointed out). We need to discuss with the team a general strategy, however for scope of this PR we only have three options:
- remove all commented code and affect dependencies
- comment the if (keep the problem already exists) and get rid of the lint error
- **uncomment the commented code and get rid of the lint error. since there is a debug condition guard is not supposed to affect performance. IMO this is the best approach for the scope of this PR**

src/main/scala/beam/replanning/utilitybased/ChainBasedTourVehicleAllocator.scala
```diff
@@ -74,10 +74,8 @@ case class ChainBasedTourVehicleAllocator(
       getVehicularToursSortedByStartTime(householdPlans).find(rec => rec.subtour == subtour)
 
     vehicularTours foreach { vt =>
-      if (allocateVehicles(vt)) {
-        //TODO: turn back on when using intra-household choice
-//        processAllocation(vt, plan)
-      }
+      //TODO: turn back on when using intra-household choice
+      allocateVehicles(vt)
     }
```
nikolayilyin: is allocateVehicles(vt) doing something by itself or just a check?
if it is doing something - I assume removing TODO, because it seems to be pointing at processAllocation function call
If it is not doing anything - I assume removing the whole loop

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -506,9 +506,9 @@ class BeamSim @Inject() (
       logger.warn(s"Non-negative end times found for person activities - ${activityEndTimesNonNegativeCheck.size}")
     }
 
-    if (beamConfig.beam.outputs.writeGraphs) {
-      // generateRepositioningGraphs(event)
-    }
+//    if (beamConfig.beam.outputs.writeGraphs) {
+//      generateRepositioningGraphs(event)
+//    }
```
nikolayilyin: I assume instead of commenting this out, introducing new configuration value
beamConfig.beam.outputs.writeRepositioningGraphs

src/main/scala/beam/utils/EventReplayer.scala
```diff
@@ -56,10 +56,10 @@ object EventReplayer extends BeamHelper {
...
     eventsManager.addHandler(new BasicEventHandler {
       def handleEvent(event: Event): Unit = {
-        if (event.getEventType == "PathTraversal") {
-          // Do something with need `PathTraversalEvent`
-          // buf += new PathTraversalEvent(event.getTime, event.getAttributes)
-        }
+        //  if (event.getEventType == "PathTraversal") {
+        //    Do something with need `PathTraversalEvent`
+        //    buf += new PathTraversalEvent(event.getTime, event.getAttributes)
+        //  }
```
nikolayilyin: looks like debug code, maybe it would be better to remove it completely?

src/main/scala/beam/replanning/utilitybased/ChainBasedTourVehicleAllocator.scala
JustinPihony: I agree that the modification here would be akin to:

```
/*if (*/allocateVehicles(vt)/*) {
        //TODO: turn back on when using intra-household choice
        processAllocation(vt, plan)
      }*/
```

src/main/scala/beam/sim/BeamSim.scala
JustinPihony: Yah, let's remove this altogether and let the commit history keep track if needed.

src/main/scala/beam/utils/EventReplayer.scala
JustinPihony: Agreed - same comment - let git history handle it

JustinPihony (changes requested):
As this is a no-op change, I'd say the comments mentioned should be addressed.

nikolayilyin:
test!

PR #3281, opened 2021-08-29
Scapegoat - Enabled lints for BigDecimalDoubleConstructor and EmptyTryBlock

### LINKED ISSUE #3280: Scapegoat - EmptyTryBlock and BigDecimalDoubleConstructor
Related to #2985

Currently the lint for `EmptyTryBlock` and `BigDecimalDoubleConstructor` are disabled. Ideally all lints should be enabled

carloscaldas:
Closes #3280 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3281)


carloscaldas:
test!

src/main/scala/beam/utils/MathUtils.scala
```diff
@@ -16,7 +16,7 @@ object MathUtils {
     * @return
     */
   def roundDouble(inVal: Double, scale: Int = 3): Double = {
-    BigDecimal(inVal).setScale(scale, BigDecimal.RoundingMode.HALF_UP).toDouble
+    BigDecimal.decimal(inVal).setScale(scale, BigDecimal.RoundingMode.HALF_UP).toDouble
```
heyrutvik, line 19: ~I think it's not necessary to add `.decimal` here. The `apply` method does the same thing.~

heyrutvik: Hm.. @carloscaldas, I assume the linter suggested to make such changes, correct? I think I'm missing some important facts about using BigDecimal constructors directly. Could you please shed some light on it? Thanks.

heyrutvik:
```
The results of this constructor can be somewhat unpredictable. E.g. writing new BigDecimal(0.1) in Java creates a BigDecimal which is actually equal to 0.1000000000000000055511151231257827021181583404541015625. This is because 0.1 cannot be represented exactly as a double.
```
Interesting! Got the point. @carloscaldas

heyrutvik (approved):
LGTM

src/main/scala/beam/utils/MathUtils.scala
carloscaldas: So... you realized before I reply :-)

Issue #3268, opened 2021-08-24
Config: there are too many values pointing to map.


nikolayilyin:
There are too many values that needs to point at the same folder with the map.
It leads to problems when some of values are pointed to one folder and some to another.
There are following config values that should point at the same folder (or files within):

`beam.routing.r5.directory` - most important

`beam.physsim.inputNetworkFilePath`
`beam.routing.r5.osmFile`
`beam.routing.r5.osmMapdbFile`

`beam.agentsim.scenarios.frequencyAdjustmentFile` (this one under question, probably the file itself might be placed anywhere)

One needs to reduce the number of config values that needs to point at the same location or add some sanity check during Beam initialization to reduce the number of human-made errors

alex-vv:
If I got it correctly then probably HOCON substitution may be useful here 
https://github.com/lightbend/config/blob/master/HOCON.md#substitutions

something like 
```
beam.physsim.inputNetworkFilePath = some/path
beam.routing.r5.directory = ${beam.physsim.inputNetworkFilePath}
beam.routing.r5.osmFile = ${beam.routing.r5.directory}/file.osm
...
```

probably there should be some additional common dir config value so in the end the actual path can be specified only in one place.

dgm90:
Fixed

PR #3293, opened 2021-09-01
3268 - fix config values pointing to the map

dgm90:
Fix for https://github.com/LBNL-UCB-STI/beam/issues/3268

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3293)

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -122,6 +122,13 @@ configurations {
 }
 
 
+configurations.all {
+    resolutionStrategy {
+//       we need this in order to get Elki library working fine.
+        force 'it.unimi.dsi:fastutil:8.5.4'
+    }
+}
+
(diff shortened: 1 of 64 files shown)


JustinPihony (approved):

JustinPihony:
Interestingly I cannot reproduce the CI error locally...

dgm90:
Test!

JustinPihony:
I finally figured it out - it was the renaming. I believe there are test files that use the old name, so if you rename then you have to do a global search in the project. More often at this point re-naming should be treated as an API change and done by creating two values while one is marked as deprecated in some way...or something like that

nikolayilyin (approved):
lgtm

PR #3297, opened 2021-09-06
Allow to choose the event manager for physsim

### LINKED ISSUE #3267: Low CPU usage during physSim with duplicates of PTE for physSim
Usually the total execution time of physSim (JDEQsim / BPRSim / ParBPRSim) is low comparing to agentSim, but if duplication of PTE events for physSim is turned on and the total number of events for physSim is significant (in big simulations) then execution time of physSim became as big as execution time of agentSim. 
This is not a problem by itself, the problem is that during execution of physSim the instance is barely consuming CPU (about 15-20%) while during agentSim the CPU consumption is close to 100% .

One needs to investigate possible ways to reduce execution time of physSim (for bigger simulations with duplicates of PTE) or explain why it is impossible.

dimaopen:
Introduced new praram `beam.physsim.eventManager.type`, if set to Auto it chooses Sequential eventmanager for PARBPRsim and Parallel Event manager for the other physsim engines.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3297)

### DIFF
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -431,6 +431,11 @@ beam.physsim.duplicatePTE.departureTimeShiftMax = "int | 600"
 beam.physsim.skipPhysSim = false
 # values: JDEQSim, BPRSim, PARBPRSim, CCHRoutingAssignment
 beam.physsim.name = "JDEQSim"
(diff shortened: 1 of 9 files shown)


nikolayilyin (approved):
lgtm

PR #3311, opened 2021-09-21
Fixed RoutingStatistic

dimaopen:
This PR fixes when writer wasn't fully flushed before akka terminates. Also fixed not openning a new writer when a new iteration starts.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3311)


dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

src/main/scala/beam/router/RoutingStatistic.scala
```diff
@@ -21,8 +22,9 @@ class RoutingStatistic(ioController: OutputDirectoryHierarchy) extends Actor {
 
   def handleRouterResponses(writer: CsvWriter): Receive = {
     case _: IterationEndsMessage =>
-      writer.flush()
       writer.close()
+      context.become(receive)
```
alex-vv, line 26: this will initialize a new writer, I wonder if this will overwrite the existing data in that file from the previous iteration?

JustinPihony: This would only be the case if it is the same iteration. If we want appending then we would have to modify the CsvWriter  to have the ability to call `getAppendingBufferedWriter`.

dimaopen: When we call `context.become(receive)`, the previous writer is closed. And we open a new one with a new file when `IterationStartsMessage` arrives.

JustinPihony: Yah, that should be guarantee enough :)

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -474,16 +474,17 @@ class BeamSim @Inject() (
       }
     }
 
-    if (beamConfig.beam.physsim.skipPhysSim) {
-      Await.result(Future.sequence(List(outputGraphsFuture)), Duration.Inf)
+    val physsimFuture = if (beamConfig.beam.physsim.skipPhysSim) {
```
JustinPihony: As this is only being waited on, there is no need for the first condition block:

```
val physsimFuture = Future{
  if (!beamConfig.beam.physsim.skipPhysSim){
    agentSimToPhysSimPlanConverter.startPhysSim(event, initialTravelTime.orNull)
  }
}
```

JustinPihony (approved):
Minor suggestion - otherwise lgtm

dimaopen:
Test!

Issue #3308, opened 2021-09-15
Metrics to measure health of simulation


nikolayilyin:
The problem is the following - often some problems are invisible for someone doing many parallel BEAM runs. 
For example: 
- a couple of died during simulation actors, 
- a critical error which did not end in simulation stuckness/unexpected termination

We need a script, probably a python script, to automatically run it after the simulation and analyze simulation output. 
Then this script output should be stored in the file.

The results of this script might be used in lambda function as part of notification about simulation completion (both in slack message and in google spreadsheet) 

So far let's calculate the following metrics:
- beamLog.out:  amount of all ERROR log entries
- beamLog.out:  amount of all WARN log entries
- beamLog.out:  amount of all stacktraces
- beamLog.out:  number of died actors
- beamLog.out:  number of dead letters

rajnikantsh:
closed via https://github.com/LBNL-UCB-STI/beam/pull/3314

PR #3314, opened 2021-09-22
Beam Health Analysis script

rajnikantsh:
Implementation of https://github.com/LBNL-UCB-STI/beam/issues/3308

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3314)

### DIFF
--- a/src/main/python/general_analysis/simulation_health_analysis.py
+++ b/src/main/python/general_analysis/simulation_health_analysis.py
@@ -1,0 +1,47 @@
+import os
+import sys
+from glob import glob
+
+
+detectors = {
+                "deadLetter": lambda line: "DeadLetter" in line,
+                "actor died" : lambda line: "terminated unexpectedly" in line,
+                "warn": lambda line: " WARN " in line, 
+                "error": lambda line: " ERROR " in line, 
+                "stacktrace": lambda line: line.startswith("\tat ")   
+            }
+beam_home = sys.argv[1]
+print(beam_home)
+os.chdir(beam_home)
+os.system("./gradlew :run -PappArgs=\"['--config', 'test/input/beamville/beam.conf']\"")
+
+log_file_location = glob(beam_home+"/output/beamville/*/beamLog.out")
+
+with open(log_file_location[0]) as file:
+    file = file.readlines()
+
+matric_log = {}
+stacktrace_count = 0
+for line in file:
+    
+    if detectors["stacktrace"](line):
+        stacktrace_count = stacktrace_count + 1
+        continue
+    elif stacktrace_count > 0:
+        matric = matric_log.get("stacktrace", [])
+        matric.append(stacktrace_count)
+        matric_log["stacktrace"] = matric
+        stacktrace_count = 0
+        continue
+
+    for key,value in detectors.items():
+        if value(line):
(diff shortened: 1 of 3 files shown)


src/main/python/general_analysis/simulation_health_analysis.py
```diff
@@ -0,0 +1,47 @@
...
+import sys
+from glob import glob
+
+
+detectors = {
+                "deadLetter": lambda line: "DeadLetter" in line,
+                "actor died" : lambda line: "terminated unexpectedly" in line,
+                "warn": lambda line: " WARN " in line, 
+                "error": lambda line: " ERROR " in line, 
+                "stacktrace": lambda line: line.startswith("\tat ")
```
JustinPihony: I'm not sure of a better solution off the top of my head, but this would likely count 10+ for each incident. Maybe note that in the output?

src/main/python/general_analysis/simulation_health_analysis.py
```diff
@@ -0,0 +1,47 @@
+import os
```
JustinPihony, line 1: This does not address the aspect of automatically running at the end of a simulation and/or posting to somewhere for review (maybe only if some count is too high?) At least we should document it - or add something to the end of the simulation to note how to run?

src/main/python/general_analysis/simulation_health_analysis.py
JustinPihony: I now see how you got around it. I think there should be a comment as to why it is handled separately from other items to make it so that other maintainers don't need to follow the logic to understand it. Also, there is a bug that probably isn't worth handling - if the last line of the file is a stack trace then you will be one short

src/main/python/general_analysis/simulation_health_analysis.py
rajnikantsh: hi @JustinPihony suppose beam failed before execution end. in that case we will be not able to send any notification. 

what about to add a log statement in Begining like.

```
if you want analysis this execution then run `python3 simulation_health_analysis.py` after execution complete
```

JustinPihony: I see - that works :)

rajnikantsh:
test!

JustinPihony (approved):

nikolayilyin:
test!

nikolayilyin (approved):
LGTM

nikolayilyin:
test!

src/main/python/general_analysis/simulation_health_analysis.py
```diff
@@ -0,0 +1,56 @@
...
+            matric = matric_log.get(key, [])
+            matric.append(line)
+            matric_log[key] = matric
+            break
+
+with open('RunHealthAnalysis.txt', 'w') as file:
+    for detector in detectors:
+        file.write(f"{detector},{len(matric_log.get(detector, []))}\n")
+
+token = os.environ.get('SLACK_TOKEN')
```
JustinPihony, line 48: I know this was merged, however I would suggest putting the slack notification in the lambda. Otherwise this will throw for local usage where there is no need for slack.

PR #3315, opened 2021-09-23
doc change

JustinPihony:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3315)


JustinPihony:
Force pushing because it is a doc change only

PR #3319, opened 2021-09-25
Parking zones at link level

haitamlaarabi:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319)


haitamlaarabi:
Test now!

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
```diff
@@ -440,10 +478,28 @@ object ParkingZoneFileUtils extends ExponentialLazyLogging {
...
+      val linkMaybe =
+        (if (locationXString == null || locationXString.isEmpty || locationYString == null || locationYString.isEmpty)
+           None
+         else
+           Some(new Coord(locationXString.toDouble, locationYString.toDouble))) match {
+          case Some(coord) if beamServices.isDefined =>
+            Some(
+              NetworkUtils.getNearestLink(beamServices.get.beamScenario.network, beamServices.get.geo.wgs2Utm(coord))
+            )
+          case None if beamServices.isDefined && reservedFor.managerType == VehicleManager.TypeEnum.Household =>
```
dimaopen: This branch never takes place because Some(..) is never None. Instead of pattern matching if-else can be used here.

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
```diff
@@ -440,10 +478,28 @@ object ParkingZoneFileUtils extends ExponentialLazyLogging {
...
         if (parkingZoneIdString == null || parkingZoneIdString.isEmpty) None
         else Some(ParkingZone.createId(parkingZoneIdString))
+      val linkMaybe =
+        (if (locationXString == null || locationXString.isEmpty || locationYString == null || locationYString.isEmpty)
+           None
+         else
+           Some(new Coord(locationXString.toDouble, locationYString.toDouble))) match {
+          case Some(coord) if beamServices.isDefined =>
+            Some(
+              NetworkUtils.getNearestLink(beamServices.get.beamScenario.network, beamServices.get.geo.wgs2Utm(coord))
```
dimaopen: We probably need to make this code GEO dependent. If GEO is Link then we can just use that link.

src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala
```diff
@@ -153,7 +153,9 @@ class ParkingFunctions[GEO: GeoLevel](
     parkingZone: ParkingZone[GEO],
     geoArea: GEO
   ): Coord = {
-    if (parkingZone.reservedFor.managerType == VehicleManager.TypeEnum.Household)
+    if (parkingZone.link.isDefined)
+      parkingZone.link.get.getCoord
+    else if (parkingZone.reservedFor.managerType == VehicleManager.TypeEnum.Household)
```
dimaopen, line 158: Generally GeoLvel.geoSampling for Link type returns the link coord here. Could we just use parking files on the link level?

dimaopen (commented):
We can try reusing the existing functionality that Beam already has. In this case, we need to provide a parking file on the link level.

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -143,6 +146,23 @@ class BeamVehicle(
     }
   }
 
+  def waitingToCharge(startTick: Int): Unit = {
```
JustinPihony, line 149: Can you rename the method name or the variable name? Having the same name as a variable and method is confusing.

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
```diff
@@ -235,23 +259,35 @@ object ChargingNetwork {
   }
 
   final case class ChargingStation(zone: ParkingZone[_]) {
-    import ConnectionStatus._
-    private val connectedVehiclesInternal = mutable.HashMap.empty[Id[BeamVehicle], ChargingVehicle]
+    import ChargingStatus._
+    private val chargingVehiclesInternal = mutable.HashMap.empty[Id[BeamVehicle], ChargingVehicle]
 
-    private val waitingLineInternal: mutable.PriorityQueue[ChargingVehicle] =
+    private val gracedVehiclesInternal = mutable.HashMap.empty[Id[BeamVehicle], ChargingVehicle]
```
JustinPihony: Is this an industry term? Otherwise it doesn't make itself immediately evidenced as far as I can tell.

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
```diff
@@ -401,6 +419,23 @@ object ParkingZoneFileUtils extends ExponentialLazyLogging {
...
+        .getAttribute(houseoldId, "homecoordx")
+        .asInstanceOf[Double]
+      val y = beamServices.matsimServices.getScenario.getHouseholds.getHouseholdAttributes
+        .getAttribute(houseoldId, "homecoordy")
+        .asInstanceOf[Double]
+      new Coord(x, y)
+    } match {
+      case Success(coord) => Some(coord)
+      case Failure(e) =>
+        logger.warn(s"Household without coordinate! $e")
```
JustinPihony, line 435: Should this be an error?

JustinPihony (commented):
Some minor comments - although I agree with the one sticking point about the unreachable None

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
haitamlaarabi: If you look at it carefully you will see that the match is for
`(if (locationXString == null || locationXString.isEmpty || locationYString == null || locationYString.isEmpty)
           None
         else
           Some(new Coord(locationXString.toDouble, locationYString.toDouble)))`
I will write it clearly to avoid the confusion

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
haitamlaarabi: How would you suggest to do that? because some of the parking stalls are link based but other not. In this scenario, mostly residential stalls are link, while rest are TAZ.

src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala
haitamlaarabi: as I explained above, what I tried to do is to make a hybrid form, where in TAZ GEO, some stalls are at link level. Would you suggest a better approach to this?

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
haitamlaarabi: Good call. Done!

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
haitamlaarabi: Well. it is called Grace period. When a vehicle is done charging but still plugged to the wall. In this case some charging network companies might charge you for that. Actually the grace period, is the period that allowed to stay plugged without being charged for a penalty. After the garce period ends, they start charging you. Would you suggest to change the name?

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
haitamlaarabi: Not really, since if the household doesn't have a coordinate, the stall won't be attached to that household.

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
dimaopen: You are right. I'm suggesting rewriting this part in the following way:

```
       val locationIsDefined = !StringUtils.isBlank(locationXString) && !StringUtils.isBlank(locationYString)
       val linkMaybe = if (locationIsDefined && beamServices.isDefined) {
        val coord = new Coord(locationXString.toDouble, locationYString.toDouble)
        Some(
          NetworkUtils.getNearestLink(beamServices.get.beamScenario.network, beamServices.get.geo.wgs2Utm(coord))
        )
      } else if (beamServices.isDefined && reservedFor.managerType == VehicleManager.TypeEnum.Household)
        getHouseholdLocation(beamServices.get, reservedFor.managerId.toString) map { homeCoord =>
          NetworkUtils.getNearestLink(
            beamServices.get.beamScenario.network,
            homeCoord
          )
        }
      else
        None
```

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
dimaopen: In this case, this is the most appropriate way to do this.

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
JustinPihony: I'm fine with it as is personally, after you note the subtle difference - except I think the if should be broken into a separate variable, which is then matched against. Or maybe use curly brackets, but I prefer the former. The reason is because this is so easy to mistake as was done already. If you break it apart then the match will be more evident to what it is matching. Beyond that, it is a matter of preference for yours versus Dima's suggestion.

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
JustinPihony: I do not see the change - please do not forget to push :)

src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala
JustinPihony: I will think on it, but I don't think there is anything better - becoming more aware of the domain is the solution for this really. So that is fine. Again, I'll think if there is a way to embed this directly into the name, but most of it is awareness.

zneedell: We could be super explicit with something like `vehiclesInGracePeriodAfterCharging`, maybe

haitamlaarabi:
Test Now!

haitamlaarabi:
Test Now!

haitamlaarabi:
Test Now!

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
```diff
@@ -440,10 +479,23 @@ object ParkingZoneFileUtils extends ExponentialLazyLogging {
       val pricingModel = PricingModel(pricingModelString, newCostInDollarsString)
       val timeRestrictions = parseTimeRestrictions(timeRestrictionsString)
       val chargingPoint = ChargingPointType(chargingTypeString)
-      val numStalls = calculateNumStalls(numStallsString.toDouble)
+      val numStalls = calculateNumStalls(numStallsString.toDouble, reservedFor)
       val parkingZoneIdMaybe =
         if (parkingZoneIdString == null || parkingZoneIdString.isEmpty) None
         else Some(ParkingZone.createId(parkingZoneIdString))
+      val linkMaybe = Option(!isBlank(locationXString) && !isBlank(locationYString)) match {
```
dimaopen:
```suggestion
      val linkMaybe = !isBlank(locationXString) && !isBlank(locationYString) match {
```

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
```diff
@@ -440,10 +479,23 @@ object ParkingZoneFileUtils extends ExponentialLazyLogging {
...
+        case Some(true) if beamServices.isDefined =>
```
dimaopen:
```suggestion
        case true if beamServices.isDefined =>
```

src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala
```diff
@@ -440,10 +479,23 @@ object ParkingZoneFileUtils extends ExponentialLazyLogging {
...
+          val coord = new Coord(locationXString.toDouble, locationYString.toDouble)
+          Some(NetworkUtils.getNearestLink(beamServices.get.beamScenario.network, beamServices.get.geo.wgs2Utm(coord)))
+        case Some(false) if beamServices.isDefined && reservedFor.managerType == VehicleManager.TypeEnum.Household =>
```
dimaopen:
```suggestion
        case false if beamServices.isDefined && reservedFor.managerType == VehicleManager.TypeEnum.Household =>
```

dimaopen (approved):
I suggested a little change, otherwise looks good

haitamlaarabi:
Test now!

haitamlaarabi:
Test Now!

haitamlaarabi (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319)* status: 0 of 39 files reviewed, 9 unresolved discussions (waiting on @dimaopen, @JustinPihony, and @zneedell)

___
*[src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala, line 149 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319#-MkQKsHCxKwGRKC2i6AV-r1-149:-MkxGgj-A9Ctj9XzyLVS:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/db7853dbe47c3a002ea043027aa837ab0624890f/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L149)):*

I do not see the change - please do not forget to push :)
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala, line 265 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319#-MkQKsHCxKwGRKC2i6AW-r1-265:-MkxGi2ZBscxsiLJOoYo:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/db7853dbe47c3a002ea043027aa837ab0624890f/src/main/scala/beam/agentsim/infrastructure/ChargingNetwork.scala#L265)):*
<details><summary><i>Previously, zneedell (Zach Needell) wrote…</i></summary><blockquote>

We could be super explicit with something like `vehiclesInGracePeriodAfterCharging`, maybe
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala, line 492 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319#-MkQKsHCxKwGRKC2i6Ad-r1-492:-MkxGnU_E76dmBUgp9IE:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/db7853dbe47c3a002ea043027aa837ab0624890f/src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala#L492)):*

In this case, this is the most appropriate way to do this.
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala, line 494 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319#-MkQKsHCxKwGRKC2i6Ad-r1-494:-MkxGmxp4fF2ijFqQm9w:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/db7853dbe47c3a002ea043027aa837ab0624890f/src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala#L494)):*

I'm fine with it as is personally, after you note the subtle difference - except I think the if should be broken into a separate variable, which is then matched against. Or maybe use curly brackets, but I prefer the former. The reason is because this is so easy to mistake as was done already. If you break it apart then the match will be more evident to what it is matching. Beyond that, it is a matter of preference for yours versus Dima's suggestion.
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala, line 486 at r4](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319#-MkQKsHCxKwGRKC2i6Ad-r4-486:-MkxGkC-60at3huw__I5:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/d5dbae6337bb608405f1003285b9eb26d97df933/src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala#L486)):*

```suggestion
      val linkMaybe = !isBlank(locationXString) && !isBlank(locationYString) match {
```
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala, line 487 at r4](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319#-MkQKsHCxKwGRKC2i6Ad-r4-487:-MkxGlAG9ksgSrnZtCpV:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/d5dbae6337bb608405f1003285b9eb26d97df933/src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala#L487)):*

```suggestion
        case true if beamServices.isDefined =>
```
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala, line 490 at r4](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3319#-MkQKsHCxKwGRKC2i6Ad-r4-490:-MkxGnjk0YNWoPjyfC0p:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/d5dbae6337bb608405f1003285b9eb26d97df933/src/main/scala/beam/agentsim/infrastructure/parking/ParkingZoneFileUtils.scala#L490)):*

```suggestion
        case false if beamServices.isDefined && reservedFor.managerType == VehicleManager.TypeEnum.Household =>
```
</blockquote></details>

Done.

zneedell (approved):
LGTM

haitamlaarabi:
Test Now!

src/main/scala/beam/router/gtfs/FareUtils.scala
```diff
@@ -0,0 +1,29 @@
+package beam.router.gtfs
+
+import beam.utils.FileUtils
+import com.typesafe.scalalogging.StrictLogging
+
+import java.io.FileFilter
+import java.util.zip.ZipFile
+
+object FareUtils extends StrictLogging {
```
JustinPihony: Is this used anywhere? I don't see this change utilizing it

haitamlaarabi: FareCalculator.scala , line 103.
for some reason that file got untracked and then I added it again.

haitamlaarabi:
Test now!

PR #3320, opened 2021-09-28
Add searched modes to GH and native CCH result

dimaopen:
Add searched modes to GH and native CCH routing responces in order to have RoutingStatistic works correct.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3320)

### DIFF
--- a/src/main/scala/beam/router/RoutingWorker.scala
+++ b/src/main/scala/beam/router/RoutingWorker.scala
@@ -385,9 +385,15 @@ class RoutingWorker(workerParams: R5Parameters) extends Actor with ActorLogging 
       val definedResponses = responses.flatten
       (definedResponses, r5ResponseOption) match {
         case (head +: _, Some(r5Resp)) =>
-          head.copy(itineraries = definedResponses.flatMap(_.itineraries) ++ r5Resp.itineraries)
+          head.copy(
+            itineraries = definedResponses.flatMap(_.itineraries) ++ r5Resp.itineraries,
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/router/RoutingWorker.scala
```diff
@@ -385,9 +385,15 @@ class RoutingWorker(workerParams: R5Parameters) extends Actor with ActorLogging
       val definedResponses = responses.flatten
       (definedResponses, r5ResponseOption) match {
         case (head +: _, Some(r5Resp)) =>
-          head.copy(itineraries = definedResponses.flatMap(_.itineraries) ++ r5Resp.itineraries)
+          head.copy(
+            itineraries = definedResponses.flatMap(_.itineraries) ++ r5Resp.itineraries,
+            searchedModes = definedResponses.map(_.searchedModes).reduce(_ ++ _) ++ r5Resp.searchedModes
```
JustinPihony: reduce(_ ++ _) is the same as flatten. I would prefer flatten because it is slightly more obvious.

JustinPihony (approved):
Just a minor improvement for readability (at least in my opinion)

src/main/scala/beam/router/RoutingWorker.scala
dimaopen: Implemented your suggestion, but I used flatMap instead of map..flatten as my IDE suggested. And because reduce(_ ++ _) produces a set I also swapped operators of ++

dimaopen:
Test!

alex-vv (approved):

dimaopen:
Test!

PR #3338, opened 2021-10-15
Default vehicle for car leg

zneedell:
Fix so that agents don't have to walk if they have a car trip in their plans but no access to a car.

`HouseholdFleetManager` can now have a default vehicle type, and `MobilityStatusInquiry` now has the option of demanding that a vehicle of a certain category be returned. If the demanded category matches the fleet manager's default vehicle type and no other vehicle are available, it creates a vehicle, parks it, and returns it to the sender of the inquiry.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3338)


src/test/scala/beam/agentsim/agents/PersonWithPersonalVehiclePlanSpec.scala
```diff
@@ -562,6 +562,122 @@ class PersonWithPersonalVehiclePlanSpec
       expectMsgType[CompletionNotice]
     }
 
+    it("should create a last resort car if told to drive but no cars are available") {
```
JustinPihony, line 565: Other than this description the test is unclear on how this is being verified. Could you rename some of the variables and possibly the expectations so that the reader of the test doesn't have to go tracing into the code unless required. Tests not only protect against regressions, but help document.

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -193,10 +197,40 @@ object HouseholdActor {
       case TriggerWithId(InitializeTrigger(tick), triggerId) =>
         val vehiclesByCategory =
           vehicles.filter(_._2.beamVehicleType.automationLevel <= 3).groupBy(_._2.beamVehicleType.vehicleCategory)
-        val fleetManagers = vehiclesByCategory.map { case (category, vs) =>
+        //WE NEED TO MAKE  A FLEET MANAGER EVEN IF THERE ARE NO CARSs
+
+        val vehiclesByAllCategories = List(Car, Bike)
```
JustinPihony, line 202: Should the household actor define the types which it will auto-create these vehicles or should it be fed in somewhere - should it be a config? Basically, should this be hard-coded?

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -193,10 +197,40 @@ object HouseholdActor {
...
+          .map(cat => cat -> vehiclesByCategory.getOrElse(cat, Map[Id[BeamVehicle], BeamVehicle]()))
```
JustinPihony: I believe the preference is to use `.empty`

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
JustinPihony: Or this should be a centralized location where you can call something like `AllCategories`. As I see it there is a chance that something could change where a category could be lost, because the `vehicles` list has some other category, and we "say" this is all categories, but it doesn't even merge with the original `vehiclesByCategory` - it essentially drops it in favor of this new creation (which yes wraps the original...but doesn't use it from the outside).

On top of the idea to create a central place where we can get all categories and keep it up to date, the code for the above could be more akin to:

```
val defaultMapping = List(Car, Bike).map(cat => cat -> Map.empty[Id[BeamVehicle], BeamVehicle])
val vehiclesByAllCategories = defaultMapping ++ vehiclesByCategory
```

where the values on the left overwrite any duplicates from the right side. So, it uses the categories already present and only adds in the default empty if it didn't already exist.

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -193,10 +197,40 @@ object HouseholdActor {
...
+                )
+              )
+            case VehicleCategory.Bike =>
+              beamServices.beamScenario.vehicleTypes.get(
+                Id.create(
+                  beamServices.beamConfig.beam.agentsim.agents.vehicles.dummySharedBike.vehicleTypeId,
+                  classOf[BeamVehicleType]
+                )
+              )
+            case _ => None
```
JustinPihony: Should this log a warning if a type is unexpected?

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -98,16 +109,65 @@ class HouseholdFleetManager(
...
+            )
+            logger.warn(
+              s"No vehicles available for category ${defaultBeamVehicleType.vehicleCategory} available for person ${personId.toString}, creating a new vehicle with id ${vehicle.id.toString}"
+            )
+
+            responseFuture.collect { case ParkingInquiryResponse(stall, _, otherTriggerId) =>
+              vehicle.useParkingStall(stall)
+              logger.debug("Vehicle {} is now taken, which was just created", vehicle.id)
+              vehicle.becomeDriver(mobilityRequester)
+              availableVehicles = List[BeamVehicle]()
```
JustinPihony: Modifying actor variables within a Future is dangerous. Isn't it the scenario where this empty state is already the state. It could be possible that a vehicle become available in the time between inquiry and response? I don't see why it needs set in this portion of the logic.

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -98,16 +109,65 @@ class HouseholdFleetManager(
...
+          }
+        case _ =>
+          availableVehicles = availableVehicles match {
+            case firstVehicle :: rest =>
+              logger.debug("Vehicle {} is now taken", firstVehicle.id)
+              firstVehicle.becomeDriver(sender)
+              sender() ! MobilityStatusResponse(Vector(ActualVehicle(firstVehicle)), triggerId)
+              rest
+            case Nil =>
+              logger.debug(s"Not returning vehicle because no default is defined")
```
JustinPihony, line 166: Should/Will other fleet managers be attempted? Since the householdactor creates a manager for each type.

JustinPihony (commented):
I am not approving or rejecting as it is still listed as WIP

zneedell:
Completed via #3456

PR #3340, opened 2021-10-18
In-simulation charging scale up

haitamlaarabi:
It is basically the duplication of charging events to match a 100% population run in a scenario that runs a small sample of the population. The non-linear behavior of charging is captured by the Parking/Charging MNL model, as it processes the virtual charging inquiries and therefore compete with the actual simulated agents. To calibrate the in-simulating scaling model, one would need to set the expansionFactor_X values (>= 1) and validate those values by comparing the load shape against a 100% run or a sufficiently large scenario. One also would need to set the chargingPointCountScalingFactor, chargingPointCostScalingFactor, and can have a completely separate parking files for stalls with charging points in: chargingPointFilePath.
```
beam.agentsim.chargingNetworkManager {
  timeStepInSeconds = "int | 300"
  chargingPointCountScalingFactor = "double | 1.0"
  chargingPointCostScalingFactor = "double | 1.0"
  chargingPointFilePath = ""
  scaleUp {
    enabled = "boolean | false"
    expansionFactor_home_activity = "double | 1.0"
    expansionFactor_work_activity = "double | 1.0"
    expansionFactor_charge_activity = "double | 1.0"
    expansionFactor_init_activity = "double | 1.0"
    expansionFactor_wherever_activity = "double | 1.0"
  }
  ...
}
```
initial results:
public-charging-by-scenario

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3340)


haitamlaarabi:
Test!

haitamlaarabi:
Test now!

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
```diff
@@ -0,0 +1,330 @@
...
+  private lazy val virtualParkingInquiries: TrieMap[Int, ParkingInquiry] = TrieMap()
+  private lazy val vehicleRequests = mutable.HashMap.empty[(Id[TAZ], ParkingActivityType), List[VehicleRequestInfo]]
+
+  /**
+    * @param activityType ParkingActivityType
+    * @return
+    */
+  private def scaleUpFactor(activityType: ParkingActivityType): Double = {
+    val factor = activityType match {
+      case _ if !cnmConfig.scaleUp.enabled => 1.0
```
JustinPihony: Personal preference, change it to an if else:

```
if(!cnmConfig.scaleUp.enabled) 0.0
else {
  val factor = activityType match {
    //cases
  }
  Math.max(factor - 1.0, 0.0)
}
```

JustinPihony: Also, consider changing it to a val of type Map (versus a def), but only if you expect it to be called a lot. No need to optimize prematurely

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
```diff
@@ -0,0 +1,330 @@
...
+      case _                               => cnmConfig.scaleUp.expansionFactor_wherever_activity
+    }
+    Math.max(factor - 1.0, 0.0)
+  }
+
+  override def loggedReceive: Receive = {
+    case t @ TriggerWithId(PlanParkingInquiryTrigger(_, requestId), triggerId) =>
+      log.debug(s"Received parking response: $t")
+      sender ! CompletionNotice(triggerId)
+      self ! virtualParkingInquiries(requestId)
```
JustinPihony: This will blow up if the value is not present. It should probably use `get` and/or handle the empty case appropriately.

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
```diff
@@ -0,0 +1,330 @@
...
+    case t @ TriggerWithId(PlanChargingUnplugRequestTrigger(tick, beamVehicle, requestId), triggerId) =>
+      log.debug(s"Received parking response: $t")
+      sender ! CompletionNotice(triggerId)
+      self ! ChargingUnplugRequest(tick, beamVehicle, triggerId)
```
JustinPihony, line 71: Should these be completed? And/Or should it schedule the next request so that it can be made sure to completion?

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
```diff
@@ -0,0 +1,330 @@
...
+    case reply @ UnpluggingVehicle(_, _, _) =>
+      log.debug(s"Received parking response: $reply")
+  }
+
+  /**
+    * Next Time poisson
+    * @param rate rate of charging event
+    * @return
+    */
+  private def nextTimePoisson(rate: Double): Double = 3600.0 * (-Math.log(1.0 - rand.nextDouble()) / rate)
```
JustinPihony: Should we name this next time step? This seems like technical creep entering what should be domain logic.

src/main/scala/beam/agentsim/infrastructure/power/PowerController.scala
```diff
@@ -1,33 +1,33 @@
...
+    chargingNetworkHelper.allChargingStations
+  ).value
+
   private[power] lazy val beamFederateOption: Option[BeamFederate] =
     if (isConnectedToHelics) {
       logger.warn("ChargingNetworkManager should be connected to a grid via Helics...")
       val helicsConfig = beamConfig.beam.agentsim.chargingNetworkManager.helics
       Try {
-        logger.debug("Init PowerController resources...")
+        logger.info("Init PowerController resources...")
```
JustinPihony, line 30: Should this remain bumped to info?

JustinPihony (commented):
I did not review the R and Python files - so it depends on comfort levels there if you want somebody else to review those.

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
haitamlaarabi: Okay

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
haitamlaarabi: Good point!

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
haitamlaarabi: okay

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
haitamlaarabi: okay

src/main/scala/beam/agentsim/infrastructure/power/PowerController.scala
haitamlaarabi: Yes. Inform that helics connection is working!

haitamlaarabi:
Test now!

haitamlaarabi:
Test now!

haitamlaarabi:
Test now!

haitamlaarabi:
Test now!

haitamlaarabi:
Test now!

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
```diff
@@ -0,0 +1,349 @@
...
+
+  private lazy val defaultScaleUpFactor: Double =
+    if (!cnmConfig.scaleUp.enabled) 1.0 else cnmConfig.scaleUp.expansionFactor_wherever_activity
+
+  override def loggedReceive: Receive = {
+    case t @ TriggerWithId(PlanParkingInquiryTrigger(_, requestId), triggerId) =>
+      log.debug(s"Received parking response: $t")
+      self ! virtualParkingInquiries.getOrElse(
+        requestId,
+        log.error(
```
JustinPihony: This will send a Unit back to self which will go to DeadLetters. Since this is logged here anyway then I'm ok with it - more of a note in case you wanted to adjust in any way.

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
JustinPihony: As long as ParkingInquiry is the right place to Complete then this is fine (as I saw you removed the older two completes and changed to be done at the ParkingInquiryResponse

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
JustinPihony: I like the new name having both aspects. Well done!

JustinPihony (approved):

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
haitamlaarabi: Fixed!

haitamlaarabi:
Test now!

PR #3345, opened 2021-10-25
Unit test for round uniformly

dimaopen:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3345)

### DIFF
--- a/src/test/scala/beam/utils/MathUtilsSpec.scala
+++ b/src/test/scala/beam/utils/MathUtilsSpec.scala
@@ -1,0 +1,38 @@
+package beam.utils
+
+import org.scalatest.matchers.should.Matchers
+import org.scalatest.wordspec.AnyWordSpecLike
+
+import scala.util.Random
+
+/**
+  * @author Dmitry Openkov
+  */
+class MathUtilsSpec extends AnyWordSpecLike with Matchers {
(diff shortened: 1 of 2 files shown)


src/test/scala/beam/utils/MathUtilsSpec.scala
```diff
@@ -0,0 +1,38 @@
...
+
+import scala.util.Random
+
+/**
+  * @author Dmitry Openkov
+  */
+class MathUtilsSpec extends AnyWordSpecLike with Matchers {
+  "round uniformly" when {
+    "receive a positive double value" should {
+      "round it to the closest integer most of the time (proportionally to the fraction)" in {
```
JustinPihony, line 14: Is this built strong enough such that it won't cause flaky tests? Random makes for some hard to control tests...

JustinPihony (approved):
Other than possibly introducing a flaky test - lgtm

src/test/scala/beam/utils/MathUtilsSpec.scala
dimaopen: It's not going to cause flaky tests because it uses seeding and a strict sequence of random numbers.

PR #3348, opened 2021-10-27
Scala 2.13 migration

### LINKED ISSUE #2505: Project: Update the project to Scala 2.13.1
Scala 2.13 is here for long enough, so we [can migrate it](https://github.com/scala/scala/releases/tag/v2.13.1)

JustinPihony:
Resolves #2505

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3348)


src/main/java/beam/utils/GroupableUtil.scala
```diff
@@ -1,25 +0,0 @@
-package beam.utils
```
JustinPihony, line 1: `CanBuildFrom` removed the empty `apply` - and after looking for usage I only found [the original commit here](https://github.com/LBNL-UCB-STI/beam/commit/e0e151034574953f6906ef16fc6ddc91adc4e970). Which did not have any actual usage. I'm flagging it here as a possible public API for use in a routing framework in case it is worth fixing.

src/main/scala/beam/agentsim/events/handling/BeamEventsWriterParquet.scala
```diff
@@ -84,7 +84,7 @@ class BeamEventsWriterParquet(
       .build()
   }
 
-  override protected def writeEvent(event: Event): Unit = {
```
JustinPihony, line 87: This is due to the warning:

> weaker access privileges in overriding
> def ... (defined in class BeamEventsWriterBase)
>   override should be public

And I wanted to point it out.

This was done in BeamEventsWriterXML also.

src/main/scala/beam/router/ParquetRequester.scala
```diff
@@ -83,7 +85,7 @@ object ParquetRequester extends BeamHelper with LazyLogging {
 
     val requests = requestRecords.map { req =>
       val reqJsonStr = new String(req.get("requestAsJson").asInstanceOf[Utf8].getBytes, StandardCharsets.UTF_8)
-      io.circe.parser.parse(reqJsonStr).right.get.as[RoutingRequest].right.get
+      io.circe.parser.parse(reqJsonStr).getOrElse(throw new NoSuchElementException("Either.right.get on Left")).as[RoutingRequest].getOrElse(throw new NoSuchElementException("Either.right.get on Left"))
```
JustinPihony: `get` is deprecated. I opted to keep the old implementation of throwing to use a left

JustinPihony: An alternative would be to eat the exception and do a match, returning empty for a left?

src/main/scala/beam/agentsim/agents/household/FastHouseholdCAVScheduling.scala
```diff
@@ -430,7 +430,7 @@ object HouseholdTripsHelper {
     var firstPickupOfTheDay: Option[MobilityRequest] = None
     breakable {
       householdPlans.foldLeft(householdNbOfVehicles) { case (counter, plan) =>
-        val usedCarOut = plan.trips.sliding(2).foldLeft(false) { case (usedCar, Seq(prevTrip, curTrip)) =>
+        val usedCarOut = plan.trips.sliding(2).map(slide => (slide.head, slide.last)).foldLeft(false) { case (usedCar, (prevTrip, curTrip)) =>
```
JustinPihony: This appeases an exhaustiveness warning as error. That said, it will blow up if there is no head or tail...however it already is the case - except it would blow up as a match miss. So, the question is if we want to keep the exceptions or handle the empty and single trip cases?

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -396,9 +396,9 @@ class BeamVehicle(
     val mode = beamVehicleType.vehicleCategory match {
       case Bike =>
         BIKE
-      case Car | LightDutyTruck | HeavyDutyTruck if isCAV =>
+      case Car | LightDutyTruck | MediumDutyPassenger | HeavyDutyTruck if isCAV =>
```
JustinPihony, line 398: Somebody please verify this - I'm surprised it wasn't caught as a match miss before the 2.13 change. But I believe they explicitly worked on improving the compiler in this regard.

src/main/scala/beam/sim/config/MatSimBeamConfigBuilder.scala
```diff
@@ -113,6 +113,7 @@ class MatSimBeamConfigBuilder(beamConf: Config) extends LazyLogging {
           case None =>
             logger.warn(s"MATSim module '$moduleName' not found")
         }
+      case(moduleSplit, value) => logger.warn(s"Unexpected module split $moduleSplit in MatsimBeamConfigBuilder. Value is $value")
```
JustinPihony: I'm unsure if this is possible as I did not dig into the config splitpath code, but it seemed possible, so maybe the warning is fine and we can adjust if needed?

src/main/scala/beam/router/RoutingWorker.scala
```diff
@@ -397,6 +397,7 @@ class RoutingWorker(workerParams: R5Parameters) extends Actor with ActorLogging
         case (Seq(), Some(r5Resp)) =>
           r5Resp
         case (Seq(), None) => r5.calcRoute(request)
+        case (_, _) => ???//temporarily in place to appease the 2.13 compiler - Possibly https://github.com/scala/bug/issues/12144
```
JustinPihony: I have asked the Scala team about this one and will update the code if verified, or with a new issue if it is found to be one

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
Please check the dependency upgrades also - as there were a larger number than expected.

PR #3367, opened 2021-11-17
Xuan/#3330 additional parameters for links

Xuan-1998:
Adding additional parameters for links.
First stage is to add parameters to all the links.
Now it can achieve adding parameters for different categories of links seperately.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3367)


src/main/resources/beam-template.conf
```diff
@@ -442,6 +442,9 @@ beam.physsim.pickUpDropOffAnalysis.secondsFromPickUpPropOffToAffectTravelTime =
 beam.physsim.pickUpDropOffAnalysis.additionalTravelTimeMultiplier = "double | 1.0"
 
 # JDEQSim
+beam.physsim.jdeqsim.parameters.alpha = "double | 1.0"
+beam.physsim.jdeqsim.parameters.beta = "double | 2.0"
```
nikolayilyin: we need either self-explanatory names for parameters or a good comments about them, 
also a good idea is to put a comment with a range of possible values

Xuan-1998: Good point, thank you! Will add comments after we figure out the range of possible values.

Xuan-1998:
Test!

src/main/scala/beam/physsim/jdeqsim/JDEQSimRunner.scala
```diff
@@ -345,7 +350,7 @@ object JDEQSimRunner {
               val ftt = link.getLength / link.getFreespeed(time)
               if (volume >= minVolumeToUseBPRFunction) {
                 val tmp = volume / (link.getCapacity(time) * flowCapacityFactor)
-                val originalTravelTime = ftt * (1 + tmp * tmp)
+                val originalTravelTime = ftt * ftt * (1 + tmp * tmp)
```
zneedell: I think you want to do the same thing down here that you did up there

zneedell (dismissed):
LGTM besides the one comment!

src/main/scala/beam/physsim/jdeqsim/JDEQSimRunner.scala
Xuan-1998: fixed, thanks!

src/main/resources/beam-template.conf
Xuan-1998: abadonded this version, instead we have alpha and beta for different link categories now.

Xuan-1998:
Test!

src/main/scala/beam/physsim/jdeqsim/JDEQSimRunner.scala
```diff
@@ -330,9 +331,11 @@ object JDEQSimRunner {
                   caccSettings.roadCapacityAdjustmentFunction.getCapacityWithCACCPerSecond(link, caccShare, time)
                 //volume is calculated as number of vehicles entered the road per hour
                 //capacity from roadCapacityAdjustmentFunction is number of vehicles per second
-                val tmp = volume / (capacity * 3600)
-                val result = ftt * (1 + tmp * tmp)
 
+                val alpha = link.getAttributes.getAttribute("alpha").toString.toDouble
+                val beta = link.getAttributes.getAttribute("beta").toString.toDouble
```
dimaopen, line 336: I think we could use `.asInstanceOf[Double]` instead of formatting a Double and parsing it back.

src/main/scala/beam/router/r5/NetworkCoordinator.scala
```diff
@@ -219,6 +231,8 @@ object NetworkCoordinator {
       val speeds = getSpeeds(highwayType)
       val capacities = getCapacities(highwayType)
       val lanes = getLanes(highwayType)
+      val alphas = getAlphas(highwayType)
+      val betas = getBetas(highwayType)
```
dimaopen, line 235: Where do these alphas and betas go? It seems like they are not used anywhere.

dimaopen (changes requested):
Now alpha and beta can be read from a file defined by this param `beam.physsim.overwriteLinkParamPath` for each individual link. We also need to load these params from `beam.physsim.network.overwriteRoadTypeProperties `

src/main/scala/beam/router/r5/NetworkCoordinator.scala
Xuan-1998: Good point, deleted.

src/main/scala/beam/physsim/jdeqsim/JDEQSimRunner.scala
Xuan-1998: Good catch, adjusted

dimaopen (commented):
We need to somehow get values from the config file: `beam.physsim.network.overwriteRoadTypeProperties`. You can do it using your new methods.
      val alphas = getAlphas(highwayType)
      val betas = getBetas(highwayType)

Xuan-1998:
> We need to somehow get values from the config file: `beam.physsim.network.overwriteRoadTypeProperties`. You can do it using your new methods. val alphas = getAlphas(highwayType) val betas = getBetas(highwayType)


Very critical! Done by `createHighwaySetting`!

dimaopen (approved):
Generally looks good, though we still don't update links basing on config parameters. To do this we need to change beam-utils library. It's better to discuss this with @wrashid or @zneedell

Xuan-1998:
> Generally looks good, though we still don't update links basing on config parameters. To do this we need to change beam-utils library. It's better to discuss this with @wrashid or @zneedell

Thanks Dima! Let’s discuss it today

Xuan-1998:
Test!

PR #3370, opened 2021-11-19
Initial State of Charge for beam vehicles

### LINKED ISSUE #3369: Initial State of Charge for beam vehicles

dimaopen:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370)


src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -344,18 +347,19 @@ trait BeamHelper extends LazyLogging {
   def privateVehicles(
     beamConfig: BeamConfig,
     vehicleTypes: Map[Id[BeamVehicleType], BeamVehicleType]
-  ): TrieMap[Id[BeamVehicle], BeamVehicle] =
+  ): (TrieMap[Id[BeamVehicle], BeamVehicle], mutable.Map[Id[BeamVehicle], Double]) =
```
alex-vv: have you considered adding initial state of charge to `BeamVehicle` itself or if it is not possible maybe to create a wrapper for `BeamVehicle` with `Option[Double]` for SoC using either a case class or a tuple + custom type. I think it would be easier to work with than using two separate maps and joining them later.

dimaopen: I thought about it. But the initial SoC is mostly related to the scenario and to the vehicle state. Maybe I'm wrong here.

haitamlaarabi (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @alex-vv, @JustinPihony, and @nikolayilyin)

___
*[src/main/scala/beam/sim/BeamHelper.scala, line 350 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-MosuzS505uj3K2OSbWY-r1-350:-MpFEmDjAB6ngDunm-vK:b1yzlyi) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/dc119924827ad53d5ae7d15c98dcb70f54d5a6c9/src/main/scala/beam/sim/BeamHelper.scala#L350)):*

I thought about it. But the initial SoC is mostly related to the scenario and to the vehicle state. Maybe I'm wrong here.
</blockquote></details>

Here is how things should be:
1) At iteration 0 the SoC of a vehicle should be random when vehicles.csv is not provided, or it doesn't have a column SoC, otherwise we use the SoC from the file.
2) At iteration i > 0, we initialize the SoC of a vehicle from the last SoC of the same vehicle in the previous iteration. 

it makes sense to not change SoC of vehicle in BeamHelp, but instead in HouseholdActor (I think), because we need to initialize the SOC every iteration.
Is that what the code is doing here @dimaopen ?

dimaopen:
@haitamlaarabi 
1. Correct with a little exception: if SoC is not provided then electric vehicles are initialized with a random SoC that is around the value taken from `beam.agentsim.agents.vehicles.meanPrivateVehicleStartingSOC` param. 
2. It's not correct. The same initialization happens on each iteration (starting from iteration 0). Do you mean that we need to remember the initial SoC from iteration 0 and reuse it to initialize the vehicles at all subsequent iterations? I doubt that we need to use the last SoC (SoC at the end of iteration) to initialize the vehicles.

haitamlaarabi (commented):
1. Great
2. I think we should initialize vehicles in iteration i+1 from SoC of the end of iteration i. The objective is to reach energy conservation. If we do that after many iterations we will have the SoC at the beginning of an iteration equal to the end of the iteration.

> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-:-MpIdw7E1JvM5TW77_zA:bmzj5gf)* status: 0 of 15 files reviewed, 1 unresolved discussion (waiting on @alex-vv, @JustinPihony, and @nikolayilyin)

dimaopen:
@haitamlaarabi I added a new parameter for that: initializeSocFromEndOfIteration. If set to true then BEAM uses the final SoC from the previous iteration to initialize SoC at the current iteration.

dimaopen:
Test!

haitamlaarabi (changes requested):
Can I suggest a different approach? instead of initializeSocFromEndOfIteration, we use actually something like terminateAtRideHailFleetStoredElectricityConvergence. see class TerminateAtRideHailFleetStoredElectricityConvergence and RideHailFleetStoredElectricityEventTracker.
These classes decide when to stop, when difference between beginning and end iteration is close enough. I think also it is not activated, because there are no RideHailFleetStoredElectricityEvent events being thrown. 
Can we rearrange thing to take advantage of this code? and in the same time activate it for Ridehail.

in ridehail we initialize with: val initialStateOfCharge = BeamVehicle.randomSocFromUniformDistribution(rand, vehicleType, meanSoc)
but we should do the same as you did for PEV i.e. initialize from end of iteration. 

thanks

> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-:-Mpnmum_B3o23DnDkIHW:bsu2092)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @alex-vv, @JustinPihony, and @nikolayilyin)

dimaopen:
> in ridehail we initialize with: val initialStateOfCharge = BeamVehicle.randomSocFromUniformDistribution(rand, vehicleType, meanSoc)
> but we should do the same as you did for PEV i.e. initialize from end of iteration.

We can do this with this parameter
`beam.agentsim.agents.rideHail.linkFleetStateAcrossIterations`

dimaopen:
Implemented sending `RideHailFleetStoredElectricityEvent` for ride-hail fleet as the name of event implies. In order to get it working, you can use the following params```

beam.sim.termination.criterionName = "TerminateAtRideHailFleetStoredElectricityConvergence"
beam.sim.termination.terminateAtRideHailFleetStoredElectricityConvergence.minLastIteration = 0
beam.sim.termination.terminateAtRideHailFleetStoredElectricityConvergence.maxLastIteration = 5
beam.agentsim.agents.rideHail.linkFleetStateAcrossIterations = true
```

I left initializeSocFromEndOfIteration because it allows reinitializing the private electric vehicles.

haitamlaarabi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370)* status: 0 of 24 files reviewed, 6 unresolved discussions (waiting on @alex-vv, @dimaopen, @JustinPihony, and @nikolayilyin)

___
*[src/main/scala/beam/agentsim/agents/Population.scala, line 170 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-MqHTtGh-AXTOq2Y4DoS:-MqHTtGh-AXTOq2Y4DoT:b-ykgxvp) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/agentsim/agents/Population.scala#L170)):*
> ```Scala
>       .foldLeft(0.0, 0.0) { case ((fuelLevel, fuelCapacity), vehicle) =>
>         (
>           fuelLevel + vehicle.primaryFuelLevelInJoules,
> ```

At this level, did we already initialize vehicle fuel level, either with uniform distribution or through a vehicles.csv ?
___
*[src/main/scala/beam/agentsim/infrastructure/DefaultRidehailFunctions.scala, line 211 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-Mr5d5i96iTimhdCqD-J:-Mr5d5i96iTimhdCqD-K:b-mxtysi) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/agentsim/infrastructure/DefaultRidehailFunctions.scala#L211)):*
> ```Scala
>   def hasHighSocAndZoneIsDCFast(beamVehicle: BeamVehicle, parkingZone: ParkingZone[GEO]): Boolean = {
>     val soc = beamVehicle.getStateOfCharge
>     soc >= 0.8 && parkingZone.chargingPointType.exists(_.asInstanceOf[CustomChargingPoint].installedCapacity > 20.0)
> ```

can we make both 0.8 and 20.0 parametrizable ?  I think we use them both in different places in BEAM
___
*[src/main/scala/beam/sim/BeamHelper.scala, line 357 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-Mr5du_0-RonuinOKTsn:-Mr5du_0-RonuinOKTso:b-aq011t) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/sim/BeamHelper.scala#L357)):*
> ```Scala
>         vehicleTypes,
>         beamConfig.matsim.modules.global.randomSeed,
>         VehicleManager.AnyManager.managerId
> ```

This is tricky. I just realized that here is shouldn't be VehicleManager.AnyManager.managerId, but instead: VehicleManager.createOrGetReservedFor(household.getId.toString, VehicleManager.TypeEnum.Household)
as in Population.scala line 122
___
*[src/test/scala/beam/utils/scenario/InitialSocLoadingTest.scala, line 29 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-Mr5j5JIAV_Rp7YI8WQJ:-Mr5j5JIAV_Rp7YI8WQK:bj95gs7) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/test/scala/beam/utils/scenario/InitialSocLoadingTest.scala#L29)):*
> ```Scala
>   lazy implicit val system: ActorSystem = ActorSystem("soc_loading", config)
> 
>   "Scenario loader" must {
> ```

Can you also add a test class that would check if the state of charge of private EVs and Ridehail are actually converging to a value (difference between beginning and end state of charge) lower than the parameter "relativeTolerance" in old class TerminateAtRideHailFleetStoredElectricityConvergence
___
*[src/main/scala/beam/agentsim/events/FleetStoredElectricityEvent.scala, line 38 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-MqHUwtTEH4x9eTYLcAT:-MqHUwtTEH4x9eTYLcAU:b-g435k8) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/agentsim/events/FleetStoredElectricityEvent.scala#L38)):*
> ```Scala
>     attributes.put(ATTRIBUTE_FLEET_ID, fleetId)
>     attributes.put(ATTRIBUTE_STORAGE_CAPACITY_IN_JOULES, storageCapacityInJoules.toString)
>     attributes.put(ATTRIBUTE_STORAGE_CAPACITY_IN_JOULES, storageCapacityInJoules.toString)
> ```

Did you mean to put a different attribute than ATTRIBUTE_STORAGE_CAPACITY_IN_JOULES ?

haitamlaarabi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370)* status: 0 of 24 files reviewed, 6 unresolved discussions (waiting on @alex-vv, @dimaopen, @JustinPihony, and @nikolayilyin)

___
*[src/test/scala/beam/utils/scenario/InitialSocLoadingTest.scala, line 29 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-Mr5j5JIAV_Rp7YI8WQJ:-Mr5k5wLEmBnr9UHOvFU:b-dzfrtx) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/test/scala/beam/utils/scenario/InitialSocLoadingTest.scala#L29)):*

Can you also add a test class that would check if the state of charge of private EVs and Ridehail are actually converging to a value (difference between beginning and end state of charge) lower than the parameter "relativeTolerance" in old class TerminateAtRideHailFleetStoredElectricityConvergence
</blockquote></details>

Also maybe there is a test class already there that came with the MARAIN patch

dimaopen (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370)* status: 0 of 24 files reviewed, 6 unresolved discussions (waiting on @alex-vv, @haitamlaarabi, @JustinPihony, and @nikolayilyin)

___
*[src/main/scala/beam/agentsim/agents/Population.scala, line 170 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-MqHTtGh-AXTOq2Y4DoS:-Mr6EhrN0lCZTes0j6m7:b-bq30d9) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/agentsim/agents/Population.scala#L170)):*

At this level, did we already initialize vehicle fuel level, either with uniform distribution or through a vehicles.csv ?
</blockquote></details>

Yes, vehicle SoC initialization happens at the iteration starts event. Which happens before BeamMobsim.run method where Population is created.
___
*[src/main/scala/beam/agentsim/infrastructure/DefaultRidehailFunctions.scala, line 211 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-Mr5d5i96iTimhdCqD-J:-MrLvN-IBeY2q_NCrfpH:b-2n7af6) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/agentsim/infrastructure/DefaultRidehailFunctions.scala#L211)):*

can we make both 0.8 and 20.0 parametrizable ?  I think we use them both in different places in BEAM
</blockquote></details>

I think this is possible but it's better to create a separate issue for this in order to not make this PR too big and unrelated.
___
*[src/main/scala/beam/sim/BeamHelper.scala, line 350 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-MosuzS505uj3K2OSbWY-r1-350:-Mr6M7AI-wlNMAB_lMzr:b-r40g72) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/dc119924827ad53d5ae7d15c98dcb70f54d5a6c9/src/main/scala/beam/sim/BeamHelper.scala#L350)):*

Here is how things should be:
1) At iteration 0 the SoC of a vehicle should be random when vehicles.csv is not provided, or it doesn't have a column SoC, otherwise we use the SoC from the file.
2) At iteration i > 0, we initialize the SoC of a vehicle from the last SoC of the same vehicle in the previous iteration. 

it makes sense to not change SoC of vehicle in BeamHelp, but instead in HouseholdActor (I think), because we need to initialize the SOC every iteration.
Is that what the code is doing here @dimaopen ?
</blockquote></details>

It's all related to private vehicles. (RH vehicles handled separately)
1 and 2 are true.
Vehicle SoC is set at BeamSim as it was before this PR. The initial SoC in the beam scenario is set at the end of iteration at the same time when the vehicle state is reset.
___
*[src/main/scala/beam/sim/BeamHelper.scala, line 357 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-Mr5du_0-RonuinOKTsn:-Mr6NvS6D_puVUI6kHI1:bpt51mw) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/sim/BeamHelper.scala#L357)):*

This is tricky. I just realized that here is shouldn't be VehicleManager.AnyManager.managerId, but instead: VehicleManager.createOrGetReservedFor(household.getId.toString, VehicleManager.TypeEnum.Household)
as in Population.scala line 122
</blockquote></details>

There's no household at this stage. Let's create a separate issue as it's not related to this PR.
___
*[src/test/scala/beam/utils/scenario/InitialSocLoadingTest.scala, line 29 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-Mr5j5JIAV_Rp7YI8WQJ:-MrLvsDKBZFytiS0liGX:b-ww20ji) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/test/scala/beam/utils/scenario/InitialSocLoadingTest.scala#L29)):*

Also maybe there is a test class already there that came with the MARAIN patch
</blockquote></details>

This converging process is somewhat random and may cause tests to execute too long. I created a test that validates that linking of SoC for electric vehicles across iterations actually happens.
___
*[src/main/scala/beam/agentsim/events/FleetStoredElectricityEvent.scala, line 38 at r5](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370#-MqHUwtTEH4x9eTYLcAT:-Mr6HSdx0wc8tFqAozpW:b-886gsx) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/f323f054d0016b7ce853cae82abe464b18cc48ca/src/main/scala/beam/agentsim/events/FleetStoredElectricityEvent.scala#L38)):*

Did you mean to put a different attribute than ATTRIBUTE_STORAGE_CAPACITY_IN_JOULES ?
</blockquote></details>

Fixed

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

haitamlaarabi (commented):
> Reviewed 2 of 13 files at r1, 3 of 3 files at r2, 2 of 9 files at r3, 2 of 5 files at r5, 7 of 12 files at r6, 11 of 11 files at r8, 2 of 2 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3370)* status: all files reviewed, 6 unresolved discussions (waiting on @alex-vv, @haitamlaarabi, @JustinPihony, and @nikolayilyin)

haitamlaarabi (approved):
Good to go for me

haitamlaarabi:
Test now!

nikolayilyin:
test!

dimaopen:
Test!

PR #3373, opened 2021-11-23
#3341 implement enroute charging protocol

### LINKED ISSUE #3341: Enroute Charging

heyrutvik:
- [x] Find charging stall in the middle of the path
- [x] Avoid enroute even if config values suggests, if charging level is going to be more than 80% of capacity after reaching charging stall (calculate before starting enroute trip)
- [x] Enroute trip protocol 
- [x] Tests

closes #3341

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373)


heyrutvik:
Test!

heyrutvik:
Test!

haitamlaarabi (changes requested):
> Reviewed 20 of 29 files at r1, 28 of 29 files at r2, 1 of 1 files at r3, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373)* status: all files reviewed, 21 unresolved discussions (waiting on @dimaopen and @heyrutvik)

___
*[src/main/resources/beam-template.conf, line 294 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEmWjr0f5ew5VcgK5W:-MpEmWjr0f5ew5VcgK5X:bsc2uyx) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L294)):*
> ```
> beam.agentsim.agents.vehicles.meanRidehailVehicleStartingSOC = "double | 1.0"
> beam.agentsim.agents.vehicles.transitVehicleTypesByRouteFile = ""
> beam.agentsim.agents.vehicles.enroute.refuelRequiredThresholdInPercent = 1.05 # 105%
> ```

I am thinking of changing this from percent to mile
___
*[src/main/resources/beam-template.conf, line 298 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEl3m9AyebQfL4Thep:-MpEl3m9AyebQfL4Theq:b-jshja9) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L298)):*
> ```
> beam.agentsim.agents.vehicles.enroute.maxDurationInSeconds = 1800 # 80% of the battery charges in 30min
> beam.agentsim.agents.vehicles.enroute.thresholdForNotWalkingToDestinationInMeters = 1000 # 1km otherwise drive. This configuration requires fast charging as well
> beam.agentsim.agents.vehicles.enroute.newLegStartDelayInSeconds = 3600 # TODO may be it should be variable, and should depend on enroute trip distance?
> ```

What's the role of this parameter?
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 887 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEo88L8payLbYcTHW-:-MpEo88L8payLbYcTHW0:bhzrqd9) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L887)):*
> ```Scala
>       val stall2DestinationCarLegs = data.enrouteState.stall2DestLegs
>       val walkTemp = data.currentTrip.head.legs.head
>       val startTime = tick + beamServices.beamConfig.beam.agentsim.agents.vehicles.enroute.newLegStartDelayInSeconds
> ```

Why would you make the agent walk after 1 hour, as the default value of newLegStartDelayInSeconds ?
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 901 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEozeh22Ym8CMCZTMW:-MpEozeiCs2iXMH1skO-:b-jv3q3h) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L901)):*
> ```Scala
>       )
>     case _ @Event(UnhandledVehicle(_, _, _), data: BasePersonData) =>
>       // TODO is this valid transition?
> ```

What would happen in this case? Would the agent continue the default workflow ? I think we should consider same steps as in "UnpluggingVehicle". We need to add driving to destination if something goes wrong.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 968 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEqi2y5mTBgXAB2ZI3:-MpEqi2y5mTBgXAB2ZI4:b-4wjfjp) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L968)):*
> ```Scala
>         val needEnroute = {
>           asDriver && isElectric && isRefuelNeeded &&
>           data.enrouteState.notEnroute && data.enrouteState.notAttempted
> ```

what notEnroute and notAttempted means ?
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 1001 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEs7QA5-aESw_yaS5V:-MpEs7QA5-aESw_yaS5W:bzhuap1) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L1001)):*
> ```Scala
>           case ReadyToChooseParking =>
>             (false, copiedData.copy(enrouteState = EnrouteState(enroute = true)))
>           case ReleasingParkingSpot if data.enrouteState.attempted =>
> ```

Does this mean that agent attempted to enroute but failed? or attempted enroute and finished charging?
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 1014 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEsis42Wtm-E-SHROy:-MpEsis42Wtm-E-SHROz:bersadd) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L1014)):*
> ```Scala
> 
>         // complete trigger only if following conditions match
>         if ((sendCompletionNotice && data.enrouteState.notAttempted) || nextLeg.beamLeg.endTime > lastTickOfSimulation)
> ```

the only time sendCompletionNotice == true is when enrouteState is attempted. Am I missing something? So this will never be true
___
*[src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala, line 382 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEth6E26bcIUKl6k2R:-MpEth6E26bcIUKl6k2S:b4ufh3a) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala#L382)):*
> ```Scala
>                     triggerId,
>                     shiftStatus = NotApplicable,
>                     refuelTripType = if (inEnroute) Enroute else Normal
> ```

What do you need a refuelTripType?
If necessary then I would substitute Normal by Destination
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 96 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF5csv4COsQNcBzMBO:-MpF5csv4COsQNcBzMBP:bdf03o4) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L96)):*
> ```Scala
>   this: PersonAgent => // Self type restricts this trait to only mix into a PersonAgent
> 
>   // TODO This might be a violation of Actor framework. Need to verify
> ```

it doesn't. we can delete this comment
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 125 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF60nn-v1X17JRfbez:-MpF60nn-v1X17JRfbf-:b-jith5c) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L125)):*
> ```Scala
>       val reservedFor = VehicleManager.getReservedFor(vehicle.vehicleManagerId.get).get
>       ParkingInquiry.init(
>         calculateMidpointUtm(vehicleTrip).getOrElse(destinationUtm),
> ```

Why are you calculating the midpoint. The MNL parking/charging search already does that. I would mess results if you provide a midpoint.
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 301 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF6oWP7D9z6d_0V5ss:-MpF6oWP7D9z6d_0V5st:b46jpq1) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L301)):*
> ```Scala
>           isDistanceGreaterThanThreshold && isParkingDurationSmallEnoughFor && isStallHasFastCharger
> 
>         val carIfEnroute = if (isEnroute || isDestinationChargeTurningToAnEnrouteCharge) {
> ```

this is a problem here, we cannot turn a destination charging to an enroute charging, because we are already at CAR2? Or did you find a way to make it Enroute?
___
*[src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala, line 135 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF7Y_h4lRk0WUgyxgV:-MpF7Y_i255Ju5B9z2xD:b-vlxef8) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L135)):*
> ```Scala
>   }
> 
>   def unsetReservedParkingStall(): Unit = {
> ```

What is the reason for this?
___
*[src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala, line 664 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF7tDm4emGuheGMHoj:-MpF7tDm4emGuheGMHok:bn3nfhk) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L664)):*
> ```Scala
>   }
> 
>   def fuelConsumptionInJoules(beamVehicle: BeamVehicle, legs: Vector[EmbodiedBeamLeg]): Double = {
> ```

I don't remember adding this function. Did you or did it come from another PR? If you did what does it do?
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 47 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF9Rnl4bed28pIujYd:-MpF9Rnl4bed28pIujYe:b-y1qy18) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L47)):*
> ```Scala
>   }
> 
>   private val refuelTripTypeString = refuelTripType match {
> ```

We won't need this if we merge that PR I am mentioning bellow
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 68 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF8R6n2jD8s11gzV68:-MpF8R6n2jD8s11gzV69:b-9bls7i) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L68)):*
> ```Scala
>     attributes.put(ATTRIBUTE_PERSON, personId.toString)
>     attributes.put(ATTRIBUTE_SHIFT_STATUS, shiftStatusString)
>     attributes.put(ATTRIBUTE_ACTIVITY_TYPE, refuelTripTypeString)
> ```

I have added the Activity Type in a different PR #3340 that has not been merged. We might need to merge that because it is serving other purpose. 
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 96 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFA76HA-rLoMm7nHPM:-MpFA76HA-rLoMm7nHPN:b-dqnzrm) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L96)):*
> ```Scala
>   case object NotApplicable extends ShiftStatus
> 
>   sealed trait RefuelTripType
> ```

We would not need this when we merge #3340
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 100 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFA_HE5sTWoKAL4rU9:-MpFA_HE5sTWoKAL4rUA:bk3jj4w) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L100)):*
> ```Scala
>         case Some(parkingResponse) => sender() ! parkingResponse
>         case _ if inquiry.searchMode == ParkingSearchMode.EnRoute =>
>           (parkingNetworkManager ? inquiry.copy(searchMode = ParkingSearchMode.Destination)).pipeTo(sender())
> ```

Are you taking into consideration this. If the CNM doesn't find a stall with charging point, it will return a stall without charging point? What do you do in this situation?
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 157 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFAsco8IXLCUf5cvNc:-MpFAsco8IXLCUf5cvNd:b-eb6cfd) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L157)):*
> ```Scala
>       sender ! CompletionNotice(triggerId)
> 
>     case request @ ChargingPlugRequest(tick, vehicle, stall, _, triggerId, _, _, refuelTripType) =>
> ```

we won't need this when we will merge #3340
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 407 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFAoa3EdB8dFje8xfJ:-MpFAoa3EdB8dFje8xfK:b-eb6cfd) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L407)):*
> ```Scala
>     shiftDuration: Option[Int] = None,
>     refuelTripType: RefuelTripType = Normal
>   ) extends HasTriggerId
> ```

we won't need this when we will merge #3340
___
*[src/main/scala/beam/router/model/EmbodiedBeamLeg.scala, line 56 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFCsBkCpNCNegQh2ip:-MpFCsBkCpNCNegQh2iq:b-3wybzz) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/router/model/EmbodiedBeamLeg.scala#L56)):*
> ```Scala
>   }
> 
>   def splitLegForParking(
> ```

Why are we splitting the leg?
___
*[src/test/scala/beam/integration/EnrouteChargingSpec.scala, line 128 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFDJl01r8uiF0qPimJ:-MpFDJl01r8uiF0qPimK:b-hd8pd0) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/test/scala/beam/integration/EnrouteChargingSpec.scala#L128)):*
> ```Scala
>               _,
>               _,
>               Enroute
> ```

we will need to update the test once we merge that PR

heyrutvik (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373)* status: all files reviewed, 21 unresolved discussions (waiting on @dimaopen, @haitamlaarabi, and @heyrutvik)

___
*[src/main/resources/beam-template.conf, line 294 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEmWjr0f5ew5VcgK5W:-MpG98MYAvkMm_o7QWSL:bhy9ot6) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L294)):*

I am thinking of changing this from percent to mile
</blockquote></details>

Will do.
___
*[src/main/resources/beam-template.conf, line 298 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEl3m9AyebQfL4Thep:-MpG8X2T4KL-w24vZl0f:bau9bch) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L298)):*

What's the role of this parameter?
</blockquote></details>

The parameter is to specify how much delay we want to add to start a new leg after enroute recharging.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 887 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEo88L8payLbYcTHW-:-MpG9GBJ-nGdQuV72lA0:bjrqyas) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L887)):*

Why would you make the agent walk after 1 hour, as the default value of newLegStartDelayInSeconds ?
</blockquote></details>

Yes, so what was happening was if we schedule the next leg using available tick it was causing error in scheduler saying "tick x when 'nowInSeconds' is at y", where x << y. So I figured that agent should consider some time delay to avoid such issue. Also, the value was randomly chosen, as mentioned in template that it should be decided based on the detour distance or something. Thoughts?
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 901 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEozeh22Ym8CMCZTMW:-MpGAX3q71giW-odXhw7:bhy9ot6) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L901)):*

What would happen in this case? Would the agent continue the default workflow ? I think we should consider same steps as in "UnpluggingVehicle". We need to add driving to destination if something goes wrong.
</blockquote></details>

Will do.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 968 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEqi2y5mTBgXAB2ZI3:-MpGAkDPCLmIRDXP85iP:b-qxisae) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L968)):*
The case class defined to hold enroute related states has following two flags,
- `enroute`, it indicates that agent is on enroute charging detour
- `attempt`, it indicates that agent has attempted enroute for current trip
  - if tried but avoided due to 80% charging level condition
  - enroute charging finished, and now heading toward original destination

> what notEnroute and notAttempted means ?

These two flags are just a negation of them. We enter enrotue detour only if those conditions match.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 1001 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEs7QA5-aESw_yaS5V:-MpGBxbR5RGn5FR0MYMh:bdqb70g) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L1001)):*

Does this mean that agent attempted to enroute but failed? or attempted enroute and finished charging?
</blockquote></details>

I can see now that I can make improvement in the code here to make it less confusing. Will do. 
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 1014 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEsis42Wtm-E-SHROy:-MpGC76SAB3whPbkHKsw:b-7ayq1x) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L1014)):*

the only time sendCompletionNotice == true is when enrouteState is attempted. Am I missing something? So this will never be true
</blockquote></details>

This will be removed once I simplify the code. 
___
*[src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala, line 382 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEth6E26bcIUKl6k2R:-MpGCH5HEXo-qq1NKh1j:b-9ph8yp) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala#L382)):*

What do you need a refuelTripType?
If necessary then I would substitute Normal by Destination
</blockquote></details>

It'll be used to indicate what kind of refuel session it was? normal (parking at the end of the trip) or enroute. We use it in tests too.

Oh, okay. As you mentioned in other comment that we already have solution implemented for this in other branch. I'll remove this change, and merge the mentioned branch. 
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 96 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF5csv4COsQNcBzMBO:-MpGCXUr7IVg-gLjfz3M:bebrt1w) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L96)):*

it doesn't. we can delete this comment
</blockquote></details>

Will do
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 125 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF60nn-v1X17JRfbez:-MpGCyaK8jDmRE_T7Jv-:bnfcc30) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L125)):*
> The MNL parking/charging search already does that. 
I'm not sure what it means. But if you are talking about existing code (`destinationUtm`), it uses location nearby the destination of the trip.

The purpose of calculation is fulfil one of the sub-tasks, to find charging stall in the middle of the path, rather than at the source or destination. I also posted the question about whether we have an existing such method/API, but none was exist. So I wrote one myself.

Did I misunderstand the sub-task? 
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 301 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF6oWP7D9z6d_0V5ss:-MpGDzwZ4cQM1QyKJGPZ:b-50y1iv) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L301)):*
> Or did you find a way to make it Enroute?
Yes. Before coming to this state, we "travel" a leg in normal workflow. We don't do that if agent is on enroute. So we are good here.
___
*[src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala, line 135 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF7Y_h4lRk0WUgyxgV:-MpGEPkM6ArXQ_dhTgf7:bntkyp) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L135)):*

What is the reason for this?
</blockquote></details>

So in the scenario when agent figures out that enroute can be avoided (charging level > 80%), we use this method to unset the parking stall which we stored while receiving charging stall inquiry response.  
___
*[src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala, line 664 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF7tDm4emGuheGMHoj:-MpGEtWI6nYVe7ADvd69:b-omfgtc) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L664)):*

I don't remember adding this function. Did you or did it come from another PR? If you did what does it do?
</blockquote></details>

I added it. We use this to estimate how much fuel will required to reach a destination (in our case, charging stall), to figure out whether enroute can be avoided  (charging level > 80%). Please review [these lines](https://github.com/LBNL-UCB-STI/beam/pull/3373/files#diff-0a5c9a46d00301d67b7ecfc466eb132407e6fdfbf2f417dc90566761eba06096R367-R376).
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 47 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF9Rnl4bed28pIujYd:-MpGFxokBulO7UEZKsXy:bhy9ot6) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L47)):*

We won't need this if we merge that PR I am mentioning bellow
</blockquote></details>

Will do.
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 68 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF8R6n2jD8s11gzV68:-MpGFsln9w6unUwL4YMA:b-yzs2d) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L68)):*

I have added the Activity Type in a different PR #3340 that has not been merged. We might need to merge that because it is serving other purpose. 
</blockquote></details>

Got it. Will do.
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 96 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFA76HA-rLoMm7nHPM:-MpGGipZ1YtNGQcXf6Dm:b-80tlqk) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L96)):*

We would not need this when we merge #3340
</blockquote></details>

Sure.
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 100 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFA_HE5sTWoKAL4rU9:-MpGGkPt5Vr3PgXv4SHd:b6pyxcv) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L100)):*

Are you taking into consideration this. If the CNM doesn't find a stall with charging point, it will return a stall without charging point? What do you do in this situation?
</blockquote></details>

Ah, good catch. I think I don't check whether responded stall is "fast charging" or simple parking stall. What should we be doing in normal parking stall? 
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 157 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFAsco8IXLCUf5cvNc:-MpGHWmi1uoIv8nDB5_P:b-80tlqk) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L157)):*

we won't need this when we will merge #3340
</blockquote></details>

Sure.
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 407 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFAoa3EdB8dFje8xfJ:-MpGHZZMDXX8LuOkntJ1:b-80tlqk) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L407)):*

we won't need this when we will merge #3340
</blockquote></details>

Sure.
___
*[src/main/scala/beam/router/model/EmbodiedBeamLeg.scala, line 56 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFCsBkCpNCNegQh2ip:-MpGH_zu5bfIv_Nsu2Ou:bwzm4ht) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/router/model/EmbodiedBeamLeg.scala#L56)):*

Why are we splitting the leg?
</blockquote></details>

So usually car trip is like, WALK, CAR, CAR, WALK. 
The first CAR is actual trip and second CAR is for parking. The leg we get from routing response is complete single CAR leg which we suppose to break into trip/parking legs. We already had the method to do that, I just made it more usable.
We use it to create legs for vehicle to stall, and stall to destination trip.
___
*[src/test/scala/beam/integration/EnrouteChargingSpec.scala, line 128 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFDJl01r8uiF0qPimJ:-MpGIFzB0VN04BBa7PRG:b-80tlqk) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/test/scala/beam/integration/EnrouteChargingSpec.scala#L128)):*

we will need to update the test once we merge that PR
</blockquote></details>

Sure.

heyrutvik:
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 887 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEo88L8payLbYcTHW-:-MpMBFrN6u2qoRHplRK7:b58dqw6) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L887)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Yes, so what was happening was if we schedule the next leg using available tick it was causing error in scheduler saying "tick x when 'nowInSeconds' is at y", where x << y. So I figured that agent should consider some time delay to avoid such issue. Also, the value was randomly chosen, as mentioned in template that it should be decided based on the detour distance or something. Thoughts?
</blockquote></details>

Okay. It looks like we don't need this param. I removed it in latest commit, I'm not facing the 'tick in past compared to nowInSeconds'.

heyrutvik (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373)* status: 44 of 49 files reviewed, 21 unresolved discussions (waiting on @dimaopen, @haitamlaarabi, and @heyrutvik)

___
*[src/main/resources/beam-template.conf, line 294 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEmWjr0f5ew5VcgK5W:-MpQBdXBEzyGw33EhCLX:b-ih0771) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L294)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Will do.
</blockquote></details>

I assume you meant meters? What should be the default values for both params, noRefuelThreshold and refuelRequiredThreshol? 
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 901 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEozeh22Ym8CMCZTMW:-MpQBsuV8P8qH4WOUHSt:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L901)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Will do.
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 96 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF5csv4COsQNcBzMBO:-MpQC2Di3MhPFFJh7-dO:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L96)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Will do
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 47 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF9Rnl4bed28pIujYd:-MpRU2oiBarE9h85YUxK:b-jbo2mu) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L47)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Will do.
</blockquote></details>

Currently working on merging the branch. It looks like the way we handle `vehicle2InquiryMap` storage is problematic. It only remembers the most recent parking inquiry, and what happens is, it overrides previous 'activityType'. I was testing whether refuel session correctly logs actType or not. If I'm understanding it correctly, then we can use my solution to pass activityType to refuel session  event.

haitamlaarabi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373)* status: 44 of 49 files reviewed, 21 unresolved discussions (waiting on @dimaopen, @haitamlaarabi, and @heyrutvik)

___
*[src/main/resources/beam-template.conf, line 294 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEmWjr0f5ew5VcgK5W:-Mpijdz5AVVViTl2rhBi:bify8p0) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L294)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

I assume you meant meters? What should be the default values for both params, noRefuelThreshold and refuelRequiredThreshol? 
</blockquote></details>

I will do it myself just to be sure that the calculation I made in ChargingFunctions stays correct!
___
*[src/main/resources/beam-template.conf, line 298 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEl3m9AyebQfL4Thep:-Mpijns65F7-qdYpz6vU:b-6ntav9) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L298)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

The parameter is to specify how much delay we want to add to start a new leg after enroute recharging.
</blockquote></details>

Why would we need a delay of 1 hour after a 30min charging period ? The car can leave immediately or after a parallelism window duration after the EnrouteCharging concludes
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 887 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEo88L8payLbYcTHW-:-MpikEor-I8ITjrzsnHX:b-hhv7o3) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L887)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Okay. It looks like we don't need this param. I removed it in latest commit, I'm not facing the 'tick in past compared to nowInSeconds'.  
</blockquote></details>

Thanks. We should not mask errors with hacks. We would rather solve the problem than hide it.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 901 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEozeh22Ym8CMCZTMW:-MpikTB5-SZcO4R1S1ee:bb7m07t) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L901)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

okay
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 968 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEqi2y5mTBgXAB2ZI3:-Mpil3Hw13b3ePALyRIS:b-1zw3e1) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L968)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

The case class defined to hold enroute related states has following two flags,
- `enroute`, it indicates that agent is on enroute charging detour
- `attempt`, it indicates that agent has attempted enroute for current trip
  - if tried but avoided due to 80% charging level condition
  - enroute charging finished, and now heading toward original destination

> what notEnroute and notAttempted means ?

These two flags are just a negation of them. We enter enrotue detour only if those conditions match.
</blockquote></details>

Why does it attempt charging if charging level condition is 80%. I think
We should allow multiple enroute happening one after another if the battery is not enough to reach destination. Which means EnrouteCharge then when leaving recalculate remained energy, if not enough to destination we should ask for enroute between current stall and original destination. 
In other words, I do not see reason for the flag "attempt", but "enroute" flag makes sense. 
___
*[src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala, line 382 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEth6E26bcIUKl6k2R:-Mpimcei4D_3Jd31JAfY:b-9js7jr) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala#L382)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

It'll be used to indicate what kind of refuel session it was? normal (parking at the end of the trip) or enroute. We use it in tests too.

Oh, okay. As you mentioned in other comment that we already have solution implemented for this in other branch. I'll remove this change, and merge the mentioned branch. 
</blockquote></details>

I think Justin approved the PR. I will look into it and merge
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 125 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF60nn-v1X17JRfbez:-MpimsYw79enRxHj5lh2:b-49uwnm) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L125)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

> The MNL parking/charging search already does that. 
I'm not sure what it means. But if you are talking about existing code (`destinationUtm`), it uses location nearby the destination of the trip.

The purpose of calculation is fulfil one of the sub-tasks, to find charging stall in the middle of the path, rather than at the source or destination. I also posted the question about whether we have an existing such method/API, but none was exist. So I wrote one myself.

Did I misunderstand the sub-task? 
</blockquote></details>

I actually did that subtask before I shared with you the branch. In ChargingFunction and Parking function, we are using a Ellipse to look for a station between origin and destination. The whole work I done in EnrouteCharging was mainly to calculate the midpoint. We will need to chat on zoom about it.
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 301 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF6oWP7D9z6d_0V5ss:-MpinOcU58N-Ga2kJ09C:bem1ozd) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L301)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

> Or did you find a way to make it Enroute?
Yes. Before coming to this state, we "travel" a leg in normal workflow. We don't do that if agent is on enroute. So we are good here.
</blockquote></details>

I still don't understand. Let's talk about it on Zoom
___
*[src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala, line 135 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF7Y_h4lRk0WUgyxgV:-MpinX-_DuvaZbqBH4CQ:bhc7q6y) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L135)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

So in the scenario when agent figures out that enroute can be avoided (charging level > 80%), we use this method to unset the parking stall which we stored while receiving charging stall inquiry response.  
</blockquote></details>

I feel that something is wrong here, but I might be mistaken. Let's talk about on Zoom
___
*[src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala, line 664 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF7tDm4emGuheGMHoj:-MpinhbK-3H5Sla59wj2:b-mc6kq7) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala#L664)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

I added it. We use this to estimate how much fuel will required to reach a destination (in our case, charging stall), to figure out whether enroute can be avoided  (charging level > 80%). Please review [these lines](https://github.com/LBNL-UCB-STI/beam/pull/3373/files#diff-0a5c9a46d00301d67b7ecfc466eb132407e6fdfbf2f417dc90566761eba06096R367-R376).
</blockquote></details>

Okay then, I think we ended up duplicating work. I have already did all the calculation in that branch. let's talk about it on Zoom
___
*[src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala, line 47 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpF9Rnl4bed28pIujYd:-MpinzSx9JjAuq_4bqh6:b-9aarac) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/events/RefuelSessionEvent.scala#L47)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Currently working on merging the branch. It looks like the way we handle `vehicle2InquiryMap` storage is problematic. It only remembers the most recent parking inquiry, and what happens is, it overrides previous 'activityType'. I was testing whether refuel session correctly logs actType or not. If I'm understanding it correctly, then we can use my solution to pass activityType to refuel session  event. 
</blockquote></details>

Let's talk about it on Zoom (I'm marking all comments with Zoom so we can track them better)
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 100 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFA_HE5sTWoKAL4rU9:-MpioABX5G95FZQbrpjU:bqkhm25) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L100)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

Ah, good catch. I think I don't check whether responded stall is "fast charging" or simple parking stall. What should we be doing in normal parking stall? 
</blockquote></details>

If it doesn't return a stall then we have to continue with normal workflow.
___
*[src/main/scala/beam/router/model/EmbodiedBeamLeg.scala, line 56 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFCsBkCpNCNegQh2ip:-MpioN3m9qcBpcd6PixQ:bb71kmx) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/router/model/EmbodiedBeamLeg.scala#L56)):*
<details><summary><i>Previously, heyrutvik (Rutvik Patel) wrote…</i></summary><blockquote>

So usually car trip is like, WALK, CAR, CAR, WALK. 
The first CAR is actual trip and second CAR is for parking. The leg we get from routing response is complete single CAR leg which we suppose to break into trip/parking legs. We already had the method to do that, I just made it more usable.
We use it to create legs for vehicle to stall, and stall to destination trip.
</blockquote></details>

Okay

heyrutvik (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373)* status: 44 of 49 files reviewed, 21 unresolved discussions (waiting on @dimaopen, @haitamlaarabi, and @heyrutvik)

___
*[src/main/resources/beam-template.conf, line 294 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEmWjr0f5ew5VcgK5W:-MppaFb-DFMsIECanrND:b-bty3uv) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L294)):*

I will do it myself just to be sure that the calculation I made in ChargingFunctions stays correct!
</blockquote></details>

Sure. Thanks.
___
*[src/main/resources/beam-template.conf, line 298 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEl3m9AyebQfL4Thep:-MppaHevEx-pJMyfSZ3h:b-896fix) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/resources/beam-template.conf#L298)):*

Why would we need a delay of 1 hour after a 30min charging period ? The car can leave immediately or after a parallelism window duration after the EnrouteCharging concludes
</blockquote></details>

Done.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 887 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEo88L8payLbYcTHW-:-MppaMEB0_tHNpEn2t-u:b-lni3zf) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L887)):*

Thanks. We should not mask errors with hacks. We would rather solve the problem than hide it.
</blockquote></details>

I agree. How tick works was a bit black box for me. I've made an appropriate change, but it seems like once in a while agent dies due to the same error. Let's talk about it over ZOOM call.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 968 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpEqi2y5mTBgXAB2ZI3:-MppamIOBu0P3Hl-iNJU:b-e8kumc) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L968)):*

Why does it attempt charging if charging level condition is 80%. I think
We should allow multiple enroute happening one after another if the battery is not enough to reach destination. Which means EnrouteCharge then when leaving recalculate remained energy, if not enough to destination we should ask for enroute between current stall and original destination. 
In other words, I do not see reason for the flag "attempt", but "enroute" flag makes sense. 
</blockquote></details>

Got it. We had such a behaviour in my initial draft where a vehicle keep looking for charging stall if its required. I thought it's flawed design. I also use 'attempt' to avoid re-searching charging stall if 80% charge is available. Let's talk about it over ZOOM call.
___
*[src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala, line 100 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpFA_HE5sTWoKAL4rU9:-MppbkLX99nLrYoVZ40j:b-ye2v1) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/eaaecb274bf6c470bfabd32ef5aca1bc697bcddd/src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala#L100)):*

If it doesn't return a stall then we have to continue with normal workflow.
</blockquote></details>

Handled. Done.

src/main/scala/beam/router/skim/readonly/DriveTimeSkims.scala
```diff
@@ -2,4 +2,4 @@ package beam.router.skim.readonly
 
 import beam.router.skim.core.AbstractSkimmerReadOnly
 
-case class DriveTimeSkims() extends AbstractSkimmerReadOnly {}
+class DriveTimeSkims() extends AbstractSkimmerReadOnly {}
```
JustinPihony, line 5: Why remove the `case`s?

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -797,6 +840,87 @@ class PersonAgent(
...
+        id,
+        parkingManager,
+        eventsManager,
+        triggerId
+      )
+      val (updatedTick, updatedData) = createStallToDestTripForEnroute(data, tick)
+      holdTickAndTriggerId(updatedTick, triggerId)
+      goto(ProcessingNextLegOrStartActivity) using updatedData
+    case Event(UnhandledVehicle(tick, vehicleId, triggerId), data: BasePersonData) =>
+      log.debug(
```
JustinPihony, line 881: Should this be a warning - or even an error?

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -842,57 +967,82 @@ class PersonAgent(
...
+        val vehicleTrip = data.restOfCurrentTrip.takeWhile(_.beamVehicleId == firstLeg.beamVehicleId)
+        val totalDistance = vehicleTrip.map(_.beamLeg.travelPath.distanceInM).sum
+        val refuelRequiredThresholdInMeters = totalDistance * enrouteConfig.refuelRequiredThresholdInPercent
+        val noRefuelThresholdInMeters = totalDistance * enrouteConfig.noRefuelThresholdInPercent
+        val asDriver = data.restOfCurrentTrip.head.asDriver
+        val isElectric = vehicle.isEV
+        val isRefuelNeeded = vehicle.isRefuelNeeded(refuelRequiredThresholdInMeters, noRefuelThresholdInMeters)
+        val needEnroute = asDriver && isElectric && isRefuelNeeded
+
+        val isSharedVehicle = beamVehicles(nextLeg.beamVehicleId)
```
JustinPihony, line 984: Can you use `vehicle` that was already created earlier?

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -842,57 +967,82 @@ class PersonAgent(
...
-              "ProcessingNextLegOrStartActivity, going to ReleasingParkingSpot with legsToInclude: {}",
-              legsToInclude
-            )
-            ReleasingParkingSpot
-          } else {
-            releaseTickAndTriggerId()
-            WaitingToDrive
-          }
-        goto(stateToGo) using data.copy(
+        val copiedData = data.copy(
```
JustinPihony, line 996: Is this even used? As far as I see it is only used for further copying, but never used.

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
```diff
@@ -49,6 +59,21 @@ class ChargingFunctions[GEO: GeoLevel](
     }
   }
 
+  /**
+    * function that verifies if Enroute Then Fast Charging Only
+    * @param zone ParkingZone
+    * @param inquiry ParkingInquiry
+    * @return
+    */
+  def ifEnrouteChargingTheFastChargingOnly(zone: ParkingZone[GEO], inquiry: ParkingInquiry): Boolean = {
```
JustinPihony, line 68: Typo - ...ThenFast...

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -189,20 +208,24 @@ trait ChoosesParking extends {
...
-        )
-
+        val (tick, triggerId) = releaseTickAndTriggerId()
+        data match {
+          case data: BasePersonData if data.enrouteState.attempted && currentBeamVehicle.isEV =>
+            scheduler ! ScheduleTrigger(StartLegTrigger(tick, nextLeg), self)
+          case _ =>
+            scheduler ! CompletionNotice(
+              triggerId,
+              Vector(ScheduleTrigger(StartLegTrigger(tick, nextLeg), self))
```
JustinPihony, line 226: Should this be `nextLeg.startTime` from the original code. I believe only the enroute is new - otherwise it should be the original logic

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -244,43 +267,140 @@ trait ChoosesParking extends {
         )
         val futureVehicle2StallResponse = router ? veh2StallRequest
 
+        val isEnroute = data match { case data: BasePersonData => data.enrouteState.isEnroute; case _ => false }
+        val hasChargingPointAndIsFastCharger = chargingPointMaybe.exists(ChargingPointType.isFastCharger)
+
+        val carIfEnroute = if (isEnroute && hasChargingPointAndIsFastCharger) {
+          currentBeamVehicle.setReservedParkingStall(Some(stall))
+          // get car route from stall to destination, TODO note we give a dummy start time and update later based on drive time to stall
```
JustinPihony, line 275: Should the start time be something absurd to track it back in case something goes off track?

JustinPihony (changes requested):
Most are FYI - there is one deviation to be addressed at least.

haitamlaarabi (changes requested):
> Reviewed 8 of 18 files at r6, 1 of 24 files at r7, 18 of 23 files at r8, 6 of 6 files at r9, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373)* status: all files reviewed, 29 unresolved discussions (waiting on @dimaopen, @haitamlaarabi, @heyrutvik, and @JustinPihony)

___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 881 at r9](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpCTktSch4E3hGJB-WT-r9-881:-MsOvGG36NlHHLfKw0u_:bq808tk) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/9b0e60e319ed67b402f4a7e17e41e502213fef11/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L881)):*

Should this be a warning - or even an error?
</blockquote></details>

I agree. Let's change it from debug to warn
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 977 at r9](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MsOvox7DZj4s5-Mpzbw:-MsOvox7DZj4s5-Mpzbx:b-99hkba) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/9b0e60e319ed67b402f4a7e17e41e502213fef11/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L977)):*
> ```Scala
>         val vehicleTrip = data.restOfCurrentTrip.takeWhile(_.beamVehicleId == firstLeg.beamVehicleId)
>         val totalDistance = vehicleTrip.map(_.beamLeg.travelPath.distanceInM).sum
>         val refuelRequiredThresholdInMeters = totalDistance * enrouteConfig.refuelRequiredThresholdInPercent
> ```

I am not seeing the change I introduced to how we compute this value.
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 984 at r9](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpCTktSch4E3hGJB-WT-r9-984:-MsOw43u4hwV0hVccWrW:b7anmnw) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/9b0e60e319ed67b402f4a7e17e41e502213fef11/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L984)):*

Can you use `vehicle` that was already created earlier?
</blockquote></details>

good call
___
*[src/main/scala/beam/agentsim/agents/PersonAgent.scala, line 996 at r9](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpCTktSch4E3hGJB-WT-r9-996:-MsOwKj51T9IRHQPCU-Y:b2k5nfg) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/9b0e60e319ed67b402f4a7e17e41e502213fef11/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L996)):*

Is this even used? As far as I see it is only used for further copying, but never used.
</blockquote></details>

this section of the code is a bit confusing I have tried to change it once, but since there was a bug to solve first I retracted my changes
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 226 at r9](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpCTktSch4E3hGJB-WW-r9-226:-MsOxOrN4odGr9cVJjUY:bbspxdr) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/9b0e60e319ed67b402f4a7e17e41e502213fef11/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L226)):*

Should this be `nextLeg.startTime` from the original code. I believe only the enroute is new - otherwise it should be the original logic
</blockquote></details>

I agree. @heyrutvik was nextLeg.startTime causing the old tick ? if yes  maybe we should take max of (next.Leg.startTime) and (tick). Otherwise, let's keep normal workflow with nextLeg.startTime
___
*[src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala, line 275 at r9](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3373#-MpCTktSch4E3hGJB-WW-r9-275:-MsOy-Y4B_VWI1Qy_8X7:b-iztc19) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/9b0e60e319ed67b402f4a7e17e41e502213fef11/src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala#L275)):*

Should the start time be something absurd to track it back in case something goes off track?
</blockquote></details>

I might have introduced that comment at some point an forgot why I wrote that. The code changed a lot since then

heyrutvik:
We were facing some issues in the enroute protocol such as, stuckness and agents were dieing. @dimaopen worked on top of this branch to resolve those issues. Dima's branch is `do/#3341-enroute-charing-fix-1`. I'm closing this PR as it's out dated. cc @JustinPihony @haitamlaarabi

PR #3381, opened 2021-11-25
autoconfigure max ram

### LINKED ISSUE #3341: Enroute Charging

rajnikantsh:
Closes #3341 

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3381)

### DIFF
--- a/aws/src/main/python/beam_lambda/lambda_function.py
+++ b/aws/src/main/python/beam_lambda/lambda_function.py
@@ -7,6 +7,27 @@ import glob
 import base64
 from botocore.errorfactory import ClientError
 
+HELICS_RUN = '''sudo /home/ubuntu/install-and-run-helics-scripts.sh
+  -    cd /home/ubuntu/git/beam
+  -    '''
+
+HELICS_OUTPUT_MOVE_TO_BEAM_OUTPUT = '''
+  -    opth="output"
+  -    echo $opth
+  -    finalPath=""
+  -    for file in $opth/*; do
+  -       for path2 in $file/*; do
+  -         finalPath="$path2";
+  -       done;
+  -    done;
+  -    finalPath="${finalPath}/helics_output" 
+  -    mkdir "$finalPath"
+  -    sudo mv /home/ubuntu/git/beam/src/main/python/gemini/*.log "$finalPath"
+  -    sudo mv /home/ubuntu/git/beam/src/main/python/gemini/recording_output.txt "$finalPath"
+  -    cd "$finalPath"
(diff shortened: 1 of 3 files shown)


aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -607,6 +671,10 @@ def stop_instance(instance_ids):
 def terminate_instance(instance_ids):
     return ec2.terminate_instances(InstanceIds=instance_ids)
 
+def calculate_max_ram(instance_type):
```
JustinPihony: This isn't horrible, but a possible simplified solution would be:

```
  max_system_ram = 15
  percent_towards_system_ram = .25
  return ram - min(ram * percent_towards_system_ram, max_system_ram)
```

Then you wouldn't need a map of percentages, and you get about the same thing in one shot.

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -265,6 +265,33 @@
                   'z1d.large', 'z1d.xlarge', 'z1d.2xlarge', 'z1d.3xlarge', 'z1d.6xlarge', 'z1d.12xlarge',
                   'x2gd.metal', 'x2gd.16xlarge']
 
+instance_types_with_memory = {
```
JustinPihony: Could we rename to either `instance_type_to_memory` or `memory_in` (that way you could write `memory_in[instance_type]`)

JustinPihony (approved):
Nothing worth blocking

JustinPihony:
test!

aws/src/main/python/beam_lambda/lambda_function.py
rajnikantsh: applied this suggestion

aws/src/main/python/beam_lambda/lambda_function.py
rajnikantsh: @JustinPihony is it fine to run large instances on 15 GB like 2000GB or 4000GB . because we have instances having this much of memory.

JustinPihony: That is more than the current calculation - 15 gb versus (4000*.0025=10). So, I figured it was about what was expected overall. As to if that is enough system memory left - that is something we can tweak as needed.

rajnikantsh: thanks @JustinPihony , Changed max ram calculation as per suggestion

JustinPihony:
Test!

PR #3382, opened 2021-11-25
Xuan/#3330 additional parameters for links -- R5MnetBuilder

### LINKED ISSUE #3330: Additional parameters for links
For the curb management project we're going to need to tweak the link-level volume delay functions used in BPRSim and JDEQSim. To do this, we want the ability to store additional optional parameters for each link. This could be the alpha and beta parameters for the BPR function:

image

Or coefficients for a polynomial approximating the volume/delay relationship:

image

I think it should be pretty easy to set these fields to empty by default but then update them for specific links through the `updatedLinkParams.csv` input, but let's discuss this week.

Xuan-1998:
Build up R5MnetBuilder for additional parameters for links

Fixes #3330 

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3382)


Xuan-1998:
Test!

dimaopen (commented):
I think that it is better to change OsmToMATSim class in the beam-utilities project. Have you had a chance to discuss this way with @wrashid ?

Xuan-1998:
> I think that it is better to change OsmToMATSim class in the beam-utilities project. Have you had a chance to discuss this way with @wrashid ?

I didn’t, @wrashid could you take a look in this way? If it’s not good, I can definitely switch to change OsmToMATSim class in beam-utilities project

JustinPihony (dismissed):
Rather than overwriting, we should update the source directly and push a new version to use - https://github.com/LBNL-UCB-STI/beam-utilities/blob/master/src/main/java/beam/router/r5/OsmToMATSim.java

Xuan-1998:
> 

@zneedell what do you think? If you think this solution is ok. I can work on this after you grant me the access right to beam-utilities

Xuan-1998:
Test!

Xuan-1998:
> Rather than overwriting, we should update the source directly and push a new version to use - https://github.com/LBNL-UCB-STI/beam-utilities/blob/master/src/main/java/beam/router/r5/OsmToMATSim.java

I update the source, I don't need to change anythig in  build.gradle since the it is already using:
 ```
implementation(group: 'com.github.LBNL-UCB-STI', name: 'beam-utilities', version: 'v0.2.12') {
        exclude group: 'com.github.LBNL-UCB-STI', module: 'r5'
        exclude group: 'org.matsim', module: 'matsim'
    }
```
Just pushed the latest change

Xuan-1998:
> > Rather than overwriting, we should update the source directly and push a new version to use - https://github.com/LBNL-UCB-STI/beam-utilities/blob/master/src/main/java/beam/router/r5/OsmToMATSim.java
> 
> I update the source, I don't need to change anythig in build.gradle since the it is already using:
> 
> ```
> implementation(group: 'com.github.LBNL-UCB-STI', name: 'beam-utilities', version: 'v0.2.12') {
>        exclude group: 'com.github.LBNL-UCB-STI', module: 'r5'
>        exclude group: 'org.matsim', module: 'matsim'
>    }
> ```
> 
> Just pushed the latest change

Actually I also need to publish new release

Xuan-1998:
Test!

Xuan-1998:
Test!

Xuan-1998:
Test!

Xuan-1998:
Test!

Xuan-1998:
@JustinPihony @dimaopen please review this change when you get a chance :)

dimaopen (approved):

PR #3402, opened 2021-12-06
Freight: changes to work with new nrel format

nikolayilyin:
New reader for input freight plans + functionality to generate freight activities with fixed duration

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3402)


nikolayilyin:
test!

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -663,6 +693,20 @@ trait BeamHelper extends LazyLogging {
...
+        beamServices.geo,
+        beamScenario,
+        scenario.getPopulation,
+        scenario.getHouseholds
+      )
+      logger.info(s"""Freight population generated:
+                     |Number of households: ${scenario.getHouseholds.getHouseholds.keySet.size}
+                     |Number of persons: ${scenario.getPopulation.getPersons.keySet.size}""".stripMargin)
+    }
+
```
dimaopen, line 729: If we move population generation to runBeam, then `prepareBeamService` will not return these additional persons when is called in other places (i.e. PhyssimReplayer). And we would have inconsistency in this case.

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -835,13 +879,7 @@ trait BeamHelper extends LazyLogging {
           throw new NotImplementedError(s"ScenarioSource '$src' is not yet implemented")
         }
       }
-    generatePopulationForPayloadPlans(
-      beamConfig,
-      geoUtils,
```
dimaopen, line 867: we don't need these beamConfig and geoUtils anymore in this method params

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -284,27 +284,56 @@ trait BeamHelper extends LazyLogging {
...
+      )
+      val carriers = PayloadPlansConverter.readFreightCarriers(
+        freightConfig,
+        geoUtils,
+        networkCoordinator.transportNetwork.streetLayer,
         tours,
         plans,
         vehicleTypes,
-        tazMap,
         rand
```
dimaopen: I think this code is big enough to extract it to some method with a meaningful method name.

src/test/scala/beam/agentsim/agents/freight/FreightReplannerSpec.scala
```diff
@@ -18,121 +18,123 @@ import scala.util.Random
...
+//      val activity0 = plan.getPlanElements.get(0).asInstanceOf[Activity]
+//      activity0.getType should be("Warehouse")
+//      activity0.getEndTime should be(11915.0)
+//      plan.getPlanElements.get(2).asInstanceOf[Activity].getType should be("Unloading")
+//      plan.getPlanElements.get(4).asInstanceOf[Activity].getType should be("Unloading")
+//      plan.getPlanElements.get(6).asInstanceOf[Activity].getType should be("Loading")
+//      plan.getPlanElements.get(8).asInstanceOf[Activity].getType should be("Warehouse")
+//    }
+//
+//  }
```
dimaopen: Why was this test disabled?

dimaopen (changes requested):
I think that tests need to be restored, and also adding freight population should happen during scenario loading because this freight stuff is a part of scenario.

nikolayilyin:
test!

src/main/scala/beam/agentsim/agents/Population.scala
```diff
@@ -100,7 +100,7 @@ class Population(
       }
       if (
         scenario.getHouseholds.getHouseholdAttributes
-          .getAttribute(household.getId.toString.toLowerCase(), "homecoordy") == null
+          .getAttribute(household.getId.toString, "homecoordy") == null
```
JustinPihony, line 103: This seems like it was put there for a reason - do we want to be case sensitive now?

src/test/scala/beam/agentsim/agents/freight/input/PayloadPlansConverterSpec.scala
```diff
@@ -22,129 +26,145 @@ import scala.util.Random
...
+    toursFilePath = s"$freightInputDir/freight-tours.csv",
+    convertWgs2Utm = false,
+    enabled = true,
+    name = "Freight",
+    replanning = new Freight.Replanning(departureTime = 0, disableAfterIteration = -1, strategy = ""),
+    generateFixedActivitiesDurations = false
+  )
+
+// issue for fixing old tests with old reader - https://github.com/LBNL-UCB-STI/beam/issues/3385
+//  "PayloadPlansConverter" should {
```
JustinPihony: If there is an issue then we should either make this @ignore so it is explicit or remove it for now and let git history be the guide for the future work

src/main/scala/beam/agentsim/agents/freight/input/PayloadPlansConverter.scala
```diff
@@ -256,7 +388,19 @@ object PayloadPlansConverter {
     currentPlan
   }
 
-  def createPersonId(vehicleId: Id[BeamVehicle]): Id[Person] = Id.createPersonId(s"freight-agent-$vehicleId")
+  def createPersonId(vehicleId: Id[BeamVehicle]): Id[Person] = {
+    if (vehicleId.toString.startsWith(freightIdPrefix)) {
+      Id.createPersonId(s"$vehicleId-agent")
+    } else {
+      Id.createPersonId(s"freight-$vehicleId-agent")
```
JustinPihony: This and the other create should use the `freightIdPrefix`

src/main/scala/beam/agentsim/agents/freight/input/PayloadPlansConverter.scala
```diff
@@ -21,76 +23,178 @@ import scala.util.Random
...
+    geoUtils: GeoUtils,
+    streetLayer: StreetLayer
+  ): Map[Id[FreightTour], FreightTour] = {
+
+    val maybeTours = GenericCsvReader
+      .readAsSeq[Option[FreightTour]](freightConfig.toursFilePath) { row =>
+        def get(key: String): String = getRowValue(freightConfig.toursFilePath, row, key)
+        // tourId,departureTimeInSec,departureLocation_zone,maxTourDurationInSec,departureLocationX,departureLocationY
+        val tourId: Id[FreightTour] = get("tour_id").createId[FreightTour]
+        val departureTimeInSec = get("departureTimeInSec").toDouble.toInt
```
JustinPihony: Why not go directly to an `Int`?

alex-vv: I think because using directly `toInt` cannot parse decimal values. I now added `round` to avoid truncating

src/main/scala/beam/agentsim/agents/freight/input/PayloadPlansConverter.scala
alex-vv: done

src/main/scala/beam/sim/BeamHelper.scala
alex-vv: I think it makes sense, @nikolayilyin why it was placed there?

alex-vv:
test!

src/main/scala/beam/agentsim/agents/Population.scala
nikolayilyin: The thing is - for "homecoordx" it is case sensitive. Also this change was required to work with not-lower-case ID's. 
We can revert the change, but then we will need to fix the creation of wrong-case-ID's and do the same change for "homecoordx".
Personally I think we might leave it as is for now.

src/main/scala/beam/agentsim/agents/freight/input/PayloadPlansConverter.scala
nikolayilyin: 'departureTimeInSec' column contains doubles with fractional part equal to 0 (i.e. 1430.0)

nikolayilyin:
test!

src/main/scala/beam/sim/BeamHelper.scala
alex-vv: @dimaopen so this change was made because population generation must be done after population sampling which is located here in `runBeam`. As I checked the other usages of `buildBeamServicesAndScenario` and `prepareBeamService` population from scenario not is not used there, so it should not be the issue.

alex-vv:
test!

dimaopen (approved):

PR #3403, opened 2021-12-06
Remove default max from auto configuration

rajnikantsh:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3403)

### DIFF
--- a/aws/build.gradle
+++ b/aws/build.gradle
@@ -93,7 +93,6 @@ task deploy(type: AWSLambdaInvokeTask) {
   "experiments": "${ext.getParameterValue('beamExperiments') ?: ext.getParameterValue(getCurrentGitBranch() + '.experiments')}",
   "execute_class": "${ext.getParameterValue('executeClass')}",
   "execute_args": "${ext.getParameterValue('executeArgs')}",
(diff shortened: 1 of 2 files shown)


aws/build.gradle
```diff
@@ -93,7 +93,6 @@ task deploy(type: AWSLambdaInvokeTask) {
   "experiments": "${ext.getParameterValue('beamExperiments') ?: ext.getParameterValue(getCurrentGitBranch() + '.experiments')}",
   "execute_class": "${ext.getParameterValue('executeClass')}",
   "execute_args": "${ext.getParameterValue('executeArgs')}",
-  "forced_max_ram": "${ext.getParameterValue('forcedMaxRAM')}",
```
dgm90, line 96: why remove this parameter?

rajnikantsh: this parameter was taking default as 'null' . we don't need to use its value as null. so completely removed it. If someone need this param then they need to provide some valid value

dgm90 (approved):

nikolayilyin (approved):
lgtm

dgm90:
Test!

PR #3410, opened 2021-12-09
Possible fix for flaky tests

nikolayilyin:
This PR should fix the problem with parallel run of tests - multiple rewrites of system property to store config location. 
In this PR the usage of system property to store config location was removed.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3410)

### DIFF
--- a/src/main/scala/beam/sim/BeamConfigChangesObservable.scala
+++ b/src/main/scala/beam/sim/BeamConfigChangesObservable.scala
@@ -1,17 +1,18 @@
 package beam.sim
 
-import beam.analysis.physsim.PhyssimCalcLinkStats
 import beam.sim.config.BeamConfig
 import beam.utils.BeamConfigUtils
 import com.typesafe.scalalogging.LazyLogging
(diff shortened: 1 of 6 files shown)


src/main/scala/beam/sim/BeamConfigChangesObservable.scala
```diff
@@ -72,14 +73,12 @@ class BeamConfigChangesObservable @Inject() (beamConfig: BeamConfig) {
     }
   }
 
-  val observable = new WeaklyObservable()
+  private val observable = new WeaklyObservable()
 
   var lastBeamConfig: BeamConfig = beamConfig
```
dimaopen, line 78: This var should be either `volatile` or `AtomicReference`. But this trick decreases performance since during simulation agents refer to this lastBeamConfig. I think that if it worked before with a static field then we can leave it as it is for now.

dimaopen (approved):
I think that we can leave it without that volatile annotation.

PR #3424, opened 2021-12-19
NYC: terminate simulation at the beginning of iteration if config flag set

### LINKED ISSUE #3374: Stop simulation after secondary activities generation.
We need a way to obtain plans with generated secondary activities.
One of ways to do so - stop a simulation after the first iteration started.
One needs to 
1. add a boolean config entry, beam.output.writePlansAndStopSimulation (by default false)
2. dump all plans to output files.
3. stop simulation if the flag is true.

nikolayilyin:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3424)

### DIFF
--- a/src/test/scala/beam/sim/WritePlansAndStopSimulationTest.scala
+++ b/src/test/scala/beam/sim/WritePlansAndStopSimulationTest.scala
@@ -28,7 +28,7 @@ class WritePlansAndStopSimulationTest extends AnyFlatSpec with Matchers with Bea
       beamScenario: BeamScenario,
       services: BeamServices,
       plansMerged: Boolean
-      ) = prepareBeamService(config, None)
+    ) = prepareBeamService(config, None)
 
     val exception: RuntimeException = intercept[RuntimeException] {
       runBeam(
--- a/src/test/scala/beam/sim/WritePlansAndStopSimulationTest.scala
+++ b/src/test/scala/beam/sim/WritePlansAndStopSimulationTest.scala
@@ -1,0 +1,50 @@
+package beam.sim
+
+import beam.sim.config.BeamExecutionConfig
+import beam.utils.TestConfigUtils.testConfig
+import com.typesafe.config.ConfigFactory
+import org.matsim.core.scenario.MutableScenario
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+import java.nio.file.Paths
+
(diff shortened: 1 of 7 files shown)


nikolayilyin:
test!

nikolayilyin:
test!

dimaopen (commented):
Is it possible instead of throwing an exception create a new utility app that loads the scenario, generates secondary activities, writes plans, and exits?
Or this would take a lot more effort?

src/test/scala/beam/sim/WritePlansAndStopSimulationTest.scala
```diff
@@ -0,0 +1,50 @@
...
+      runBeam(
+        services,
+        scenario,
+        beamScenario,
+        beamExecutionConfig.outputDirectory,
+        plansMerged
+      )
+    }
+
+    exception.getMessage shouldBe "The simulation was stopped because beam.output.writePlansAndStopSimulation set to true."
```
JustinPihony, line 43: nit = I'm never a fan of a string comparison like this (brittle).

src/test/scala/beam/sim/WritePlansAndStopSimulationTest.scala
```diff
@@ -0,0 +1,50 @@
...
+
+    val outputDir = Paths.get(beamExecutionConfig.outputDirectory).toFile
+    val outputFiles = outputDir.listFiles()
+    outputFiles.size shouldNot equal(0)
+    outputFiles.map(_.getName) should contain("generatedPlans.csv.gz")
```
JustinPihony, line 48: Should we cleanup files? Granted I think this is an overall problem - maybe we could create an overall cleanup in the after all tests....but that would be a new issue.....at the same time it allows later debugging and CI is cleared anyway....so maybe not worth it...just thoughts I guess...

JustinPihony (approved):
Some minor notes - otherwise lgtm

src/test/scala/beam/sim/WritePlansAndStopSimulationTest.scala
nikolayilyin: Yes, generally I agree, but for this case we need to either check the error message or to throw a special type of error just for this case, which, in my opinion, is overkill.

nikolayilyin:
> Is it possible instead of throwing an exception create a new utility app that loads the scenario, generates secondary activities, writes plans, and exits?
> Or this would take a lot more effort?

I was looking for this approach, and yes, there would be way more work.

PR #3425, opened 2021-12-21
FIx None Value of trip_id

### LINKED ISSUE #3368: Include activitysim trip_id as attribute in BEAM plans
Right now the BEAM urbansim/activitysim csv inputs have a trip_id column. Can we find a way to read this in and store it as a trip attribute that can be output in our plans and some relevant events (mode choice, PersonEntersVehicle, etc.)?

image

Xuan-1998:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3425)

### DIFF
--- a/src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
+++ b/src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
@@ -566,7 +566,6 @@ class UrbanSimScenarioLoader(
             case Some(mode) =>
               val leg = PopulationUtils.createLeg(mode)
 
-
               if (planInfo.tripId != null) {
                 leg.getAttributes.putAttribute("trip_id", planInfo.tripId)
               } else {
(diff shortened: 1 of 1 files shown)


Xuan-1998:
Test!

Xuan-1998:
Test!

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -565,13 +565,27 @@ class UrbanSimScenarioLoader(
           planInfo.legMode match {
             case Some(mode) =>
               val leg = PopulationUtils.createLeg(mode)
-              leg.getAttributes.putAttribute("trip_id", planInfo.tripId)
+
+              if (planInfo.tripId != null) {
```
JustinPihony: You can embed this:

```
leg.getAttributes.putAttribute("trip_id", if(planInfo.tripId != null) planInfo.tripId else "")
```

or (more terse, but more objects for GC)

```
leg.getAttributes.putAttribute("trip_id", Option(planInfo.tripId).getOrElse(""))
```

JustinPihony: Or since it is all off of `tripId`, then you could move it outside of the match to be used in a single place:

```
val tripId = //whichever path to set - if/else or Option
..
//use stored tripId directly
```

JustinPihony (approved):
Stylistic suggestions more than anything, bit more scala-esque than java-esque.

Xuan-1998:
Test!

PR #3428, opened 2021-12-22
add error types in csv

### LINKED ISSUE #3390: Health Metrics: group errors by type
As additional way to understand what went wrong in simulation we need to group errors somehow.
Grouped errors with counts should be added as additional lines into HealthMetrics file.

rajnikantsh:
Resolves #3390

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3428)


JustinPihony (approved):
It should be spot checked in a live run or two, but otherwise seems good

PR #3429, opened 2021-12-23
Fix GH alt route flaky test

alex-vv:
Previous version was running full beamville simulation and verified that there were 2 alternative routes generated for GH. Unfortunately this was not fully deterministic, on CI in some rare cases last route was not generated, probably because RoutingRequest has different parameters. 

This new test uses one predefined RoutingRequest which is known to always return alternative route (if this config option is on) and sends it to RoutingWorker and check the resulting RoutingResponse

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3429)

### DIFF
--- a/src/test/scala/beam/router/GHRouterSpec.scala
+++ b/src/test/scala/beam/router/GHRouterSpec.scala
@@ -1,17 +1,29 @@
 package beam.router
 
 import akka.actor.ActorSystem
-import beam.sim.BeamHelper
-import beam.utils.ParquetReader
+import akka.pattern.ask
+import akka.testkit.TestActorRef
+import akka.util.Timeout
+import beam.agentsim.agents.vehicles.BeamVehicleType
+import beam.agentsim.agents.vehicles.VehicleProtocol.StreetVehicle
+import beam.agentsim.events.SpaceTime
+import beam.router.BeamRouter.{Location, RoutingRequest, RoutingResponse}
+import beam.router.Modes.BeamMode
+import beam.router.Modes.BeamMode._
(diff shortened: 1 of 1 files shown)


alex-vv:
I will make a few runs on CI to check if it is ok

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

src/test/scala/beam/router/GHRouterSpec.scala
```diff
@@ -29,25 +40,80 @@ class GHRouterSpec extends AnyWordSpecLike with Matchers with BeamHelper {
...
+            true,
+            true
+          ),
+          StreetVehicle(
+            Id.createVehicleId("body-2"),
+            Id.create("BODY-TYPE-DEFAULT", classOf[BeamVehicleType]),
+            SpaceTime(167138.0, 2234.0, 36952),
+            WALK,
+            true,
+            false
```
dimaopen: For better readability could you add names to the params that are not clear from the context?

dimaopen (approved):

src/test/scala/beam/router/GHRouterSpec.scala
alex-vv: done

alex-vv:
test!

PR #3432, opened 2021-12-30
Fix of wrong logical error message

### LINKED ISSUE #3426: Error: Customer has made two RideHail Inquiries
20:42:39.488 ERROR b.a.agents.ridehail.RideHailManager - Customer 016801-2012001020283-0-7430620 has made two RideHail Inquiries, with the departAt for the second being before or equal to the first: (27000 < 25600. This is likely to cause logical errors.

dimaopen:
We also need to validate the order of the RH requests before warning about wrong depart times. We also save the latest RH response to the cache.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432)


haitamlaarabi (changes requested):
> Reviewed 1 of 1 files at r1.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @dimaopen, @JustinPihony, and @nikolayilyin)

___
*[src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala, line 615 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432#-MtqTW0jAbgNhdPCjeMQ:-MtqTW0jAbgNhdPCjeMR:b2ixkp8) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e52b177d2862dcb11866ac3de71f55599252c8f9/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala#L615)):*
> ```Scala
>            * as a mode, there is no simple way to remove that cached response inside the RHM. We can still safely
>            * overwrite the response below with the latest.
>            * Sometimes inquiries get to this point not in order they were sent (because of the router with multiple
> ```

DO you mean here that we are seeing this error because there were two requests from different customers and not requests from same customer?
___
*[src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala, line 625 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432#-MtqTjwZ5oktwG22bmer:-MtqTjwZ5oktwG22bmes:b-5v35d3) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e52b177d2862dcb11866ac3de71f55599252c8f9/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala#L625)):*
> ```Scala
>         case _ =>
>       }
>       if (maybePreviousResponse.isEmpty || maybePreviousResponse.get.request.requestId < request.requestId) {
> ```

Also here you add them to the cache only when the request is coming from a different customer?
Which brings me to 

"But I didn't quite get: if we have multiple RH inquiries but store in the cache only the latest response how we can be sure that the customer goes with the latest one later when they are reserving a ride?"

I think we should test it thoroughly to see which request the customer is going with if we keep only the latest one. If it is conflicting then we probably should keep all the requests or develop a timer mechanism that delete requests that has been cached for too long.

dimaopen (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432)* status: all files reviewed (commit messages unreviewed), 2 unresolved discussions (waiting on @haitamlaarabi, @JustinPihony, and @nikolayilyin)

___
*[src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala, line 615 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432#-MtqTW0jAbgNhdPCjeMQ:-MuUP8Ff7h_pB2exLRRW:bayzwdj) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e52b177d2862dcb11866ac3de71f55599252c8f9/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala#L615)):*

DO you mean here that we are seeing this error because there were two requests from different customers and not requests from same customer?
</blockquote></details>

No, requests are from the same customer that was sent about at the same time. 
___
*[src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala, line 625 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432#-MtqTjwZ5oktwG22bmer:-MuUPwMlCGyIcptsIy0I:bsmfyme) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e52b177d2862dcb11866ac3de71f55599252c8f9/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala#L625)):*

Also here you add them to the cache only when the request is coming from a different customer?
Which brings me to 

"But I didn't quite get: if we have multiple RH inquiries but store in the cache only the latest response how we can be sure that the customer goes with the latest one later when they are reserving a ride?"

I think we should test it thoroughly to see which request the customer is going with if we keep only the latest one. If it is conflicting then we probably should keep all the requests or develop a timer mechanism that delete requests that has been cached for too long.
</blockquote></details>

The requests are from the same customer. I add a request in the cache only if it's newer. Though I think that you are right here. We need to keep multiple requests in the cache as you suggested. I'll add it.

haitamlaarabi:
Test now!

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -602,24 +602,29 @@ class RideHailManager(
...
-          }
-        case None =>
+          log.error(
+            s"Customer ${request.customer.personId} has made two RideHail Inquiries, with the departAt for the " +
+            s"second being before or equal to the first: (${previousRideHailResponse.request.departAt} < ${request.departAt}. This is " +
+            s"likely to cause logical errors."
+          )
+        case _ =>
+      }
+      if (maybePreviousResponse.isEmpty || maybePreviousResponse.get.request.requestId < request.requestId) {
```
JustinPihony: Should we log (even if at debug) if a previous response is being overridden?

JustinPihony (approved):

JustinPihony:
> Though I think that you are right here. We need to keep multiple requests in the cache as you suggested. I'll add it.

The only I can see to do that would be to store the value as a Vector

haitamlaarabi:
@dimaopen for some reason the tests are failing here. Strange!

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
dimaopen: Yes, I think this is right. Though if we keep multiple requests then no override happens.

haitamlaarabi: Let's keep the original way of doing it.. that is storing multiple requests. Also it would be great to know how bad for memory to keep all these requests

dimaopen:
Test!

dimaopen:
Test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -1281,6 +1258,26 @@ class RideHailManager(
...
+  private def getActualResponses(request: RideHailRequest) = {
+    val previousResponses = rideHailResponseCache.getOrElse(request.customer, IndexedSeq.empty)
+    val currentTime = request.requestTime.getOrElse(0)
+    previousResponses.filter(_.request.departAt >= currentTime)
+  }
+
+  private def removeOriginalResponseFromCache(request: RideHailRequest) = {
+    val actualResponses: IndexedSeq[RideHailResponse] = getActualResponses(request)
+    val departureTimeOrdering =
+      Ordering.by((rsp: RideHailResponse) => Math.abs(request.departAt - rsp.request.departAt))
```
JustinPihony, line 1272: Is `Math.abs` the correct option? This will potentially result in lower values being made into higher values (i.e. request's depart = 5 and depart responses of 3 and 6 you get 2 and 1, respectively - so the higher, 6 is chosen). Plus, see my comment on the reduction, if it is needed or not

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -1281,6 +1258,26 @@ class RideHailManager(
...
+    val maybeOriginalResponse = actualResponses.reduceOption(departureTimeOrdering.min)
```
JustinPihony, line 1273: If choosing the minimum, then `if(actualResponses.isEmpty) None else actualResponses.minBy(_.departAt)` seems a better option?

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -604,29 +604,7 @@ class RideHailManager(
...
-            s"Customer ${request.customer.personId} has made two RideHail Inquiries, with the departAt for the " +
-            s"second being before or equal to the first: (${previousRideHailResponse.request.departAt} < ${request.departAt}. This is " +
-            s"likely to cause logical errors."
-          )
-        case _ =>
-      }
-      if (maybePreviousResponse.isEmpty || maybePreviousResponse.get.request.requestId < request.requestId) {
-        rideHailResponseCache.put(request.customer, rideHailResponse)
-      }
+      rideHailResponseCache.put(request.customer, getActualResponses(request) :+ rideHailResponse)
```
JustinPihony, line 609: Is there still a possible issue with duplicate ridehail inquiries? Should we still log something? I guess since there are two, then we at least expect both to be removed at some point - or should we remove multiple when removing?

JustinPihony (changes requested):
Some of the changes now require review

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
dimaopen: We need to choose the closest by time (and maybe origin and destination locations) original response. Generally, a person during mode choice makes multiple RH requests (direct RH trip, and 2 RH legs for a RideHailTansit trip). Then the person chooses a trip and makes a Reservation request that is matched to the original response here. It's not necessarily the earliest one.
In your example, 6 should be chosen.

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
dimaopen: In this case, we would have 2 branches instead of one.

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
dimaopen: Now we keep all the actual ridehail inquiries we cannot find logical errors at this point.
We do remove not actual RH inquiries. Method `getActualResponses` filter out outdated responses.

haitamlaarabi:
Test now!

dimaopen:
Test!

JustinPihony (approved):

haitamlaarabi (approved):
> Reviewed 1 of 2 files at r2, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @haitamlaarabi, @JustinPihony, and @nikolayilyin)

___
*[src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala, line 1271 at r2](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432#-MsATeGDZh2tcsSqrQRs-r2-1271:-MwDY8Pg1qEdhQSXG1uh:b-617725) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/1ff2aca16fc899af322cd9545e926ac20bd1ab79/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala#L1271)):*

In this case, we would have 2 branches instead of one.
</blockquote></details>

Following up your example, with minor changes: 2 direct RH trip, and 2 RH legs for a RideHailTansit trip.
This code would remove "2 RH legs for a RideHailTansit trip" from cache if the actual decision made by the agent is a "direct RH trip"?

dimaopen (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432)* status: 1 of 2 files reviewed, 6 unresolved discussions (waiting on @haitamlaarabi, @JustinPihony, and @nikolayilyin)

___
*[src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala, line 1271 at r2](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3432#-MsATeGDZh2tcsSqrQRs-r2-1271:-MwQczaw2oGRxyNq4sr6:b-xhx1qw) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/1ff2aca16fc899af322cd9545e926ac20bd1ab79/src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala#L1271)):*

Following up your example, with minor changes: 2 direct RH trip, and 2 RH legs for a RideHailTansit trip.
This code would remove "2 RH legs for a RideHailTansit trip" from cache if the actual decision made by the agent is a "direct RH trip"?
</blockquote></details>

These requests aren't removed at this time because we don't have a clear sign here (in RideHailManager) that they aren't needed anymore. But they are going to be removed later (when they are expired).

dimaopen:
Test!

dimaopen:
Test!

PR #3443, opened 2022-01-15
Removing vehicle manager from vehicles and renaming energy conservation params

haitamlaarabi:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3443)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -40,9 +40,6 @@ import beam.router.{Modes, RoutingWorker}
 trait ChoosesMode {
   this: PersonAgent => // Self type restricts this trait to only mix into a PersonAgent
 
-  private val emergencyGeoId =
-    GeoLevel.getSpecialGeoIds(beamServices.beamConfig.beam.agentsim.taz.parkingManager.level)._1
-
   val dummyRHVehicle: StreetVehicle = createDummyVehicle(
     "dummyRH",
     beamServices.beamConfig.beam.agentsim.agents.rideHail.initialization.procedural.vehicleTypeId,
@@ -84,7 +81,7 @@ trait ChoosesMode {
     mode: BeamMode,
     asDriver: Boolean,
     needsToCalculateCost: Boolean
-  ) =
+  ): StreetVehicle =
     StreetVehicle(
       Id.create(id, classOf[BeamVehicle]),
       Id.create(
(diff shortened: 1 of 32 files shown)


haitamlaarabi:
Test now!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -725,7 +722,7 @@ trait ChoosesMode {
               seq :+ (vehicleOnTrip -> ParkingInquiry.init(
                 SpaceTime(geo.wgs2Utm(leg.beamLeg.travelPath.endPoint.loc), leg.beamLeg.endTime),
                 nextAct.getType,
-                VehicleManager.getReservedFor(veh.vehicleManagerId.get).get,
+                VehicleManager.getReservedFor(this.vehicleManagerId).get,
```
dimaopen, line 725: What if we use a shared vehicle that has its own vehicle manager? In this case, we receive a wrong parking stall.

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -113,7 +113,7 @@ trait ChoosesParking extends {
     val parkingInquiry = ParkingInquiry.init(
       SpaceTime(destinationUtm, lastLeg.beamLeg.endTime),
       nextActivityType,
-      VehicleManager.getReservedFor(currentBeamVehicle.vehicleManagerId.get).get,
+      VehicleManager.getReservedFor(this.vehicleManagerId).get,
```
dimaopen, line 116: Again: what if we use a shared vehicle here?

haitamlaarabi (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3443)* status: 0 of 32 files reviewed, 2 unresolved discussions (waiting on @dimaopen and @haitamlaarabi)

___
*[src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala, line 725 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3443#-MtREQFeeAz4HM7ayTnG-r1-725:-MtqPM6dAcoF6ZCsjijQ:b5uhlg6) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/42969f980bdbf583f34e042d9099c57f4a7b083b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L725)):*

What if we use a shared vehicle that has its own vehicle manager? In this case, we receive a wrong parking stall.
</blockquote></details>

This is a good point. Let's not merge this until I figure out whether there is a workaround or we just have to keep it as it is!

haitamlaarabi:
Do not merge #WIP

JustinPihony:
Bumping to a draft since it is a WIP

PR #3447, opened 2022-01-19
Skim improvements

### LINKED ISSUE #3438: Skims improvements: Transit, ridehail, parking

## Transit

## Freight (2-d skims)
## Freight (1-d skim)

dimaopen:
Added failedTrips field to ODSkims. One can use failedTrips/observations ratio to evaluate probabilty of succesful trip.
If trip is TRANSIT then the trip starts right at the activity ends in order to take into account waiting for the transit vehicle.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3447)


src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1318,8 +1361,20 @@ trait ChoosesMode {
...
+    val pendingTrip = data.pendingChosenTrip.get
     val (tick, triggerId) = releaseTickAndTriggerId()
+    val chosenTrip =
+      if (
+        pendingTrip.tripClassifier.isTransit
+        && pendingTrip.legs.head.beamLeg.startTime > tick
+      ) {
+        //we need to start trip as soon as our activity finishes (current tick) in order to
+        //correctly show waiting time for the transit in the OD skims
+        pendingTrip.updatePersonalLegsStartTime(tick)
```
zneedell: Just out of curiosity--would we expect this to give different behavior if instead of the tick we gave the previous activity end time? Would that be more robust to parallelism issues and/or introduce other problems?

zneedell (commented):
One non-blocking question, but LGTM other than that.

I think it probably makes sense to do what you have here and not directly penalize failed trips in the skims (but rather to just keep track of failed trips and penalize on ActivitySim's side), but I'll confirm with them that that should work

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
dimaopen: This tick is the same as the Activity end time. But we could check that the Activity end time is not in the past and just use it here. I'll change it. I can imagine that the Mode Choice may happen not at the activity end time but at some time before that. In this case, using activity end time is better. But we shouldn't use a tick in the past here because it causes issues with the scheduler.

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

zneedell (approved):
LGTM pending any comments from @JustinPihony

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1209,6 +1214,14 @@ trait ChoosesMode {
         parkingResponses
       ) ++ rideHail2TransitIinerary.toVector
 
+      choosesModeData.personData.currentTourMode match {
+        case Some(mode) if mode.isTransit && !combinedItinerariesForChoice.exists(_.tripClassifier == mode) =>
```
JustinPihony: To help with readability and understanding why this is considered a failed event I suggest rename `mode` to `expectedMode` - or maybe turn the `exists` into a `def` to control the readability?

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1463,10 +1522,10 @@ trait ChoosesMode {
               vehiclesUsed.headOption.filter(mustBeDrivenHome).map(_.id)
             else
               data.personData.currentTourPersonalVehicle
-                .orElse(vehiclesUsed.headOption.filter(mustBeDrivenHome).map(_.id))
+                .orElse(vehiclesUsed.headOption.filter(mustBeDrivenHome).map(_.id)),
+          failedTrips = data.personData.failedTrips ++ data.personData.currentTrip
```
JustinPihony, line 1528: If I am reading this correct, then any trip that is not a teleportation event should be considered a failure?

src/main/scala/beam/router/skim/CsvSkimReader.scala
```diff
@@ -36,7 +36,7 @@ class CsvSkimReader(
       skimsTryMap match {
         case Success(skimMap) => skimMap
         case Failure(ex) =>
-          logger.warn(s"Could not load warmStart skim from '$aggregatedSkimsFilePath'", ex)
+          logger.warn(s"Could not load warmStart skim from '$aggregatedSkimsFilePath': ${ex.getMessage}", ex)
```
JustinPihony: Doesn't the second parameter do this already?

JustinPihony (commented):
There is one question I'd like cleared up about failedTrips before approving

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
dimaopen: Generally, this is true. If we have some trip defined (`currentTrip` is an `Option`) that we cannot finish (we're choosing a new trip now (in `FinishingModeChoice` state)) then the currentTrip is considered a failure.
About teleportation, I'm not sure but it seems that they suppose that no teleportation failure is possible.

dimaopen:
Test!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
JustinPihony: I will defer to @zneedell on that aspect - and he already approved - so otherwise LGTM

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test now!

PR #3451, opened 2022-01-24
Rrp/#3310 seattle blocks csv and link radius meters param changes

heyrutvik:
It covers:
1. changes to generate blocks csv for urbansim_v2 inputs
2. changes to use configured (beam parameter) linkRadiusMeters for split operations
3. updates `r5` version

`r5` library release: https://github.com/LBNL-UCB-STI/r5/releases/tag/v4.5.3_linkRadiusMeters

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3451)

### DIFF
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -883,7 +883,7 @@ beam.routing {
     }
     bikeLaneScaleFactor = "double | 1.0"
     bikeLaneLinkIdsFilePath = ""
-    linkRadiusMeters = "double | 9000.0"
+    linkRadiusMeters = "double | 10000.0"
   }
   gh {
     useAlternativeRoutes = "boolean | false"
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -883,6 +883,7 @@ beam.routing {
     }
     bikeLaneScaleFactor = "double | 1.0"
     bikeLaneLinkIdsFilePath = ""
+    linkRadiusMeters = "double | 9000.0"
   }
   gh {
     useAlternativeRoutes = "boolean | false"
--- a/src/main/scala/beam/agentsim/agents/freight/input/NRELFreightReader.scala
+++ b/src/main/scala/beam/agentsim/agents/freight/input/NRELFreightReader.scala
@@ -3,6 +3,7 @@ package beam.agentsim.agents.freight.input
 import beam.agentsim.agents.freight._
 import beam.agentsim.agents.vehicles.{BeamVehicle, BeamVehicleType}
 import beam.sim.common.GeoUtils
+import beam.sim.config.BeamConfig
 import beam.sim.config.BeamConfig.Beam.Agentsim.Agents.Freight
 import beam.utils.csv.GenericCsvReader
 import beam.utils.matsim_conversion.MatsimPlanConversion.IdOps
@@ -14,8 +15,13 @@ import org.matsim.households.Household
 
 import scala.util.Random
 
-class NRELFreightReader(val config: Freight, val geoUtils: GeoUtils, rnd: Random, streetLayer: StreetLayer)
-    extends LazyLogging
+class NRELFreightReader(
+  val config: Freight,
+  val geoUtils: GeoUtils,
+  rnd: Random,
+  streetLayer: StreetLayer,
+  val beamConfig: BeamConfig
+) extends LazyLogging
     with FreightReader {
 
   val freightIdPrefix = "freight"
@@ -30,7 +36,7 @@ class NRELFreightReader(val config: Freight, val geoUtils: GeoUtils, rnd: Random
 
   private def findClosestUTMPointOnMap(utmCoord: Coord): Option[Coord] = {
     val wsgCoord = geoUtils.utm2Wgs(utmCoord)
-    val theSplit = geoUtils.getR5Split(streetLayer, wsgCoord)
+    val theSplit = geoUtils.getR5Split(streetLayer, wsgCoord, beamConfig.beam.routing.r5.linkRadiusMeters)
     if (theSplit == null) {
       None
     } else {
--- a/src/main/scala/beam/agentsim/agents/freight/input/NRELFreightReader.scala
+++ b/src/main/scala/beam/agentsim/agents/freight/input/NRELFreightReader.scala
@@ -20,7 +20,7 @@ class NRELFreightReader(
   val geoUtils: GeoUtils,
   rnd: Random,
   streetLayer: StreetLayer,
-  val beamConfig: BeamConfig
+  val r5LinkRadiusMeters: Double
 ) extends LazyLogging
     with FreightReader {
 
@@ -36,7 +36,7 @@ class NRELFreightReader(
 
   private def findClosestUTMPointOnMap(utmCoord: Coord): Option[Coord] = {
     val wsgCoord = geoUtils.utm2Wgs(utmCoord)
-    val theSplit = geoUtils.getR5Split(streetLayer, wsgCoord, beamConfig.beam.routing.r5.linkRadiusMeters)
+    val theSplit = geoUtils.getR5Split(streetLayer, wsgCoord, r5LinkRadiusMeters)
     if (theSplit == null) {
       None
     } else {
--- a/src/main/scala/beam/utils/data/synthpop/ScenarioGenerator.scala
+++ b/src/main/scala/beam/utils/data/synthpop/ScenarioGenerator.scala
@@ -613,27 +613,22 @@ object SimpleScenarioGenerator extends StrictLogging {
 
     val parkingFilePath = s"$pathToOutput/taz-parking.csv"
     CsvParkingInfoWriter.write(parkingFilePath, gen.geoSvc, scenarioResult.geoIdToAreaResidentsAndWorkers)
-    println(s"Wrote parking information to $parkingFilePath")
 
     val households = generatedData.map(_._1).toVector
     val householdFilePath = s"$pathToOutput/households.csv"
     CsvHouseholdInfoWriter.write(householdFilePath, households)
-    logger.info(s"Wrote households information to $householdFilePath")
 
     val blocks = generatedData.map(_._3).toVector
     val blockFilePath = s"$pathToOutput/blocks.csv"
     CsvBlocksInfoWriter.write(blockFilePath, blocks)
-    logger.info(s"Wrote blocks information to $blockFilePath")
 
     val persons = generatedData.flatMap(_._2.map(_.person)).toVector
     val personsFilePath = s"$pathToOutput/persons.csv"
     CsvPersonInfoWriter.write(personsFilePath, persons)
-    logger.info(s"Wrote persons information to $personsFilePath")
 
     val planElements = generatedData.flatMap(_._2.flatMap(_.plans)).toVector
     val plansFilePath = s"$pathToOutput/plans.csv"
     CsvPlanElementWriter.write(plansFilePath, planElements)
-    logger.info(s"Wrote plans information to $plansFilePath")
 
     val geoUtils: GeoUtils = new GeoUtils {
       override def localCRS: String = parsedArgs.localCRS
@@ -662,7 +657,7 @@ object SimpleScenarioGenerator extends StrictLogging {
     '--defaultValueOfTime', '8.0',
     '--localCRS', 'EPSG:3084',
     '--outputFolder', 'D:/Work/beam/NewYork/results',
-    '--linkRadiusMeters', '9000.0'
+    '--linkRadiusMeters', '10000.0'
     "] -PlogbackCfg=logback.xml
      */
     ProfilingUtils.timed("Scenario generation", x => logger.info(x)) {
--- a/src/main/scala/beam/utils/data/synthpop/ScenarioGenerator.scala
+++ b/src/main/scala/beam/utils/data/synthpop/ScenarioGenerator.scala
@@ -4,7 +4,6 @@ import java.io.File
 import java.time.LocalDateTime
 import java.time.format.DateTimeFormatter
 import java.util.concurrent.atomic.AtomicInteger
-
 import beam.agentsim.infrastructure.geozone._
 import beam.sim.common.GeoUtils
 import beam.sim.population.PopulationAdjustment
@@ -24,6 +23,7 @@ import beam.utils.data.synthpop.models.Models
 import beam.utils.data.synthpop.models.Models.{BlockGroupGeoId, County, Gender, GenericGeoId, State, TazGeoId}
 import beam.utils.scenario._
 import beam.utils.scenario.generic.writers.{
+  CsvBlocksInfoWriter,
   CsvHouseholdInfoWriter,
   CsvParkingInfoWriter,
   CsvPersonInfoWriter,
@@ -49,7 +49,7 @@ case class PersonWithExtraInfo(
 case class PersonWithPlans(person: PersonInfo, plans: List[PlanElement])
 
 case class ScenarioResult(
-  householdWithTheirPeople: Iterable[(HouseholdInfo, List[PersonWithPlans])],
+  householdWithTheirPeople: Iterable[(HouseholdInfo, List[PersonWithPlans], BlockInfo)],
   geoIdToAreaResidentsAndWorkers: Map[GenericGeoId, (Int, Int)]
 )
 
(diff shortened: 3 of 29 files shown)


heyrutvik:
test!

JustinPihony:
image

src/main/scala/beam/agentsim/agents/freight/input/NRELFreightReader.scala
```diff
@@ -30,7 +36,7 @@ class NRELFreightReader(val config: Freight, val geoUtils: GeoUtils, rnd: Random
 
   private def findClosestUTMPointOnMap(utmCoord: Coord): Option[Coord] = {
     val wsgCoord = geoUtils.utm2Wgs(utmCoord)
-    val theSplit = geoUtils.getR5Split(streetLayer, wsgCoord)
+    val theSplit = geoUtils.getR5Split(streetLayer, wsgCoord, beamConfig.beam.routing.r5.linkRadiusMeters)
```
JustinPihony: It seems a bit heavy-handed to pass the whole config when we really just need the one item. Shouldn't that be what is passed into the freightreader?

src/main/scala/beam/utils/data/synthpop/models/Models.scala
```diff
@@ -40,7 +40,7 @@ object Models {
   }
 
   case class BlockGroupGeoId(state: State, county: County, tract: String, blockGroup: String) extends GenericGeoId {
-    def asUniqueKey: String = s"${state.value}-${county.value}-$tract-$blockGroup"
+    def asUniqueKey: String = s"${state.value}${county.value}$tract$blockGroup"
```
JustinPihony, line 43: Will this create compatibility issues? Why make the change?

src/main/scala/beam/utils/map/R5MapStatsCalculator.scala
```diff
@@ -16,8 +16,10 @@ object R5MapStatsCalculator {
     // LoggerFactory.getLogger("com.conveyal").asInstanceOf[Logger].setLevel(Level.INFO)
     val pathToOsm = args(0)
     val minimumTagFrequencyToPrintTagOut = 1000
+    // TODO use `beamConfig.beam.routing.r5.linkRadiusMeters`
```
JustinPihony: Why not set it as part of this PR? Or was this a note for you to remember before closing the PR? Either way, this seems like it should be addressed.

src/main/resources/beam-template.conf
```diff
@@ -883,6 +883,7 @@ beam.routing {
     }
     bikeLaneScaleFactor = "double | 1.0"
     bikeLaneLinkIdsFilePath = ""
+    linkRadiusMeters = "double | 9000.0"
```
JustinPihony: The default seemed to be 10000, any reason for the change here? It is likely the cause for the failing tests.

src/main/scala/beam/utils/scenario/generic/writers/CsvBlocksInfoWriter.scala
```diff
@@ -0,0 +1,29 @@
...
+    try {
+      xs.foreach { household =>
+        csvWriter.write(
+          household.blockId.id,
+          household.x,
+          household.y
+        )
+      }
+    } finally {
+      Try(csvWriter.close())
```
JustinPihony: Why wrap this in a `Try`. If it is a non-fatal event then we should at least log a failure, and I see no point in using up an object if it isn't being passed back - so a regular try/catch should work.

src/main/scala/beam/utils/data/synthpop/ScenarioGenerator.scala
```diff
@@ -607,6 +620,11 @@ object SimpleScenarioGenerator extends StrictLogging {
     CsvHouseholdInfoWriter.write(householdFilePath, households)
     logger.info(s"Wrote households information to $householdFilePath")
 
+    val blocks = generatedData.map(_._3).toVector
+    val blockFilePath = s"$pathToOutput/blocks.csv"
+    CsvBlocksInfoWriter.write(blockFilePath, blocks)
+    logger.info(s"Wrote blocks information to $blockFilePath")
```
JustinPihony: Should this logger be added to the writer directly?

src/main/resources/beam-template.conf
heyrutvik: True, it is the reason behind failing tests. The reason we reduced the value is that we assumed that it will work for all cities. In Seattle, we were facing some issues due to 10k value. @nikolayilyin and I agreed to make it 9k.

We have two options:
1. we can update failing test to use 10k value for linkRadiusMeters,
2. make 10k default, and use 9k in seattle base conf

Thoughts?

src/main/scala/beam/agentsim/agents/freight/input/NRELFreightReader.scala
heyrutvik: Makes sense. Thanks.

src/main/scala/beam/utils/data/synthpop/models/Models.scala
heyrutvik: So it looks like I made this change to make the unique `Long` value. At the time, I had little understanding of urbansim_v2 and I reviewed [this csv](https://github.com/LBNL-UCB-STI/beam-data-austin/blob/develop/urbansim_v2/blocks.csv.gz) to get more context. The original hyphenated string is used in `PumaLevelScenarioGenerator` which I'm not familiar with. The `ScenarioGenerator` class is our focus.

I think we can revert this and the type of a value should not be the problem, I'll try to generate a plan using the old format and see if it causes any problem or not to run simulation with. I'll get back to you on this one.

src/main/scala/beam/utils/map/R5MapStatsCalculator.scala
heyrutvik: The comment was meant for someone who try to use this class and run across radius related errors. I should have removed the `TODO` from in front of it. It seems original author is @nikolayilyin, and he can chime in to suggest how we want to do this. 

I believe we don't want to pass beam config to the program just to get linkRadiusMeters value. 

We can do one of the followings,
1. add `args(1)` to pass linkRadiusMeters
2. elaborate comment for end user for existing change.

Thoughts?

src/main/scala/beam/utils/scenario/generic/writers/CsvBlocksInfoWriter.scala
heyrutvik: Apparently I copied it from other csv writers. We may want to change them all as following. Let me know what you think. Thanks.

```diff
diff --git a/src/main/scala/beam/utils/csv/CsvWriter.scala b/src/main/scala/beam/utils/csv/CsvWriter.scala
index ddc02b9a7..9e91ca0da 100644
--- a/src/main/scala/beam/utils/csv/CsvWriter.scala
+++ b/src/main/scala/beam/utils/csv/CsvWriter.scala
@@ -1,7 +1,10 @@
 package beam.utils.csv
 
+import com.typesafe.scalalogging.LazyLogging
+
 import java.io.Writer
 import org.matsim.core.utils.io.IOUtils
+
 import scala.util.Try
 
 case class Delimiter(value: String = ",")
@@ -12,7 +15,8 @@ class CsvWriter(
   headers: Seq[String],
   implicit val delimiter: Delimiter = Delimiter(),
   implicit val lineSeparator: LineSeparator = LineSeparator()
-) extends AutoCloseable {
+) extends AutoCloseable
+    with LazyLogging {
   def this(path: String, headers: String*) = this(path, headers)
 
   implicit val writer: Writer = IOUtils.getBufferedWriter(path)
@@ -42,9 +46,10 @@ class CsvWriter(
     CsvWriter.writeLineSeparator
   }
 
-  override def close(): Unit = {
-    Try(writer.close())
-  }
+  override def close(): Unit =
+    Try(writer.close()).failed.foreach { throwable =>
+      logger.error(s"Error while closing csv writer", throwable)
+    }
 
   def writeAllAndClose(rows: Iterable[Seq[Any]]): Try[Unit] = {
     val result = Try(rows.foreach(writeRow))
diff --git a/src/main/scala/beam/utils/scenario/generic/writers/CsvBlocksInfoWriter.scala b/src/main/scala/beam/utils/scenario/generic/writers/CsvBlocksInfoWriter.scala
index e06e30dab..8ffb11412 100644
--- a/src/main/scala/beam/utils/scenario/generic/writers/CsvBlocksInfoWriter.scala
+++ b/src/main/scala/beam/utils/scenario/generic/writers/CsvBlocksInfoWriter.scala
@@ -23,7 +23,7 @@ object CsvBlocksInfoWriter extends CsvBlocksInfoWriter {
         )
       }
     } finally {
-      Try(csvWriter.close())
+      csvWriter.close()
     }
   }
 }

diff --git a/src/main/scala/beam/utils/scenario/generic/writers/CsvParkingInfoWriter.scala b/src/main/scala/beam/utils/scenario/generic/writers/CsvParkingInfoWriter.scala
index 2bb37ca51..73c9656e3 100644
--- a/src/main/scala/beam/utils/scenario/generic/writers/CsvParkingInfoWriter.scala
+++ b/src/main/scala/beam/utils/scenario/generic/writers/CsvParkingInfoWriter.scala
@@ -47,7 +47,7 @@ object CsvParkingInfoWriter extends ParkingInfoWriter {
         }
       }
     } finally {
-      Try(csvWriter.close())
+      csvWriter.close()
     }
   }
 
diff --git a/src/main/scala/beam/utils/scenario/generic/writers/HouseholdInfoWriter.scala b/src/main/scala/beam/utils/scenario/generic/writers/HouseholdInfoWriter.scala
index 11b041b47..6454254e4 100644
--- a/src/main/scala/beam/utils/scenario/generic/writers/HouseholdInfoWriter.scala
+++ b/src/main/scala/beam/utils/scenario/generic/writers/HouseholdInfoWriter.scala
@@ -25,7 +25,7 @@ object CsvHouseholdInfoWriter extends HouseholdInfoWriter {
         )
       }
     } finally {
-      Try(csvWriter.close())
+      csvWriter.close()
     }
   }
 }
diff --git a/src/main/scala/beam/utils/scenario/generic/writers/PersonInfoWriter.scala b/src/main/scala/beam/utils/scenario/generic/writers/PersonInfoWriter.scala
index b8a1d8c25..76a99dfa2 100644
--- a/src/main/scala/beam/utils/scenario/generic/writers/PersonInfoWriter.scala
+++ b/src/main/scala/beam/utils/scenario/generic/writers/PersonInfoWriter.scala
@@ -28,7 +28,7 @@ object CsvPersonInfoWriter extends PersonInfoWriter {
         )
       }
     } finally {
-      Try(csvWriter.close())
+      csvWriter.close()
     }
   }
 }
diff --git a/src/main/scala/beam/utils/scenario/generic/writers/PlanElementWriter.scala b/src/main/scala/beam/utils/scenario/generic/writers/PlanElementWriter.scala
index d5273b7d7..32617b844 100644
--- a/src/main/scala/beam/utils/scenario/generic/writers/PlanElementWriter.scala
+++ b/src/main/scala/beam/utils/scenario/generic/writers/PlanElementWriter.scala
@@ -65,7 +65,7 @@ object CsvPlanElementWriter extends PlanElementWriter {
         )
       }
     } finally {
-      Try(csvWriter.close())
+      csvWriter.close()
     }
 
   }
```

src/main/scala/beam/utils/data/synthpop/models/Models.scala
heyrutvik: Found the reason. The `UrbansimReaderV2` expects block id to be a long value. Following are important symbols,
`beam.utils.scenario.urbansim.censusblock.entities.InputHousehold#blockId: Long`
`beam.utils.scenario.urbansim.censusblock.entities.Block#blockId: Long`

Error logs:
```
11:46:10.746 INFO  b.u.s.u.censusblock.UrbansimReaderV2 - Households info has been read successfully.
Exception in thread "main" java.lang.NumberFormatException: For input string: "53-033-006800-1"
	at java.lang.NumberFormatException.forInputString(NumberFormatException.java:65)
	at java.lang.Long.parseLong(Long.java:589)
```

Again we can do one of the following two things:
1. add another method/value to return long value of unique key
2. change above two mentioned symbols to be a string which will not be a breaking change (see following changes)

```diff
diff --git a/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/Block.scala b/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/Block.scala
index d53b79dcb..be735028d 100644
--- a/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/Block.scala
+++ b/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/Block.scala
@@ -5,7 +5,7 @@ import java.util
 import beam.utils.scenario.urbansim.censusblock.EntityTransformer
 
 case class Block(
-  blockId: Long,
+  blockId: String,
   x: Double,
   y: Double
 )
@@ -13,7 +13,7 @@ case class Block(
 object Block extends EntityTransformer[Block] {
 
   override def transform(rec: util.Map[String, String]): Block = {
-    val blockId = getIfNotNull(rec, "block_id").toLong
+    val blockId = getIfNotNull(rec, "block_id")
     val x = getIfNotNull(rec, "x").toDouble
     val y = getIfNotNull(rec, "y").toDouble
 
diff --git a/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/InputHousehold.scala b/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/InputHousehold.scala
index 127e9373d..2c0dc0787 100644
--- a/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/InputHousehold.scala
+++ b/src/main/scala/beam/utils/scenario/urbansim/censusblock/entities/InputHousehold.scala
@@ -8,7 +8,7 @@ case class InputHousehold(
   householdId: String,
   income: Int,
   cars: Int,
-  blockId: Long
+  blockId: String
 )
 
 object InputHousehold extends EntityTransformer[InputHousehold] {
@@ -17,7 +17,7 @@ object InputHousehold extends EntityTransformer[InputHousehold] {
     val householdId = getIfNotNull(rec, "household_id")
     val income = getIfNotNull(rec, "income").toInt
     val cars = getIfNotNull(rec, "cars").toInt
-    val blockId = getIfNotNull(rec, "block_id").toLong
+    val blockId = getIfNotNull(rec, "block_id")
 
     InputHousehold(householdId, income, cars, blockId)
   }
diff --git a/src/main/scala/beam/utils/scenario/urbansim/censusblock/merger/HouseholdMerger.scala b/src/main/scala/beam/utils/scenario/urbansim/censusblock/merger/HouseholdMerger.scala
index e0342ce2e..291ca3a8d 100644
--- a/src/main/scala/beam/utils/scenario/urbansim/censusblock/merger/HouseholdMerger.scala
+++ b/src/main/scala/beam/utils/scenario/urbansim/censusblock/merger/HouseholdMerger.scala
@@ -4,7 +4,7 @@ import beam.sim.population.PopulationAdjustment
 import beam.utils.scenario.urbansim.censusblock.entities.{Block, InputHousehold}
 import beam.utils.scenario.{HouseholdId, HouseholdInfo}
 
-class HouseholdMerger(blocks: Map[Long, Block]) extends Merger[InputHousehold, HouseholdInfo] {
+class HouseholdMerger(blocks: Map[String, Block]) extends Merger[InputHousehold, HouseholdInfo] {
   override def merge(iter: Iterator[InputHousehold]): Iterator[HouseholdInfo] = iter.map(inputToOutput)
 
   private def inputToOutput(inputHousehold: InputHousehold): HouseholdInfo = {
```

src/main/scala/beam/utils/data/synthpop/ScenarioGenerator.scala
heyrutvik: Sure. I'll do this once you go through https://github.com/LBNL-UCB-STI/beam/pull/3451#discussion_r791370805 comment. Thanks.

src/main/resources/beam-template.conf
JustinPihony: I am a fan of backwards compatibility, so I would go for leaving the default 10k and making the override to 9k in Seattle

src/main/scala/beam/utils/scenario/generic/writers/CsvBlocksInfoWriter.scala
JustinPihony: That looks good, except I would just use a regular `try`/`catch`. A `Try` object is more for when you need to pass it out of the function. Also we should avoid overly defensive code such as this anyway.

src/main/scala/beam/utils/map/R5MapStatsCalculator.scala
JustinPihony: I would say we should have it take args(1), but with a default of 10000.

src/main/scala/beam/utils/data/synthpop/models/Models.scala
JustinPihony: Interesting - I have no opinion then. As long as it doesn't break compatibility...which it sounds like it only HELPS compatibility.

heyrutvik:
test!

JustinPihony (approved):

PR #3452, opened 2022-01-24
RideHail and Parking skimmers

### LINKED ISSUE #3438: Skims improvements: Transit, ridehail, parking

dimaopen:
2 new skimmers that saves data to files #.skimsParking.csv.gz and
#.skimsRidehail.csv.gz

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3452)


dimaopen:
Do not merge: WIP, adding Freight skimmer here.

dimaopen:
Test!

JustinPihony:
Converted to a draft because it is not ready to merge as commented

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
This one can be reviewed

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -774,6 +793,18 @@ class PersonAgent(
...
+      eventsManager.processEvent(
+        new RideHailSkimmerEvent(
+          eventTime = tick,
+          tazId = beamScenario.tazTreeMap.getTAZ(req.pickUpLocationUTM).tazId,
+          reservationType = if (req.asPooled) Pooled else Solo,
+          waitTime = travelProposal.get.timeToCustomer(req.customer),
+          costPerMile =
+            travelProposal.get.estimatedPrice(req.customer.personId) / travelProposal.get.travelDistanceForCustomer(
+              req.customer
+            ) * 1609
```
zneedell: Might want to define this as a constant to clarify the units conversion

zneedell (changes requested):
LGTM with one small suggestion

src/test/scala/beam/agentsim/infrastructure/ChargingNetworkManagerSpec.scala
```diff
@@ -45,6 +45,7 @@ class ChargingNetworkManagerSpec
     with BeforeAndAfterEach {
 
   private val filesPath = s"${System.getenv("PWD")}/test/test-resources/beam/input"
+  logger.error("filesPath = {}", filesPath)
```
JustinPihony: Why is this an error logger?

dimaopen:
Test!

src/main/scala/beam/agentsim/agents/modalbehaviors/DrivesVehicle.scala
```diff
@@ -421,7 +421,22 @@ trait DrivesVehicle[T <: DrivingData] extends BeamAgent[T] with Stash with Expon
             goto(EnrouteRefueling) using data.asInstanceOf[T]
           } else goto(ConnectingToChargingPoint) using data.asInstanceOf[T]
         } else {
-          handleUseParkingSpot(tick, currentBeamVehicle, id, geo, eventsManager)
+          val maybePersonData = findPersonData(data)
+          val maybeNextActivity =
+            maybePersonData.flatMap(personData => this.asInstanceOf[PersonAgent].nextActivity(personData))
```
JustinPihony: What gives the certainty that this is a PersonAgent? It seems like a potential for a runtime exception?

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -55,10 +69,89 @@ object ChoosesParking {
...
+  }
+
+  private def createParkingSkimmerEvent(
+    tick: Int,
+    tazTreeMap: TAZTreeMap,
+    nextActivity: Option[Activity],
+    stall: ParkingStall,
+    restOfTrips: List[EmbodiedBeamLeg]
+  ): ParkingSkimmerEvent = {
+    val walkLeg = restOfTrips(1)
```
JustinPihony: There is no guarantee that there is a second item. How about `restOfTrips.lift(1)`, which will return `Option[EmbodiedLeg]`

JustinPihony: Or if it is important, then we should be more explicit in the throwing of an exception

src/main/scala/beam/router/skim/core/AbstractSkimmer.scala
```diff
@@ -157,12 +160,7 @@ abstract class AbstractSkimmer(beamConfig: BeamConfig, ioController: OutputDirec
       case e: AbstractSkimmerEvent if e.getEventType == eventType =>
         currentSkimInternal.compute(
           e.getKey,
-          (_, v) => {
-            val value =
```
JustinPihony, line 161: You can just return from the `if`/`else`. Or else I'd suggest a better name than `value`. As it stands I do not see much benefit to the `val`, as it would usually be for readability here.

JustinPihony (commented):
I am about to head off, so I wanted to submit my current review - I still have a little over half left.

src/main/scala/beam/router/skim/core/AbstractSkimmer.scala
dimaopen: I actually removed that `val value`. Option(null) creates None which allowed me to remove the `if/else`.

dimaopen:
Test!

src/main/scala/beam/router/skim/core/FreightSkimmer.scala
```diff
@@ -0,0 +1,100 @@
...
+  import FreightSkimmer._
+  override protected[skim] val readOnlySkim = new FreightSkims
+  override protected val skimFileBaseName: String = FreightSkimmer.fileBaseName
+
+  override protected val skimFileHeader =
+    "tazId,hour,numberOfLoadings,numberOfUnloadings,costPerMile,walkAccessDistanceInM,parkingCostPerHour,observations,iterations"
+  override protected val skimName: String = FreightSkimmer.name
+  override protected val skimType: Skims.SkimType.Value = Skims.SkimType.TC_SKIMMER
+
+  override protected def fromCsv(
```
JustinPihony, line 29: Should we handle missing data? Or let it crash

JustinPihony: Same for other skimmers

JustinPihony (approved):
One note not worth blocking

zneedell (approved):

src/main/scala/beam/router/skim/core/FreightSkimmer.scala
dimaopen: You are right about the possible crash here. Though I think that we need to handle those missed data that may not be written by our skimmer to the file. This `fromCsv` usually read just written files as I understand. It shouldn't contain missed data.

PR #3454, opened 2022-01-26
Freight OD Skims

### LINKED ISSUE #3438: Skims improvements: Transit, ridehail, parking

dimaopen:
Added FREIGHT mode that is used only in ODSkims. Payload weight goes to OD Skims for FREIGHT mode.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3454)


dimaopen:
Test!

haitamlaarabi (approved):
> Reviewed 1 of 12 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3454)* status: 1 of 12 files reviewed, all discussions resolved (waiting on @zneedell)

haitamlaarabi:
Looks good to go to me!

haitamlaarabi:
Test!

dimaopen:
Test!

PR #3456, opened 2022-01-29
Create household cars from Vehicle types

haitamlaarabi:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3456)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
+++ b/src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
@@ -17,13 +17,9 @@ import beam.agentsim.agents.ridehail.RideHailAgent.{
   ModifyPassengerScheduleAcks
 }
 import beam.agentsim.agents.ridehail.RideHailManager.RoutingResponses
-import beam.agentsim.agents.vehicles.{
-  BeamVehicle,
-  BeamVehicleType,
-  PassengerSchedule,
-  PersonIdWithActorRef,
-  VehicleManager
-}
+import beam.agentsim.agents.vehicles.EnergyEconomyAttributes.Powertrain
+import beam.agentsim.agents.vehicles.VehicleCategory.{Bike, Car, VehicleCategory}
+import beam.agentsim.agents.vehicles._
 import beam.agentsim.events.SpaceTime
 import beam.agentsim.infrastructure.{ParkingInquiry, ParkingInquiryResponse}
 import beam.agentsim.scheduler.BeamAgentScheduler.{CompletionNotice, ScheduleTrigger}
@@ -36,11 +32,13 @@ import beam.router.model.{BeamLeg, EmbodiedBeamLeg}
 import beam.router.osm.TollCalculator
 import beam.sim.config.BeamConfig.Beam
 import beam.sim.population.AttributesOfIndividual
+import beam.sim.vehicles.VehiclesAdjustment
 import beam.sim.{BeamScenario, BeamServices}
 import beam.utils.logging.LoggingMessageActor
 import beam.utils.logging.pattern.ask
 import com.conveyal.r5.transit.TransportNetwork
 import com.vividsolutions.jts.geom.Envelope
+import org.apache.commons.math3.distribution.UniformRealDistribution
 import org.matsim.api.core.v01.population.{Activity, Leg, Person}
 import org.matsim.api.core.v01.{Coord, Id}
 import org.matsim.core.api.experimental.events.EventsManager
@@ -52,6 +50,7 @@ import org.matsim.households.Household
 import java.util.concurrent.TimeUnit
 import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
+import scala.jdk.CollectionConverters.collectionAsScalaIterableConverter
 
 object HouseholdActor {
 
@@ -80,7 +79,8 @@ object HouseholdActor {
     sharedVehicleFleets: Seq[ActorRef] = Vector(),
     possibleSharedVehicleTypes: Set[BeamVehicleType],
     routeHistory: RouteHistory,
-    boundingBox: Envelope
+    boundingBox: Envelope,
+    vehiclesAdjustment: VehiclesAdjustment
   ): Props = {
     Props(
       new HouseholdActor(
@@ -102,7 +102,8 @@ object HouseholdActor {
         sharedVehicleFleets,
         possibleSharedVehicleTypes,
         routeHistory,
-        boundingBox
+        boundingBox,
+        vehiclesAdjustment
       )
     )
   }
@@ -111,6 +112,7 @@ object HouseholdActor {
     personId: Id[Person],
     whereWhen: SpaceTime,
     originActivity: Activity,
+    requireVehicleCategoryAvailable: Option[VehicleCategory],
     triggerId: Long
   ) extends HasTriggerId
   case class ReleaseVehicle(vehicle: BeamVehicle, triggerId: Long) extends HasTriggerId
@@ -149,11 +151,12 @@ object HouseholdActor {
     val population: org.matsim.api.core.v01.population.Population,
     val household: Household,
     vehicles: Map[Id[BeamVehicle], BeamVehicle],
-    homeCoord: Coord,
+    fallbackHomeCoord: Coord,
     sharedVehicleFleets: Seq[ActorRef] = Vector(),
     possibleSharedVehicleTypes: Set[BeamVehicleType],
     routeHistory: RouteHistory,
-    boundingBox: Envelope
+    boundingBox: Envelope,
+    vehiclesAdjustment: VehiclesAdjustment
   ) extends LoggingMessageActor
       with HasTickAndTrigger
       with ActorLogging {
@@ -188,15 +191,49 @@ object HouseholdActor {
     private var personAndActivityToCav: Map[(Id[Person], Activity), BeamVehicle] = Map()
     private var personAndActivityToLegs: Map[(Id[Person], Activity), List[BeamLeg]] = Map()
 
+    private val realDistribution: UniformRealDistribution = new UniformRealDistribution()
+    realDistribution.reseedRandomGenerator(beamScenario.beamConfig.matsim.modules.global.randomSeed)
+
     override def loggedReceive: Receive = {
 
       case TriggerWithId(InitializeTrigger(tick), triggerId) =>
+        val homeCoordFromPlans = household.members
+          .flatMap(per =>
+            per.getSelectedPlan.getPlanElements.asScala.flatMap {
+              case act: Activity if act.getType == "Home" => Some(act.getCoord)
+              case _                                      => None
+            }
+          )
+          .headOption
+          .getOrElse(fallbackHomeCoord)
+
         val vehiclesByCategory =
           vehicles.filter(_._2.beamVehicleType.automationLevel <= 3).groupBy(_._2.beamVehicleType.vehicleCategory)
-        val fleetManagers = vehiclesByCategory.map { case (category, vs) =>
+
+        //We should create a vehicle manager for cars and bikes for all households in case they are generated during the simulation
+
+        val vehiclesByAllCategories = List(Car, Bike)
+          .map(cat => cat -> vehiclesByCategory.getOrElse(cat, Map[Id[BeamVehicle], BeamVehicle]()))
+          .toMap
+        val fleetManagers = vehiclesByAllCategories.map { case (category, vs) =>
+          val emergencyGenerator = new EmergencyHouseholdVehicleGenerator(
+            household,
+            homeCoordFromPlans,
+            beamScenario,
+            vehiclesAdjustment,
+            category
+          )
           val fleetManager =
             context.actorOf(
-              Props(new HouseholdFleetManager(parkingManager, vs, homeCoord, beamServices.beamConfig.beam.debug)),
+              Props(
+                new HouseholdFleetManager(
+                  parkingManager,
+                  vs,
+                  homeCoordFromPlans,
+                  Some(emergencyGenerator),
+                  beamServices.beamConfig.beam.debug
+                )
+              ),
               category.toString
             )
           context.watch(fleetManager)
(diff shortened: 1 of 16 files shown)


haitamlaarabi:
Test!

src/main/scala/beam/sim/config/BeamConfig.scala
```diff
@@ -1528,6 +1528,7 @@ object BeamConfig {
           fractionOfInitialVehicleFleet: scala.Double,
           fractionOfPeopleWithBicycle: scala.Double,
           fuelTypesFilePath: java.lang.String,
+          generateEmergencyHouseholdVehicleWhenPlansRequireIt: scala.Boolean,
```
JustinPihony, line 1531: You are missing the template.conf change that caused this.

src/test/scala/beam/agentsim/agents/PersonWithPersonalVehiclePlanSpec.scala
```diff
@@ -562,6 +566,123 @@ class PersonWithPersonalVehiclePlanSpec
...
+      val scheduler = TestActorRef[BeamAgentScheduler](
+        SchedulerProps(
+          beamConfig,
+          stopTick = 24 * 60 * 60,
+          maxWindow = 10,
+          new StuckFinder(beamConfig.beam.debug.stuckAgentDetection)
+        )
+      )
+      val parkingManager = system.actorOf(Props(new TrivialParkingManager))
+      //val chargingNetworkManager = system.actorOf(Props(new ChargingNetworkManager(services, beamScenario, scheduler)))
```
JustinPihony, line 610: This can be removed I assume?

src/test/scala/beam/agentsim/agents/PersonWithPersonalVehiclePlanSpec.scala
```diff
@@ -562,6 +566,123 @@ class PersonWithPersonalVehiclePlanSpec
...
+      }
+
+      modeChoiceEvents.expectMsgType[ModeChoiceEvent]
+      modeChoiceEvents.expectMsgType[ModeChoiceEvent]
+
+      personEntersVehicleEvents.expectMsgType[PersonEntersVehicleEvent]
+      personEntersVehicleEvents.expectMsgType[PersonEntersVehicleEvent]
+      personEntersVehicleEvents.expectMsgType[PersonEntersVehicleEvent]
+
+      expectMsgType[CompletionNotice]
```
JustinPihony, line 682: Per the name, this test should also be testing against the vehicle, shouldn't it?

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -188,15 +191,49 @@ object HouseholdActor {
...
     private var personAndActivityToLegs: Map[(Id[Person], Activity), List[BeamLeg]] = Map()
 
+    private val realDistribution: UniformRealDistribution = new UniformRealDistribution()
+    realDistribution.reseedRandomGenerator(beamScenario.beamConfig.matsim.modules.global.randomSeed)
+
     override def loggedReceive: Receive = {
 
       case TriggerWithId(InitializeTrigger(tick), triggerId) =>
+        val homeCoordFromPlans = household.members
+          .flatMap(per =>
```
JustinPihony: Can we set the variable name to `person`? I'm never a fan of shortening if it hurts readability

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -188,15 +191,49 @@ object HouseholdActor {
...
+          .headOption
+          .getOrElse(fallbackHomeCoord)
+
         val vehiclesByCategory =
           vehicles.filter(_._2.beamVehicleType.automationLevel <= 3).groupBy(_._2.beamVehicleType.vehicleCategory)
-        val fleetManagers = vehiclesByCategory.map { case (category, vs) =>
+
+        //We should create a vehicle manager for cars and bikes for all households in case they are generated during the simulation
+
+        val vehiclesByAllCategories = List(Car, Bike)
```
JustinPihony, line 216: I've seen similar code before - should we create a central place that lists "AllCategories" for if/when it changes. Otherwise we'll have to do a copy/replace

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -188,15 +191,49 @@ object HouseholdActor {
...
+          .map(cat => cat -> vehiclesByCategory.getOrElse(cat, Map[Id[BeamVehicle], BeamVehicle]()))
+          .toMap
+        val fleetManagers = vehiclesByAllCategories.map { case (category, vs) =>
```
JustinPihony: Same naming comment here. vs should be vehicles. I know that code was existing, but I don't see why we can't fix it up :)

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -504,4 +541,66 @@ object HouseholdActor {
...
+                }
+            case VehicleCategory.Bike =>
+              beamScenario.vehicleTypes
+                .get(
+                  Id.create(
+                    beamScenario.beamConfig.beam.agentsim.agents.vehicles.dummySharedBike.vehicleTypeId,
+                    classOf[BeamVehicleType]
+                  )
+                )
+            case _ => None
```
JustinPihony: Should we log a warning here with the type attempted?

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
```diff
@@ -504,4 +541,66 @@ object HouseholdActor {
...
+                .orElse {
+                  beamScenario.vehicleTypes.get(
+                    Id.create(
+                      beamScenario.beamConfig.beam.agentsim.agents.vehicles.dummySharedCar.vehicleTypeId,
+                      classOf[BeamVehicleType]
+                    )
+                  )
+                }
+            case VehicleCategory.Bike =>
+              beamScenario.vehicleTypes
```
JustinPihony, line 583: Does this not have a first pass attempt the way a car does?

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -73,7 +70,13 @@ class HouseholdFleetManager(
 
     case NotifyVehicleIdle(vId, whenWhere, _, _, _, _) =>
       val vehId = vId.asInstanceOf[Id[BeamVehicle]]
-      vehicles(vehId).spaceTime = whenWhere
+      vehiclesInternal.get(vehId) match {
+        case Some(vehicle) =>
+          vehicle.spaceTime = whenWhere
+          logger.debug("updated vehicle {} with location {}", vehId, whenWhere)
+        case _ =>
+          logger.error("WHY ISN'T THIS VEHICLE HERE")
```
JustinPihony: Should we log the vehicle details? Maybe even the state of `vehiclesInternal`?

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -98,16 +101,63 @@ class HouseholdFleetManager(
...
-          Nil
+    case MobilityStatusInquiry(personId, whenWhere, _, requireVehicleCategoryAvailable, triggerId) =>
+      {
+        for {
+          neededVehicleCategory              <- requireVehicleCategoryAvailable
+          emergencyHouseholdVehicleGenerator <- maybeEmergencyHouseholdVehicleGenerator
+          vehicle                            <- emergencyHouseholdVehicleGenerator.createVehicle(personId, nextVehicleIndex, neededVehicleCategory)
+        } yield {
+          if (availableVehicles.isEmpty) {
+            // Create a vehicle out of thin air
```
JustinPihony: The warning logger should be moved up to this level and then you don't need the comment.

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -98,16 +101,63 @@ class HouseholdFleetManager(
...
-          rest
-        case Nil =>
-          sender() ! MobilityStatusResponse(Vector(), triggerId)
-          Nil
+    case MobilityStatusInquiry(personId, whenWhere, _, requireVehicleCategoryAvailable, triggerId) =>
+      {
+        for {
+          neededVehicleCategory              <- requireVehicleCategoryAvailable
+          emergencyHouseholdVehicleGenerator <- maybeEmergencyHouseholdVehicleGenerator
+          vehicle                            <- emergencyHouseholdVehicleGenerator.createVehicle(personId, nextVehicleIndex, neededVehicleCategory)
```
JustinPihony, line 108: Should we create the vehicle here? This should probably be pushed into the if, because if there are available vehicles then we hit the creation code for nothing.

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -98,16 +101,63 @@ class HouseholdFleetManager(
...
+              MobilityStatusResponse(Vector(ActualVehicle(vehicle)), otherTriggerId)
+            } pipeTo mobilityRequester
+          } else {
+            availableVehicles = availableVehicles match {
+              case firstVehicle :: rest =>
+                logger.debug("Vehicle {} is now taken", firstVehicle.id)
+                firstVehicle.becomeDriver(sender)
+                sender() ! MobilityStatusResponse(Vector(ActualVehicle(firstVehicle)), triggerId)
+                rest
+              case _ =>
```
JustinPihony, line 148: You can change the if to be a match and then the Nil case moves from an unnecessary logging to being the primary emergency generation. Note that you could still end up with a None vehicle in the scenario where no available vehicles and no emergency generator is configured to be available. How you want to handle that (logging?) is up to you.

JustinPihony (changes requested):
There are some items flagged which need handled before approval

haitamlaarabi:
Test now!

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
haitamlaarabi: Not required

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
haitamlaarabi: Done

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
haitamlaarabi: It's fine to end up with a None vehicle

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
haitamlaarabi: There is already a central place with all categories. This is a special case.

src/test/scala/beam/agentsim/agents/PersonWithPersonalVehiclePlanSpec.scala
haitamlaarabi: Not sure how to answer this

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
haitamlaarabi: Done

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
haitamlaarabi: okay

haitamlaarabi:
Test now!

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -73,7 +70,13 @@ class HouseholdFleetManager(
...
       val vehId = vId.asInstanceOf[Id[BeamVehicle]]
-      vehicles(vehId).spaceTime = whenWhere
+      vehiclesInternal.get(vehId) match {
+        case Some(vehicle) =>
+          vehicle.spaceTime = whenWhere
+          logger.debug("updated vehicle {} with location {}", vehId, whenWhere)
+        case _ =>
+          logger.error("Why is not this vehicle {} at location {} yet!", vehId, whenWhere)
+      }
       logger.debug("updated vehicle {} with location {}", vehId, whenWhere)
```
nikolayilyin: a duplication of debug log entries

nikolayilyin (approved):
a nitpick, otherwise lgtm

haitamlaarabi:
Test now!

JustinPihony:
> You added this into the `if` as per my last comment, but didn't remove it from the `for`

src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
JustinPihony: I guess what I meant is that instead of putting `List(Car, Bike)` directly to store it as a variable for use anywhere else it is needed:

`val allVehicleCategories = List(Car, Bike)`

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -98,16 +100,69 @@ class HouseholdFleetManager(
...
+                firstVehicle.becomeDriver(sender)
+                sender() ! MobilityStatusResponse(Vector(ActualVehicle(firstVehicle)), triggerId)
+                rest
+              case _ =>
+                logger.error(s"THE LIST OF VEHICLES SHOULDN'T BE EMPTY")
+                Nil
+            }
+          }
+        }
+      }.getOrElse {
```
JustinPihony, line 154: I am just now noticing that this is an `Option[Unit]`, which means it is basically working off of the fact whether the `for` did something at all. Is that what is wanted? If so, then maybe it is better to assign to a variable and then do an if isEmpty. Ultimately it is more about confusing a reader of the code, so it will be functionally the same and not worth blocking., though

JustinPihony (approved):
Approving, as my changes are not worth blocking, but maybe should be addressed at least

src/test/scala/beam/agentsim/agents/PersonWithPersonalVehiclePlanSpec.scala
JustinPihony: I guess I was wondering if there is a way to create an assertion that a last resort car was created? A stub that could be verified that the emergency creation was triggered - or a test of the vehicle attributes?

PR #3463, opened 2022-02-07

Allow to do a mode choice if the current mode cannot be implemented

### LINKED ISSUE #3462: Allow to do a mode choice if the current mode cannot be implemented
@zneedell: When someone has “**CAR**” as their tour mode and they’re in `choosingMode`, and they get the set of vehicles available from the fleet managers and they don’t have access to any cars, they go straight to a **WALK** trip. Would it make sense to instead send them back to mode choice so they have the option of choosing another mode? There might be some other situations in mode choice where we give up and force a **WALK** trip when it would make more sense to go back to choosing mode as well

dimaopen:
If there's no alternative to do a mode choice the process of mode choice starts again (without predefinded tour mode).

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3463)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -182,6 +182,7 @@ trait ChoosesMode {
             _,
             _,
             _,
+            _,
             _
           ) =>
         self ! MobilityStatusResponse(Vector(beamVehicles(vehicle)), getCurrentTriggerIdOrGenerate)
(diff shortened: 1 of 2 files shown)


dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

zneedell (approved):
Lgtm

zneedell:
I would be curious to confirm whether this leads to any big changes in a production run

dimaopen:
> I would be curious to confirm whether this leads to any big changes in a production run

I think that some changes would be but not too big

PR #3473, opened 2022-02-23
don't load previous plans as unselected

zneedell:
For some reason the previous unselected plans were being transformed into selected plans when being loaded, so people would end up with multiple sets of plans being selected and trying to complete two days' worth of trips, which causes chaos. Turning this off for now because it's not a super important feature

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3473)


zneedell:
Closing for now because it's a hack. Can eventually be fixed by https://github.com/LBNL-UCB-STI/beam/issues/3482

PR #3476, opened 2022-02-25
Current trip mode basing on current tour mode

### LINKED ISSUE #3441: Walk issue in ChoosesMode
There was a comment in `art/new-york-only-code` branch
image

It was removed before the branch was merged. It needs to be resolved one day.

dimaopen:
Instead of setting mode for the whole tour (Home -> Work -> Shopping -> Home) and use it for each individual trip now we use current tour mode for getting the right current trip mode.
Also HouseholdFleetManager doesn't return a personal vehicle to persons that are not at home.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3476)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/PersonAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/PersonAgent.scala
@@ -189,6 +189,7 @@ object PersonAgent {
     restOfCurrentTrip: List[EmbodiedBeamLeg] = List(),
     currentVehicle: VehicleStack = Vector(),
     currentTourMode: Option[BeamMode] = None,
+    currentTripMode: Option[BeamMode] = None,
     currentTourPersonalVehicle: Option[Id[BeamVehicle]] = None,
     passengerSchedule: PassengerSchedule = PassengerSchedule(),
     currentLegPassengerScheduleIndex: Int = 0,
@@ -475,6 +476,46 @@ class PersonAgent(
     }
   }
 
+  def isFirstTripWithinTour(personData: BasePersonData, nextAct: Activity): Boolean = {
+    val (tripIndexOfElement: Int, _) = currentTripIndexWithinTour(personData, nextAct)
+    tripIndexOfElement == 0
+  }
+
+  def isLastTripWithinTour(personData: BasePersonData, nextAct: Activity): Boolean = {
+    val (tripIndexOfElement: Int, lastTripIndex: Int) = currentTripIndexWithinTour(personData, nextAct)
+    tripIndexOfElement == lastTripIndex
+  }
+
+  def isFirstOrLastTripWithinTour(personData: BasePersonData, nextAct: Activity): Boolean = {
+    val (tripIndexOfElement: Int, lastTripIndex: Int) = currentTripIndexWithinTour(personData, nextAct)
+    tripIndexOfElement == 0 || tripIndexOfElement == lastTripIndex
+  }
+
+  def currentTripIndexWithinTour(personData: BasePersonData, nextAct: Activity): (Int, Int) = {
+    val tour = currentTour(personData)
+    val lastTripIndex = tour.trips.size - 1
+    val tripIndexOfElement = tour
+      .tripIndexOfElement(nextAct)
+      .getOrElse(throw new IllegalArgumentException(s"Element [$nextAct] not found"))
+    if (tripIndexOfElement == 0 && currentActivity(personData).getType != "Home")
+      logger.warn(
+        "~Activity {}, idx {}, id {}, next {}",
+        currentActivity(personData).getType,
+        tripIndexOfElement,
+        id,
+        nextAct.getType
+      )
+    if (tripIndexOfElement == lastTripIndex && nextAct.getType != "Home")
+      logger.warn(
+        "Cur ~Activity {}, idx {}, id {}, next {}",
+        currentActivity(personData).getType,
+        tripIndexOfElement,
+        id,
+        nextAct.getType
+      )
+    (tripIndexOfElement, lastTripIndex)
+  }
+
   def currentActivity(data: BasePersonData): Activity =
     _experiencedBeamPlan.activities(data.currentActivityIndex)
 
@@ -587,15 +628,10 @@ class PersonAgent(
         val nextCoord = nextActivity(data).get.getCoord
         goto(ChoosingMode) using ChoosesModeData(
           personData = data.copy(
-            // If the mode of the next leg is defined and is CAV, use it, otherwise,
-            // If we don't have a current tour mode (i.e. are not on a tour aka at home),
-            // use the mode of the next leg as the new tour mode.
-            currentTourMode = modeOfNextLeg match {
-              case Some(CAV) =>
-                Some(CAV)
-              case _ =>
-                data.currentTourMode.orElse(modeOfNextLeg)
-            },
+            // We do not stick to current tour mode
+            // If we have the currentTourPersonalVehicle then we should use it
+            // use the mode of the next leg as the new trip mode.
+            currentTripMode = modeOfNextLeg,
             numberOfReplanningAttempts = 0,
             failedTrips = IndexedSeq.empty,
             enrouteData = EnrouteData()
@@ -612,7 +648,7 @@ class PersonAgent(
   when(Teleporting) {
     case Event(
           TriggerWithId(PersonDepartureTrigger(tick), triggerId),
-          data @ BasePersonData(_, Some(currentTrip), _, _, _, _, _, _, false, _, _, _, _, _)
+          data @ BasePersonData(_, Some(currentTrip), _, _, _, _, _, _, _, false, _, _, _, _, _)
         ) =>
       endActivityAndDepart(tick, currentTrip, data)
 
@@ -626,7 +662,7 @@ class PersonAgent(
 
     case Event(
           TriggerWithId(TeleportationEndsTrigger(tick), triggerId),
-          data @ BasePersonData(_, Some(currentTrip), _, _, maybeCurrentTourMode, _, _, _, true, _, _, _, _, _)
+          data @ BasePersonData(_, Some(currentTrip), _, _, _, maybeCurrentTripMode, _, _, _, true, _, _, _, _, _)
         ) =>
       holdTickAndTriggerId(tick, triggerId)
 
@@ -639,7 +675,7 @@ class PersonAgent(
         startY = currentTrip.legs.head.beamLeg.travelPath.startPoint.loc.getY,
         endX = currentTrip.legs.last.beamLeg.travelPath.endPoint.loc.getX,
         endY = currentTrip.legs.last.beamLeg.travelPath.endPoint.loc.getY,
-        currentTourMode = maybeCurrentTourMode.map(_.value)
+        currentTourMode = maybeCurrentTripMode.map(_.value)
       )
       eventsManager.processEvent(teleportationEvent)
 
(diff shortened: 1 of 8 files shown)


src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -475,6 +478,30 @@ class PersonAgent(
     }
   }
 
+  def isFirstTripWithinTour(personData: BasePersonData, nextAct: Activity): Boolean = {
```
zneedell, line 481: I'm wondering if we could replace this check with instead checking if the current activity == home. I'm thinking of a case where someone did something like home -> work -> home -> shop -> home. We'd want them to be able to pick up their car for the second home -> shop -> home tour, for instance

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -587,15 +614,10 @@ class PersonAgent(
...
-            // If the mode of the next leg is defined and is CAV, use it, otherwise,
-            // If we don't have a current tour mode (i.e. are not on a tour aka at home),
-            // use the mode of the next leg as the new tour mode.
-            currentTourMode = modeOfNextLeg match {
-              case Some(CAV) =>
-                Some(CAV)
-              case _ =>
-                data.currentTourMode.orElse(modeOfNextLeg)
-            },
+            // We do not stick to current tour mode
```
zneedell, line 616: I'm wondering if here we can make use of the `currentTour` method defined above to take advantage of the tour structure already in BeamPlan.scala. It might take some archaeology to figure out how exactly everything was intended to function, but it looks like the `strategies` map in a `BeamPlan` allows for modes to be stored both for individual trips/legs and for tours.

So, I'm wondering if something like

`val tourModeOption = _experiencedBeamPlan.getStrategy(currentTour(data), classOf[ModeChoiceStrategy])`

would get the current tour mode, and we could do something similar with `putStrategy` when the initial trip mode is chosen and we want to set the tour mode.

zneedell (changes requested):
Logic all looks good to me. I left one main suggestion for trying to make use of the `Tour` class that already extends `PlanElement`, which might give a more consistent way of keeping track of tour and trip modes.

For the record, my longer term plan/hope here is that we can eventually implement a tour mode choice that happens before trip mode choice in chooses mode. E.g. when someone comes to a leg in their plans without a tour mode set, they make an initial tour mode choice based on the skims looking at all of the trips in the tour and then set the tour mode there, then they can go about the individual trip mode choices in real time when they get to them. We could also improve the existing tour formation logic [here](https://github.com/LBNL-UCB-STI/beam/blob/cb9885100e5589077e985b7b6f01fea9b1dabe60/src/main/scala/beam/agentsim/agents/planning/BeamPlan.scala#L111-L133) to allow for nested subtours and/or reading in tour modes (potentially without trip modes) from activitysim. 

BUT that's all down the road, I think this will get us 90+% of what we want.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
zneedell: Later edit -- if we rely on the Tour class in BeamPlan this gets easier, I think

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -587,15 +614,10 @@ class PersonAgent(
         val nextCoord = nextActivity(data).get.getCoord
```
zneedell, line 613: Now that I look at it, I wonder if the definition of `modeOfNextLeg` above can be replaced with

`val modeOfNextLeg = _experiencedBeamPlan.getStrategy(nextAct, classOf[ModeChoiceStrategy])`

zneedell:
@dimaopen -- I played around with making some of the changes involving `_experiencedBeamPlan` in a new branch based off this one here: https://github.com/LBNL-UCB-STI/beam/tree/zn/do/%233442-current-trip-tour-mode.

I'd be curious to see if you think the logic there makes sense and aligns with what you were thinking

Xuan-1998:
Test!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -128,102 +128,22 @@ trait ChoosesMode {
...
-            _,
-            _,
-            _,
-            _,
-            _,
-            _
+      case data: ChoosesModeData
+          if data.personData.currentTourPersonalVehicle.isDefined && (
+            data.personData.currentTourMode.exists(mode => mode == CAR || mode == BIKE)
+            || data.personData.currentTourMode.exists(mode => mode == DRIVE_TRANSIT || mode == BIKE_TRANSIT)
```
JustinPihony: Why is a separate `exists` needed - as it is checking against the same variable? Also, this is a behavior change that seems maybe not tripmode related. Was it meant for this PR/issue?

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -799,6 +702,28 @@ trait ChoosesMode {
...
 
+  private def correctCurrentTripModeAccordingToRules(
+    personData: BasePersonData,
+    nextAct: Activity,
+    availableModes: Seq[BeamMode]
+  ): Option[BeamMode] = {
+    val replanningIsAvailable =
+      personData.numberOfReplanningAttempts < beamServices.beamConfig.beam.agentsim.agents.modalBehaviors.maximumNumberOfReplanningAttempts
+    (personData.currentTripMode, personData.currentTourMode) match {
+      case (_, Some(CAR | BIKE)) if personData.currentTourPersonalVehicle.isDefined => personData.currentTourMode
```
JustinPihony, line 707: This seems also a new behavior - as long as tour/trip aren't getting confused....I guess this raises the question of naming - should a different name be used so somebody doesn't accidentally use the wrong variable.

JustinPihony (commented):
Deferring to Zach's comments moreso. Mine are not blockers.

dimaopen:
> 
> I'd be curious to see if you think the logic there makes sense and aligns with what you were thinking

Generally it looks good. Though we have some kind of duplication here: `personData.(currentTripMode, currentTourMode)` and `_experiencedBeamPlan.getStrategy(currentTour/Trip, classOf[ModeChoiceStrategy])` needs to hold the same data (and also be updated in the right order).
I think that we need to remove one or the other. 
By the way it already has some inconsistency. I did some sf-light runs using this new approach and you can see that a person forgot their car and used ridehail here:
car_ride_hail
If we look at a similar run with the old approach we can see that the tour mode is right:
car_choosen

Let me know what you think.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
dimaopen: Tours are created basing on "Home" activity: https://github.com/LBNL-UCB-STI/beam/blob/cb9885100e5589077e985b7b6f01fea9b1dabe60/src/main/scala/beam/agentsim/agents/planning/BeamPlan.scala#L118
So I think it's generally the same (current activity == home and isFirstTripWithinTour).

src/main/scala/beam/agentsim/agents/PersonAgent.scala
dimaopen: Replayed in the comment bellow.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
dimaopen: Replayed in the comment bellow.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
dimaopen: Correct, it's related to the `tour mode`. Removed double `exists`.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
dimaopen: You are right. @zneedell What do you think about `currentTourMode` vs `currentTripMode` names?

dimaopen:
@zneedell 
I just found that the current PR also produces some inconsistent trips. You can see WALK in the middle of the trip.
walk_in_the_middle
Probably I need to fix this behaviour.

Xuan-1998:
> @zneedell
> I just found that the current PR also produces some inconsistent trips. You can see WALK in the middle of the trip.
> walk_in_the_middle
> Probably I need to fix this behaviour.

Just out of curiosity here, so based on my understanding, we are forcing agents to use car when they depart with a car on the first Home trip? Why is this the case? I think it makes sense to me to let agents use other modes in the middle but just bring their car with them home. Correct me if I understand anything wrong here..

zneedell:
@Xuan-1998 I think you're totally right that what you've described would be the ideal behavior. The logic to do it correctly is going to be a little tricky, though, and AFAIK it hasn't been implemented here yet. Basically, doing it as you describe I think would require scanning forward in an agent's plan to check whether the agent returns to the same place later in the day and then only allowing a non-car trip if we know they're going to be able to pick up their car later.

So I think that would be really cool if we could get that working, but I'd also be OK with implementing a half step here where we don't allow for that yet. But I do think that one of the potential benefits of going down the `_experiencedBeamPlan.getStrategy` route is that it might allow for this more complex logic

zneedell:
OK I made some updates to https://github.com/LBNL-UCB-STI/beam/tree/zn/do/%233442-current-trip-tour-mode that seem to help (at least in my branch I had been saving the tripMode rather than tourMode to the tourMode column of events.csv, plus adding an explicit filter of the itineraries for mode choice based on the current tour mode). The columns here are the tour mode and the rows here are the actual mode taken: 
image

(not sure where the walk trips with the car tour mode come from)

Agreed with Dima that we probably want to either go with this solution or doing it based on `ChoosesModeData` rather than trying to do both, and I'm definitely not convinced that this is the right way to do it, but I'll do a little more testing of it this week.

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test now!

dimaopen:
Test now!

src/main/scala/beam/agentsim/agents/PersonAgent.scala
dimaopen: That `isFirstTripWithinTour` method is not used anywhere and is kept for symmetry.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -128,165 +128,57 @@ trait ChoosesMode {
...
-            _,
-            _,
-            _,
-            _
-          ) =>
-        val teleportationVehicle = createSharedTeleportationVehicle(currentLocation)
+      case data: ChoosesModeData
+          if data.personData.currentTourPersonalVehicle.isDefined &&
+            (
+              data.personData.currentTourMode.exists(mode => mode == CAR || mode == BIKE) ||
```
JustinPihony, line 146: This double exists against the same variable still is there - and is complicated by an &&. Can you add parentheses to delineate the intent please

JustinPihony: It looks like it was re-added as part of https://github.com/LBNL-UCB-STI/beam/pull/3476/commits/de1684a4dd732f7ca0e8aad7598eabad785c092e

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -157,13 +158,15 @@ class HouseholdFleetManager(
         }
       }.getOrElse {
         availableVehicles = availableVehicles match {
-          case firstVehicle :: rest =>
+          //in case of replanning because of TRANSIT failure WALK_TRANSIT is used
+          //but we may want to introduce maxWalkingDistance and check that the agent is close enough to the vehicle
+          case firstVehicle :: rest if atHome(originActivity) =>
```
dimaopen, line 163: Providing a private vehicle only when the person is at home might be a wrong idea. It may lead to increased number of "Long waling trips".

PR #3480, opened 2022-03-07
quick fix

haitamlaarabi:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3480)

PR #3484, opened 2022-03-09
Remove shared teleportation vehicles for non-teleportation trips

zneedell:
So that they don't end up throwing a none.get error when we ask for their manager here: https://github.com/LBNL-UCB-STI/beam/blob/012983f5d68d1c5a2ebbf2bf7992b48b7bc78bcd/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L1519

Stale vehicles staying in `data.availablePersonalStreetVehicles` might have something to do with the `Current tour vehicle is the same as the one being removed` error we've been seeing for a long time.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3484)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/PersonAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/PersonAgent.scala
@@ -643,7 +643,11 @@ class PersonAgent(
       )
       eventsManager.processEvent(teleportationEvent)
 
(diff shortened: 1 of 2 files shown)


zneedell:
Test!

JustinPihony:
This is superseded by #3480 ?

JustinPihony (approved):

PR #3496, opened 2022-03-27
Issue-3495 add a configurable seed for population sampling

### LINKED ISSUE #3495: Add a configurable seed for population sampling
We need to add a configurable [seed](https://en.wikipedia.org/wiki/Random_seed) for population down sampling. 
It is needed for reproducing the same household population. It should be a separate parameter.
So there are 2 cases:
- if the parameter is set - the same seed is used and the input household is the same in every run.
- if the parameter is not set - this seed should be randomized and logged in `beamLog.out`.

g-derzhavets:
https://github.com/LBNL-UCB-STI/beam/issues/3495
> 
> We need to add a configurable [seed](https://en.wikipedia.org/wiki/Random_seed) for population down sampling.
> It is needed for reproducing the same agents sampling. It should be a separate parameter.
> So there are 2 cases:
> 
> if the parameter is set - the same seed is used and the input agents are the same in every run.
> if the parameter is not set - this seed should be randomized and logged in `beamLog.out`.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3496)

### DIFF
--- a/src/main/scala/beam/sim/config/BeamConfig.scala
+++ b/src/main/scala/beam/sim/config/BeamConfig.scala
@@ -2162,8 +2162,9 @@ object BeamConfig {
 
       def apply(c: com.typesafe.config.Config): BeamConfig.Beam.Agentsim = {
         BeamConfig.Beam.Agentsim(
-          randomSeed = if (c.hasPathOrNull("randomSeed"))
-            c.getInt("randomSeed")
+          randomSeed =
+            if (c.hasPathOrNull("randomSeed"))
+              c.getInt("randomSeed")
             else new Random().nextInt(),
           agentSampleSizeAsFractionOfPopulation =
             if (c.hasPathOrNull("agentSampleSizeAsFractionOfPopulation"))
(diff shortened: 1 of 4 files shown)


g-derzhavets:
Test!

g-derzhavets:
Test!

zneedell (approved):
LGTM. I guess there's like a 1e14 chance that the same 25 agents are sampled with different seeds but I think we can live with that

src/main/scala/beam/sim/config/BeamConfig.scala
```diff
@@ -37,6 +39,7 @@ object BeamConfig {
   object Beam {
 
     case class Agentsim(
+      randomSeed: scala.Int,
       agentSampleSizeAsFractionOfPopulation: scala.Double,
```
JustinPihony, line 43: @zneedell randomSeed was never checked in for beam-template.conf. This will cause BeamConfig to be overwritten in future PRs

JustinPihony: For posterity - this was resolved in https://github.com/LBNL-UCB-STI/beam/pull/3520

PR #3497, opened 2022-03-28
Using ModeChoiceStrategy to keep current tour/trip mode

### LINKED ISSUE #3441: Walk issue in ChoosesMode

dimaopen:
Save current tour mode to _experiencedBeamPlan. This tour mode is used to calculate valid trip modes. In the future we could do tour mode choice before the first trip mode choice happens.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3497)

### DIFF
--- a/src/main/scala/beam/router/skim/core/FreightSkimmer.scala
+++ b/src/main/scala/beam/router/skim/core/FreightSkimmer.scala
@@ -24,7 +24,7 @@ class FreightSkimmer @Inject() (
   override protected val skimFileHeader =
     "tazId,hour,numberOfLoadings,numberOfUnloadings,costPerMile,walkAccessDistanceInM,parkingCostPerHour,observations,iterations"
   override protected val skimName: String = FreightSkimmer.name
-  override protected val skimType: Skims.SkimType.Value = Skims.SkimType.TC_SKIMMER
+  override protected val skimType: Skims.SkimType.Value = Skims.SkimType.FREIGHT_SKIMMER
 
   override protected def fromCsv(
     line: collection.Map[String, String]
--- a/src/main/scala/beam/router/skim/core/ParkingSkimmer.scala
+++ b/src/main/scala/beam/router/skim/core/ParkingSkimmer.scala
@@ -28,7 +28,7 @@ class ParkingSkimmer @Inject() (
   override protected val skimFileHeader =
     "tazId,hour,chargerType,walkAccessDistanceInM,parkingCostPerHour,observations,iterations"
   override protected val skimName: String = ParkingSkimmer.name
-  override protected val skimType: Skims.SkimType.Value = Skims.SkimType.TC_SKIMMER
+  override protected val skimType: Skims.SkimType.Value = Skims.SkimType.PARKING_SKIMMER
 
   override protected def fromCsv(
     line: collection.Map[String, String]
--- a/src/main/scala/beam/router/skim/core/RideHailSkimmer.scala
+++ b/src/main/scala/beam/router/skim/core/RideHailSkimmer.scala
@@ -26,7 +26,7 @@ class RideHailSkimmer @Inject() (
   override protected val skimFileHeader =
     "tazId,hour,reservationType,waitTime,costPerMile,unmatchedRequestsPercent,observations,iterations"
   override protected val skimName: String = RideHailSkimmer.name
-  override protected val skimType: Skims.SkimType.Value = Skims.SkimType.TC_SKIMMER
+  override protected val skimType: Skims.SkimType.Value = Skims.SkimType.RH_SKIMMER
 
   override protected def fromCsv(
     line: collection.Map[String, String]


Xuan-1998:
I'm seeing a lot of this error on a sample run after merging this:
image

dimaopen:
Test now!

dimaopen:
Test now!

Xuan-1998:
Doing some tests on this PR.
After merging this PR into develop, I'm seeing NAN, BIKE and DRIVE_TRANSIT as currentTourMode for WALK mode, which might not be caused by the changes here but should be something we want to solve
image
See the branch and config detail here:
https://github.com/LBNL-UCB-STI/beam/blob/Xuan/sfbay-scenarios-based-on%233497/test/input/sfbay/gemini/gemini-base-2035-activitysim-da-baseline.conf

dimaopen:
Test!

dimaopen:
@Xuan-1998 It happens because we save current trip mode to current tour mode. https://github.com/LBNL-UCB-STI/beam/blob/853ee4be4de175646cb526201fe5282acd2dc99e/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L1433
I'll change it to save current tour mode.

Xuan-1998:
> @Xuan-1998 It happens because we save current trip mode to current tour mode.
> 
> https://github.com/LBNL-UCB-STI/beam/blob/853ee4be4de175646cb526201fe5282acd2dc99e/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L1433
> 
> 
> I'll change it to save current tour mode.

Thanks Dima, and what's the difference between current trip mode and current tour mode?

Xuan-1998:
And the `mode` columns is `chosenTripMode` ?

dimaopen:
> Thanks Dima, and what's the difference between current trip mode and current tour mode?

Tour consists of several trip, ie Home -> Work -> Shopping -> Home. Tour is traveling from Home to Home. It consists of 3 Trips (Home -> Work, Work -> Shopping, Shopping -> Home). The Trip mode is the travel mode (BeamMode) of an individual trip. Tour mode is generally the first trip mode.

dimaopen:
> And the `mode` columns is `chosenTripMode` ?

Yes

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -475,6 +477,30 @@ class PersonAgent(
     }
   }
 
+  def isFirstTripWithinTour(personData: BasePersonData, nextAct: Activity): Boolean = {
```
JustinPihony: These methods do not seem to be used - Can we remove them?

src/main/scala/beam/agentsim/agents/planning/Strategy.scala
```diff
@@ -1,14 +1,67 @@
...
+import beam.router.Modes.BeamMode._
+import org.matsim.api.core.v01.population.Activity
 
 /**
   * BEAM
   */
 object Strategy {
 
-  sealed trait Strategy
+  trait Strategy {
```
JustinPihony: Any reason to unseal the trait?

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -8,6 +8,7 @@ import beam.agentsim.agents._
 import beam.agentsim.agents.household.HouseholdActor.{MobilityStatusInquiry, MobilityStatusResponse, ReleaseVehicle}
 import beam.agentsim.agents.modalbehaviors.ChoosesMode._
 import beam.agentsim.agents.modalbehaviors.DrivesVehicle.{ActualVehicle, Token, VehicleOrToken}
+import beam.agentsim.agents.planning.Strategy.ModeChoiceStrategy
```
JustinPihony: Which change takes precedence, this one or https://github.com/LBNL-UCB-STI/beam/pull/3476?

JustinPihony (commented):
Clarifying comments - especially about ChoosesMode. Otherwise it should be good to go.

Xuan-1998:
image
@dimaopen I'm seeing some agents start with their day having the alternatives CAR:WALK. But they choose WALK, and then for the rest of the day, they only choose WALK

Two questions here:
1. Can we make other modes like transit, ridehail available at least in the alternatives for people to choose?
2. If they are not walking far away at the first trip, can we make it possible for the agent to walk back to pick up the car after the first trips?

Xuan-1998:
https://github.com/LBNL-UCB-STI/beam/blob/140ee6aadb7646c35f4c57f8478aa87effa496a8/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L1208-L1211
I played around with some changes above, basically I changed all the CAR:WALK to be CAR.
But I got following:
image
which means the mode the agent is using is not choosing from the `availableAlts`

dimaopen:
>     Can we make other modes like transit, ridehail available at least in the alternatives for people to choose?
This should be done by modifying the scenario: ie increasing number of RH vehicles, transit routes. For small number of transit routes we could try changing our R5Wrapper so that it get more transit routes from the R5. But it is a research task.

>    If they are not walking far away at the first trip, can we make it possible for the agent to walk back to pick up the car after the first trips?
Yes we can do this by always providing the personal Car (not only when the agent is at home). In this case if the agent decides that it's better to go home and get the car (and not to do a long walking trip) then them do this.

src/main/scala/beam/agentsim/agents/planning/BeamPlan.scala
```diff
@@ -161,8 +165,17 @@ class BeamPlan extends Plan {
     }
   }
 
-  def getStrategy(planElement: PlanElement, forClass: Class[_ <: Strategy]): Option[Strategy] = {
-    strategies.getOrElse(planElement, mutable.Map()).get(forClass)
+  def getStrategy[T <: Strategy: ClassTag](planElement: PlanElement): Option[T] = {
```
zneedell: Is there any case when we'd expect getStrategy to return `None`? Since we're by default returning an empty Strategy I'm wondering if we could have this return `T` rather than `Option[T]`

dimaopen: If we have something like an empty `Strategy` then we could have this return `T`.

Xuan-1998:
Test!

PR #3501, opened 2022-04-01
allow non-dominating transit paths

zneedell:
A hypothesis for why we under-predicted subway ridership in New York and San Francisco is that R5 was using a different set of criteria to determine the final set of "optimal" transit paths it calculates than people do in the real world, and fixing this was made harder because we didn't have great control of the method that R5 evaluates whether one possible transit path is better than another one. In particular, R5 might have been able to find bus routes that were faster than a subway routes based on the GTFS feeds, but in reality they are less appealing (because they involve more transfers, are less reliable, are less comfortable, etc.). We've gotten around this by using a second router with no bus routes to ensure that at least one rail-only route is returned, but it would be ideal to get even more control of this.

Another way of doing this is by allowing R5 to keep track of and return more possible transit routes, and then letting BEAM decide between them based on the agent's utility functions. Currently, R5 only keeps routes that are on the cost / travel time pareto frontier (i.e. keeps a route only if there is no other route that is both cheaper and faster). R5 also includes a "suboptimal" way of collecting routes that keeps all possible routes that are a configurable amout of time slower than the fastest observed route, with some heuristics added to filter out routes that are too similar: https://github.com/LBNL-UCB-STI/r5/blob/master/src/main/java/com/conveyal/r5/profile/SuboptimalDominatingList.java.

Some tests on a 100k New York run show that swapping in this list does increase subway ridership.

Base run with just one router:
image

Two routers:
image

One router and 5 minutes nonoptimality:
image

One router and 10 minutes nonoptimality:
image

Small differences, but it does appear to increase subway ridership (total transit ridership remains roughly constant).

(unit tests to come)

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3501)

### DIFF
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -893,6 +893,8 @@ beam.routing {
     bikeLaneScaleFactor = "double | 1.0"
     bikeLaneLinkIdsFilePath = ""
     linkRadiusMeters = "double | 10000.0"
+    transitAlternativeList = "OPTIMAL"
+    suboptimalMinutes = "int | 0"
   }
   gh {
     useAlternativeRoutes = "boolean | false"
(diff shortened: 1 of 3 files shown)


Xuan-1998 (approved):
LGTM

zneedell:
Test!

PR #3508, opened 2022-04-04
get tripId from trip.leg rather than activity in _experiencedBeamPlan

zneedell:
Trip_id column hadn't been showing up in PILATES runs--I think it's because by default they are being stored in trips rather than activities in the beamPlan. This fix appears to work.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3508)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -1487,9 +1487,10 @@ trait ChoosesMode {
         )
     }
 
-    val tripId = Option(
-      _experiencedBeamPlan.activities(data.personData.currentActivityIndex).getAttributes.getAttribute("trip_id")
-    ).getOrElse("").toString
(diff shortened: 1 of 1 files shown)


Xuan-1998 (approved):
lgtm

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1487,9 +1487,10 @@ trait ChoosesMode {
         )
     }
 
-    val tripId = Option(
-      _experiencedBeamPlan.activities(data.personData.currentActivityIndex).getAttributes.getAttribute("trip_id")
-    ).getOrElse("").toString
+    val tripId = _experiencedBeamPlan.trips.lift(data.personData.currentActivityIndex + 1) match {
```
g-derzhavets, line 1490: Just curious: why do we need to increment the `currentActivityIndex`? @zneedell

Issue #3488, opened 2022-03-10
Switch from WALK to TELEPORTATION as last-resort mode


zneedell:
In case people can't otherwise end up with a legitimate mode choice in `choosesMode` we fall back on `WALK` as the last resort mode, e.g. 

https://github.com/LBNL-UCB-STI/beam/blob/ef651315eb0572a8bed0af3f84c3590e94f697ca/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L346-L347

and

https://github.com/LBNL-UCB-STI/beam/blob/ef651315eb0572a8bed0af3f84c3590e94f697ca/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L1354-L1373

This can lead to problems because in processing the outputs it can be difficult to distinguish between these last resort `WALK` trips and real `WALK` trips that were actually chosen by the agent. They can also mess up the rest of the agent's day because these walk trips often take many hours. 

These last resort walk trips should be rare and are always a sign that something unrealistic happened in the simulation, so let's consider defining a new `EMERGENCY_TELEPORT` (or `LAST_RESORT_MODE` or something like that) mode that can be used in these last resort cases that just moves them to their planned destination and adds a large replanning penalty to the current leg, but lets them continue their day. Then it will be easy in post-processing to identify the number of these trips that happen, their associated distance, and hopefully to fix the underlying problem. 

No preference as to whether we base this new mode off of the existing teleportation capability or just make it a faster walk trip that is labeled differently--whatever is easier.

Xuan-1998:
@zneedell  I'm doing some tests around here to check whether we are seeing any emergency cases at all. 
My changes as following in `choosesMode` and corresponding changes in other files regarding modes:

https://github.com/LBNL-UCB-STI/beam/blob/3532259e24a17da4a870e76d64f9234f880a4ee7/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L741-L742
https://github.com/LBNL-UCB-STI/beam/blob/3532259e24a17da4a870e76d64f9234f880a4ee7/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L1275-L1311

[...]

PR #3512, opened 2022-04-05
switch from walk to EMERGENCY(faster walk) as last resort mode #3488

Xuan-1998:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3512)


src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -472,9 +472,15 @@ trait ChoosesMode {
...
           makeRequestWith(withTransit = true, Vector(bodyStreetVehicle))
+        case Some(EMERGENCY) =>
+          responsePlaceholders = makeResponsePlaceholders(withRouting = true)
+          makeRequestWith(withTransit = true, Vector(bodyStreetVehicle))
         case Some(WALK_TRANSIT) =>
           responsePlaceholders = makeResponsePlaceholders(withRouting = true)
           makeRequestWith(withTransit = true, Vector(bodyStreetVehicle))
+        case Some(EMERGENCY_TRANSIT) =>
+          responsePlaceholders = makeResponsePlaceholders(withRouting = true)
+          makeRequestWith(withTransit = true, Vector(bodyStreetVehicle))
```
dimaopen, line 474: These 4 cases contains the same code. Could we use a single case `Some(WALK | WALK_TRANSIT | EMERGENCY .. )` here ?

src/main/scala/beam/router/Modes.scala
```diff
@@ -101,6 +103,13 @@ object Modes {
           TransportMode.transit_walk
         )
 
+    case object EMERGENCY_TRANSIT
+        extends BeamMode(
+          value = "emergency_transit",
+          Some(Right(TransitModes.TRANSIT)),
+          TransportMode.other
+        )
+
```
dimaopen, line 112: Do we really need this `EMERGENCY_TRANSIT` mode? Why we cannot use just EMERGENCY even if there is a TRANSIT leg in it?

src/main/scala/beam/router/model/EmbodiedBeamTrip.scala
```diff
@@ -115,7 +118,10 @@ object EmbodiedBeamTrip {
         theMode = leg.beamLeg.mode
       } else if (theMode == WALK && leg.beamLeg.mode == BIKE) {
         theMode = leg.beamLeg.mode
+      } else if (theMode == WALK && leg.beamLeg.mode == EMERGENCY) {
+        theMode = EMERGENCY
```
dimaopen, line 122: 1. Because we introduce a new possible return value for `EmbodiedBeamTrip#tripClassifier` then we need to carefully check all the pattern matching on this field to validate that there wouldn't be a `MatchError`.
2. Also because EMERGENCY trip is actually a WALK trip. Then we need to check all the places where `tripClassifer` is compared with WALK (i.e. https://github.com/LBNL-UCB-STI/beam/blob/676c68782f48780c0f00ad797b6fa9c56bf6f018/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L264) and add these EMERGENCY there.

src/main/scala/beam/router/model/EmbodiedBeamTrip.scala
```diff
@@ -126,6 +132,8 @@ object EmbodiedBeamTrip {
       DRIVE_TRANSIT
     } else if (theMode == TRANSIT && hasUsedBike) {
       BIKE_TRANSIT
+    } else if (theMode == TRANSIT && hasUsedEMERGENCY) {
+      EMERGENCY_TRANSIT
```
dimaopen, line 136: I think that it's safier not to change these method at all. But in this case we wouldn't have `EMERGENCY` in `PathTraversalEvent`.

dimaopen (changes requested):
Is this a real pain to merge these changes only after #3497 is merged? Or it would be a hard time to validate that the Emergency mode works as expected.

src/main/scala/beam/router/model/EmbodiedBeamTrip.scala
```diff
@@ -91,6 +93,7 @@ object EmbodiedBeamTrip {
     var hasUsedCar: Boolean = false
     var hasUsedBike: Boolean = false
     var hasUsedRideHail: Boolean = false
+    var hasUsedEMERGENCY: Boolean = false
```
dimaopen: I think it should be in CamelCase.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1371,12 +1377,12 @@ trait ChoosesMode {
                       .legs
                       .head
                 }
-              val expensiveWalkTrip = EmbodiedBeamTrip(
-                Vector(originalWalkTripLeg.copy(replanningPenalty = 10.0))
+              val EmergencyTrip = EmbodiedBeamTrip(
```
dimaopen: It needs to start from a lowercase letter.

dimaopen:
The actual emergency trip is a teleportation trip (as this task suggests) or a Walk trip?

Xuan-1998:
> The actual emergency trip is a teleportation trip (as this task suggests) or a Walk trip?



> The actual emergency trip is a teleportation trip (as this task suggests) or a Walk trip?

I was intending to change it to be a teleportation initially, but now I guess a faster Walk trip will be easier to validate

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -261,7 +242,22 @@ object PersonAgent {
     bodyVehicleId: Id[BeamVehicle],
     bodyVehicleTypeId: Id[BeamVehicleType]
   ): EmbodiedBeamTrip = {
-    if (trip.tripClassifier != WALK && trip.tripClassifier != WALK_TRANSIT) {
+    if (trip.tripClassifier != EMERGENCY && trip.tripClassifier != EMERGENCY) {
```
JustinPihony, line 245: This seems like a copy/paste error or something to that extent - as the condition is the same

JustinPihony (changes requested):
On top of what was already said, this double condition seems like a bug

Xuan-1998:
WIP

PR #3513, opened 2022-04-08
snap coordinates and log errors in csv

### LINKED ISSUE #3489: Snap coordinates from input scenario to the map after reading
To avoid discrepancy between coordinates and a map replace coordinates in input plans/trips/households with the result of r5 snap function. 
A good idea might be to count, or even write out into csv file all coordinates that can not be used (e.g. - split function returning nothing or coordinates are outside of map bounding box)
For filtered out coordinates according input element (household/trip) should not be included in scenario.

There is already a place where filtration by bounding box happens, it needs to be improved:
https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala#L62
https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala#L87

heyrutvik:
closes #3489

Use `beam.agentsim.snapLocationAndRemoveInvalidInputs` parameter to enable the snapping module. It will snap location and remove invalid plans (beam, urbansim, freight) before starting simulation.

Error csv can be found at the root of output directory:
1. `snapLocationPlanErrors.csv`: scenario plan errors
2. `snapLocationHouseholdErrors.csv`: scenario household errors
3. `snapLocationFreightTourErrors.csv`: freight tour errors
4. `snapLocationFreightPayloadPlanErrors.csv`: freight payload plan errors
5. `snapLocationFreightCarrierErrors.csv`: freight carrier errors

Csv will have `id,category,error,x,y` columns.
`error` could be either `OutOfBoundingBox` and `R5SplitNull`.
`category` defines the type of an entity for given id. (ex, Person, Household, PayloadPlan etc)
`x` and `y` is original location in plan.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3513)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
+++ b/src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
@@ -360,7 +360,7 @@ class GenericFreightReader(
   }
 
   private def getDistributedTazLocation(taz: TAZ): Coord =
-    convertedLocation(TAZTreeMap.randomLocationInTAZ(taz, rnd))
+    convertedLocation(TAZTreeMap.randomLocationInTAZ(taz, rnd, Some(snapLocationHelper)))
 
   private def extractCoordOrTaz(strX: String, strY: String, strZone: String): (Option[Id[TAZ]], ClosestUTMPoint) = {
     if (isBlank(strX) || isBlank(strY)) {
--- a/src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
+++ b/src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
@@ -1,13 +1,12 @@
 package beam.agentsim.agents.freight.input
 
 import beam.agentsim.agents.freight._
-import beam.agentsim.agents.freight.input.GenericFreightReader.ClosestUTMPoint
 import beam.agentsim.agents.vehicles.{BeamVehicle, BeamVehicleType}
 import beam.agentsim.infrastructure.taz.{TAZ, TAZTreeMap}
 import beam.sim.common.GeoUtils
 import beam.sim.config.BeamConfig.Beam.Agentsim.Agents.Freight
 import beam.utils.SnapCoordinateUtils
-import beam.utils.SnapCoordinateUtils.{Category, CsvFile, Error, ErrorInfo, Result, SnapLocationHelper}
+import beam.utils.SnapCoordinateUtils._
 import beam.utils.csv.GenericCsvReader
 import beam.utils.matsim_conversion.MatsimPlanConversion.IdOps
 import com.typesafe.scalalogging.LazyLogging
@@ -28,7 +27,7 @@ class GenericFreightReader(
   rnd: Random,
   tazTree: TAZTreeMap,
   val snapLocationAndRemoveInvalidInputs: Boolean,
-  val snapLocationHelperMaybe: Option[SnapLocationHelper] = None,
+  val snapLocationHelper: SnapLocationHelper,
   val outputDirMaybe: Option[String] = None
 ) extends LazyLogging
     with FreightReader {
@@ -63,7 +62,7 @@ class GenericFreightReader(
           row.get("departureLocationZone"),
           snapLocationAndRemoveInvalidInputs
         ) match {
-          case (_, Left(_) | Right(Result.Succeed(_))) =>
+          case (_, Right(_)) =>
             Some(
               FreightTour(
                 tourId,
@@ -71,30 +70,27 @@ class GenericFreightReader(
                 maxTourDurationInSec
               )
             )
-          case (_, Right(Result.OutOfBoundingBoxError)) =>
+          case (_, Left(Error.OutOfBoundingBoxError)) =>
             errors.append(
               ErrorInfo(
                 tourId.toString,
                 Category.FreightTour,
-                Error.OutOfBoundingBox,
+                Error.OutOfBoundingBoxError,
                 departureLocationX.toDouble,
                 departureLocationY.toDouble
               )
             )
             None
-          case (_, Right(Result.R5SplitNullError)) =>
+          case (_, Left(Error.R5SplitNullError)) =>
             errors.append(
               ErrorInfo(
                 tourId.toString,
                 Category.FreightTour,
-                Error.R5SplitNull,
+                Error.R5SplitNullError,
                 departureLocationX.toDouble,
                 departureLocationY.toDouble
               )
             )
-            None
-          case _ =>
-            logger.error(f"Following freight tour row discarded for unknown reason: $row")
             None
         }
       }
@@ -137,52 +133,46 @@ class GenericFreightReader(
         val locationX = row.get("locationX")
         val locationY = row.get("locationY")
 
-        // note: placeholder to update locationZone and locationUTM later
-        val payloadPlan = PayloadPlan(
-          payloadId,
-          get("sequenceRank").toDouble.round.toInt,
-          get("tourId").createId,
-          get("payloadType").createId[PayloadType],
-          get("weightInKg").toDouble,
-          requestType,
-          activityType,
-          None,
-          new Coord(),
-          get("estimatedTimeOfArrivalInSec").toDouble.toInt,
-          get("arrivalTimeWindowInSecLower").toDouble.toInt,
-          get("arrivalTimeWindowInSecUpper").toDouble.toInt,
-          operationDurationInSec
-        )
-
         extractCoordOrTaz(locationX, locationY, row.get("locationZone"), snapLocationAndRemoveInvalidInputs) match {
-          case (locationZoneMaybe, Left(locationZoneUTM)) =>
-            Some(payloadPlan.copy(locationZone = locationZoneMaybe, locationUTM = locationZoneUTM))
-          case (locationZoneMaybe, Right(Result.Succeed(splitCoord))) =>
-            Some(payloadPlan.copy(locationZone = locationZoneMaybe, locationUTM = splitCoord))
-          case (_, Right(Result.OutOfBoundingBoxError)) =>
+          case (locationZoneMaybe, Right(coord)) =>
+            Some(
+              PayloadPlan(
+                payloadId,
+                get("sequenceRank").toDouble.round.toInt,
+                get("tourId").createId,
+                get("payloadType").createId[PayloadType],
+                get("weightInKg").toDouble,
+                requestType,
+                activityType,
+                locationZoneMaybe,
+                coord,
+                get("estimatedTimeOfArrivalInSec").toDouble.toInt,
+                get("arrivalTimeWindowInSecLower").toDouble.toInt,
+                get("arrivalTimeWindowInSecUpper").toDouble.toInt,
+                operationDurationInSec
+              )
+            )
+          case (_, Left(Error.OutOfBoundingBoxError)) =>
             errors.append(
               ErrorInfo(
                 payloadId.toString,
                 Category.FreightPayloadPlan,
-                Error.OutOfBoundingBox,
+                Error.OutOfBoundingBoxError,
                 locationX.toDouble,
                 locationY.toDouble
               )
             )
             None
-          case (_, Right(Result.R5SplitNullError)) =>
+          case (_, Left(Error.R5SplitNullError)) =>
             errors.append(
               ErrorInfo(
                 payloadId.toString,
                 Category.FreightPayloadPlan,
-                Error.R5SplitNull,
+                Error.R5SplitNullError,
                 locationX.toDouble,
                 locationY.toDouble
               )
             )
-            None
-          case _ =>
-            logger.error(f"Following freight payload plan row discarded for unknown reason: $row")
             None
         }
       }
@@ -203,8 +193,6 @@ class GenericFreightReader(
     allPlans: Map[Id[PayloadPlan], PayloadPlan],
     vehicleTypes: Map[Id[BeamVehicleType], BeamVehicleType]
   ): IndexedSeq[FreightCarrier] = {
-    val errors: ListBuffer[ErrorInfo] = ListBuffer()
-
     val existingTours: Set[Id[FreightTour]] = allTours.keySet.intersect(allPlans.map(_._2.tourId).toSet)
     val plans: Map[Id[PayloadPlan], PayloadPlan] = allPlans.filter { case (_, plan) =>
       existingTours.contains(plan.tourId)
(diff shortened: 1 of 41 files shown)


heyrutvik:
test!

heyrutvik:
test!

heyrutvik:
test!

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
```diff
@@ -207,6 +209,41 @@ object TAZTreeMap {
...
 
+  def randomLocationInTAZ(
+    taz: TAZ,
+    rand: scala.util.Random = new scala.util.Random(System.currentTimeMillis()),
+    snapLocationHelperMaybe: Option[SnapLocationHelper] = None
+  ): Coord = {
+    val tazId = taz.tazId.toString
+    snapLocationHelperMaybe match {
+      case None => createRandomLocationInTAZ(taz, rand)
+      case Some(helper) =>
```
dimaopen: I think that it's better to use method overloading instead of providing optional `snapLocationHelperMaybe` for performance reasons.

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
```diff
@@ -207,6 +209,41 @@ object TAZTreeMap {
...
+        val max = 10000
+        var counter = 0
+        var split: Coord = null
+        while (split == null && counter < max) {
+          helper.computeResult(createRandomLocationInTAZ(taz, rand)) match {
+            case Result.Succeed(splitCoord) =>
+              logger.info(s"Found valid location $splitCoord within taz $tazId in $counter attempt(s).")
```
dimaopen: It needs to go to debug level with a template message: {}. It also could be removed completely.

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
```diff
@@ -241,18 +367,26 @@ class GenericFreightReader(
   }
 
   private def getDistributedTazLocation(taz: TAZ): Coord =
-    convertedLocation(TAZTreeMap.randomLocationInTAZ(taz, rnd))
-
-  private def getLocationFromCoord(location: Coord): Option[Coord] = {
-    closestUTMPointOnMapMaybe.map(_.find(location, geoUtils)).getOrElse(Some(geoUtils.wgs2Utm(location)))
-  }
+    convertedLocation(TAZTreeMap.randomLocationInTAZ(taz, rnd, snapLocationHelperMaybe))
```
dimaopen: This call always do snapping no matter what `snapLocationAndRemoveInvalidInputs` is.

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -706,6 +699,33 @@ trait BeamHelper extends LazyLogging {
...
+        s"""
+      |After snapping locations and validating scenario:
+      |Number of households: $afterHouseholdsCount. Removed: ${beforeHouseholdsCount - afterHouseholdsCount}.
+      |Number of persons: $afterPopulationCount. Removed: ${beforePopulationCount - afterPopulationCount}.""".stripMargin
+      )
+    }
+
     // write static metrics, such as population size, vehicles fleet size, etc.
     // necessary to be called after population sampling
     BeamStaticMetricsWriter.writeSimulationParameters(
```
heyrutvik, line 731: FYI, I removed the same method call within `samplePopulation` because it was already here.

src/main/scala/beam/sim/population/PopulationScaling.scala
```diff
@@ -378,14 +378,5 @@ object PopulationScaling {
 
     val populationAdjustment = PopulationAdjustment.getPopulationAdjustment(beamServices)
     populationAdjustment.update(scenario)
-
-    // write static metrics, such as population size, vehicles fleet size, etc.
-    // necessary to be called after population sampling
-    BeamStaticMetricsWriter.writeSimulationParameters(
```
heyrutvik, line 384: See, https://github.com/LBNL-UCB-STI/beam/pull/3513/files#r849516667

src/main/scala/beam/utils/SnapCoordinateUtils.scala
```diff
@@ -0,0 +1,76 @@
...
+
+object SnapCoordinateUtils extends LazyLogging {
+
+  trait Result
+
+  object Result {
+    final case object OutOfBoundingBoxError extends Result
+    final case object R5SplitNullError extends Result
+    final case class Succeed(splitCoord: Coord) extends Result
+  }
```
dimaopen, line 41: Since the results have completely different nature (actual result or error) I think it's better to return `Either[Error, Coord]`. Error could be an Enumeratum enum. Or it could share errors from `beam.utils.SnapCoordinateUtils.Error` which can simplify handling errors.

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
```diff
@@ -270,3 +404,10 @@ class GenericFreightReader(
   }
 
 }
+
+object GenericFreightReader {
+
+  // either[tazLocation, snapLocationResult]
+  type ClosestUTMPoint = Either[Coord, Result]
```
dimaopen: This type is either an Error or a `Coord`. For the users it doesn't matter how the coord was received. So it should be Either[Error, Coord]

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
```diff
@@ -86,31 +132,66 @@ class GenericFreightReader(
...
+          get("arrivalTimeWindowInSecLower").toDouble.toInt,
+          get("arrivalTimeWindowInSecUpper").toDouble.toInt,
+          operationDurationInSec
+        )
+
+        extractCoordOrTaz(locationX, locationY, row.get("locationZone"), snapLocationAndRemoveInvalidInputs) match {
+          case (locationZoneMaybe, Left(locationZoneUTM)) =>
+            Some(payloadPlan.copy(locationZone = locationZoneMaybe, locationUTM = locationZoneUTM))
+          case (locationZoneMaybe, Right(Result.Succeed(splitCoord))) =>
+            Some(payloadPlan.copy(locationZone = locationZoneMaybe, locationUTM = splitCoord))
```
dimaopen: If we change the types as I suggests these 2 `case`s could be unified and we don't need `.copy` anymore.

src/main/scala/beam/utils/scenario/ScenarioLoaderHelper.scala
```diff
@@ -0,0 +1,174 @@
...
+              errors :+ ErrorInfo(
+                personId.toString,
+                Category.ScenarioPerson,
+                Error.R5SplitNull,
+                planCoord.getX,
+                planCoord.getY
+              )
+          }
+      }
+    }
```
dimaopen, line 53: We have a very similar code that uses `snapLocationHelper.computeResult`. We could use `HasCoord` type class to reuse this code.

dimaopen (changes requested):

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
heyrutvik: Ah, right. This was great suggestion, it simplified code and made it more type-checkable. Thanks!

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
heyrutvik: Done.

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
heyrutvik: Corrected.

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
heyrutvik: Corrected.

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
heyrutvik: Done.

src/main/scala/beam/utils/SnapCoordinateUtils.scala
heyrutvik: Corrected. Great suggestion! :rocket:

src/main/scala/beam/utils/scenario/ScenarioLoaderHelper.scala
heyrutvik: I tried to look for better way to handle this as per your suggestion but I think we can't make it re-usable. Based on different logic, we do different things in those cases. Some of them accumulate errors, some modifies scenario etc. 

If you could sketch it out your idea bit more than I might able to understand your POV. Thanks.

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
```diff
@@ -40,33 +44,71 @@ class GenericFreightReader(
...
+            None
+          case (_, Left(Error.R5SplitNullError)) =>
+            errors.append(
+              ErrorInfo(
+                tourId.toString,
+                Category.FreightTour,
+                Error.R5SplitNullError,
+                departureLocationX.toDouble,
+                departureLocationY.toDouble
+              )
```
dimaopen, line 81: In all the similar error handling code we could contract 2 Error cases into a single one
```
case (_, Left(error)) =>
            errors.append(
              ErrorInfo(
                payloadId.toString,
                Category.FreightPayloadPlan,
                error,
                locationX.toDouble,
                locationY.toDouble
              )
            )
            None
```

src/main/scala/beam/utils/scenario/ScenarioLoaderHelper.scala
dimaopen: What I'm suggesting is somewhat similar to `beam.utils.matsim_conversion.ShapeUtils#quadTree` method.
But I'm Ok if you decide not to go with it.

Generally it looks like the following:
`SnapCoordinateUtils`:
```
import beam.utils.matsim_conversion.ShapeUtils.HasCoord
    def snapCoordinates[A: HasCoord](elements: Seq[A], isWgs: Boolean = false): Seq[(A, SnapCoordinateResult)] = {
      val hasCoord = implicitly[HasCoord[A]]
      elements.map { a =>
        val coord = hasCoord.getCoord(a)
        a -> computeResult(coord, isWgs)
      }
    }
```
You also can return some a tuple from this methos which includes other useful info (i.e. if errors are presented, in order not to go through the whole collection). You could create a `par` version of the same method for big coolections.

`ScenarioLoaderHelper`:
```
val householdsWithMembers: List[Household] = scenario.getHouseholds.getHouseholds.values().asScala.toList
    snapLocationHelper.snapCoordinates(householdsWithMembers).map{
      case (household, Right(splitCoord)) =>
        val householdId = household.getId.toString
        val attr: ObjectAttributes = ???
        attr.putAttribute(householdId, "homecoordx", splitCoord.getX)
        attr.putAttribute(householdId, "homecoordy", splitCoord.getY)
      case (household, Left(error)) =>
        household.getMemberIds.asScala.toList.foreach(personId => scenario.getPopulation.getPersons.remove(personId))
        scenario.getHouseholds.getHouseholds.remove(household.getId)
        householdErrors.append(
          ErrorInfo(
            householdId,
            Category.ScenarioHousehold,
            error,
            planCoord.getX,
            planCoord.getY
          )
        )
    }
```

dimaopen (commented):
We need to unify at least the error handling.

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
heyrutvik: Ah, I missed it. Thanks. Done.

src/main/scala/beam/utils/scenario/ScenarioLoaderHelper.scala
heyrutvik: @dimaopen Thanks for detailed comment. I don't think we need this change at this point of time. Let's keep it simple for now.

heyrutvik:
test!

src/main/scala/beam/utils/scenario/ScenarioLoaderHelper.scala
```diff
@@ -0,0 +1,166 @@
...
+          household.getMemberIds.asScala.toList.foreach(personId => scenario.getPopulation.getPersons.remove(personId))
+          scenario.getHouseholds.getHouseholds.remove(household.getId)
+          householdErrors.append(
+            ErrorInfo(
+              householdId,
+              Category.ScenarioHousehold,
+              error,
+              planCoord.getX,
+              planCoord.getY
+            )
```
dimaopen, line 142: These 2 error cases could be collapsed into a single one.

heyrutvik: Do you mean something like this,

```
case Left(error) =>
  householdErrors.append(
    ErrorInfo(
      householdId,
      Category.ScenarioHousehold,
      error,
      planCoord.getX,
      planCoord.getY
    )
  )
  error match {
    case Error.OutOfBoundingBoxError =>
      household.getMemberIds.asScala.toList.foreach(personId => scenario.getPopulation.getPersons.remove(personId))
      scenario.getHouseholds.getHouseholds.remove(household.getId)
    case Error.R5SplitNullError =>
      household.getMemberIds.asScala.toList.foreach(personId => scenario.getPopulation.getPersons.remove(personId))
      scenario.getHouseholds.getHouseholds.remove(household.getId)
  }
```

dimaopen: No, just
```
case Left(error) =>
          household.getMemberIds.asScala.toList.foreach(personId => scenario.getPopulation.getPersons.remove(personId))
          scenario.getHouseholds.getHouseholds.remove(household.getId)
          householdErrors.append(
            ErrorInfo(
              householdId,
              Category.ScenarioHousehold,
              error,
              planCoord.getX,
              planCoord.getY
            )
          )
```

heyrutvik: Oh, man! What am I thinking!? I think I need coffee. I'll make the changes. Sorry for asking stupid questions on a plain simple change. :pray:

heyrutvik:
test!

dimaopen (approved):

PR #3516, opened 2022-04-13
Improve LookUpVehicle in ChargingNetworkManager

haitamlaarabi:
@dimaopen  found:
"I found that ChargingNetworkManager may cause a bottleneck in some scenarios (i.e. sf-light-25k). Some scenarios (Austin-base) are fine. This is caused by the method ChargingNetwork.ChargingStation#lookupVehicle which is called too often. Do you think it's worth optimizing that method?"
sf-light-25k

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3516)


dimaopen:
Test!

dimaopen (approved):

PR #3519, opened 2022-04-22
New Parking Algorithm

dimaopen:
The changes is the following:
1. We keep all the zones which are belong to Default manager in a map ZoneType -> ParkingZones.
2. All the reserved zones are kept in a map reservedFor -> ParkingZones
3. We keep only free zones (that have at least one free stall)
4. For the zones from 1. for each zone type we return only 50% of zones (min 10) in order to decrease time of choosing zones.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519)

### DIFF
--- a/src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala
+++ b/src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala
@@ -24,6 +24,8 @@ class ParkingFunctions[GEO: GeoLevel](
   maxSearchRadius: Double,
   searchMaxDistanceRelativeToEllipseFoci: Double,
   enrouteDuration: Double,
+  fractionOfSameTypeZones: Double,
+  minNumberOfSameTypeZones: Int,
   boundingBox: Envelope,
   seed: Int,
   mnlParkingConfig: BeamConfig.Beam.Agentsim.Agents.Parking.MulitnomialLogit
@@ -37,6 +39,8 @@ class ParkingFunctions[GEO: GeoLevel](
       maxSearchRadius,
       searchMaxDistanceRelativeToEllipseFoci,
       enrouteDuration,
+      fractionOfSameTypeZones,
+      minNumberOfSameTypeZones,
       boundingBox,
       seed
     ) {
@@ -218,8 +222,6 @@ class ParkingFunctions[GEO: GeoLevel](
     inquiry: ParkingInquiry,
     preferredParkingTypes: Set[ParkingType]
   ): Boolean = {
-    val hasAvailability: Boolean = parkingZones(zone.parkingZoneId).stallsAvailable > 0
-
     val validParkingType: Boolean = preferredParkingTypes.contains(zone.parkingType)
 
     val isValidTime = inquiry.beamVehicle.forall(vehicle =>
@@ -228,11 +230,7 @@ class ParkingFunctions[GEO: GeoLevel](
         .forall(_.contains(inquiry.destinationUtm.time % (24 * 3600)))
     )
 
-    val isValidVehicleManager = inquiry.beamVehicle.forall { vehicle =>
-      zone.reservedFor.managerType == VehicleManager.TypeEnum.Default || zone.reservedFor.managerId == inquiry.reservedFor.managerId
-    }
-
-    hasAvailability & validParkingType & isValidTime & isValidVehicleManager
+    validParkingType && isValidTime
   }
 
   /**
--- a/src/main/scala/beam/utils/MathUtils.scala
+++ b/src/main/scala/beam/utils/MathUtils.scala
@@ -162,6 +162,40 @@ object MathUtils {
   def nanToZero(x: Double) = if (x.isNaN) { 0.0 }
   else { x }
 
+  /**
+    * It selects random elements out of a collection.
+    * It is designed to be performant on not indexed collections (like Sets).
+    * @param xs the collection
+    * @param n number of elements to select
+    * @param random a Random
+    * @tparam T type of elements
+    * @return an array of selected elements
+    */
+  def selectRandomElements[T: ClassTag](xs: Iterable[T], n: Int, random: Random): Array[T] = {
+    val size = xs.size
+    val numToTake = Math.min(n, size)
+    val result = Array.ofDim[T](numToTake)
+    val it = xs.iterator
+    var originalElementsLeft = size
+    var i = 0
+    while (i < numToTake) {
+      val elem = it.next()
+      val resultElementsLeft = numToTake - i
+      if (resultElementsLeft < originalElementsLeft) {
+        val probability = resultElementsLeft.toDouble / originalElementsLeft
+        if (random.nextDouble() < probability) {
+          result(i) = elem
+          i += 1
+        }
+      } else {
+        result(i) = elem
+        i += 1
+      }
+      originalElementsLeft -= 1
+    }
+    result
+  }
+
   def selectElementsByProbability[T](
     rndSeed: Long,
     elementToProbability: T => Double,
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -168,6 +168,8 @@ beam.agentsim.agents.parking.rangeAnxietyBuffer = "double | 20000.0" # if our re
 beam.agentsim.agents.parking.minSearchRadius = "double | 250.00"  # something small enough that, on average, we see 2 orders of magnitude of samples at most
 beam.agentsim.agents.parking.maxSearchRadius = "double | 8046.72" # something large enough that we do not strand our drivers in an oasis
 beam.agentsim.agents.parking.searchMaxDistanceRelativeToEllipseFoci = "double | 4.0" # something large enough that we do not travel longer than the actual distance to destination
+beam.agentsim.agents.parking.fractionOfSameTypeZones = "double | 0.5"
+beam.agentsim.agents.parking.minNumberOfSameTypeZones = "int | 10"
 
 #TAZ params
 beam.agentsim.taz.filePath = ${beam.inputDirectory}"/taz-centers.csv"
--- a/src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
+++ b/src/main/scala/beam/agentsim/agents/household/HouseholdActor.scala
@@ -48,6 +48,7 @@ import org.matsim.households
 import org.matsim.households.Household
 
 import java.util.concurrent.TimeUnit
+import java.util.concurrent.atomic.AtomicReference
 import scala.collection.mutable
 import scala.concurrent.{ExecutionContext, Future}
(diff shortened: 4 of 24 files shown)


haitamlaarabi (approved):
> Reviewed 22 of 24 files at r1, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519)* status: 22 of 24 files reviewed, all discussions resolved (waiting on @JustinPihony)

haitamlaarabi:
@dimaopen what's the performance improvement ?

dimaopen:
Test now!

dimaopen (commented):
@haitamlaarabi 3-30 times better depending on the parking zones distribution and the value of `fractionOfSameTypeZones`. Please look at https://docs.google.com/spreadsheets/d/1GOCrbcp7tclMBm-WEBheDawYu0TsdxOomTHdmfXYBIU/edit?usp=sharing

> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-:-N0ePNjZAElAiGAftB-f:benwy8n)* status: 22 of 24 files reviewed, all discussions resolved (waiting on @JustinPihony)

src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala
```diff
@@ -228,11 +230,7 @@ class ParkingFunctions[GEO: GeoLevel](
         .forall(_.contains(inquiry.destinationUtm.time % (24 * 3600)))
     )
 
-    val isValidVehicleManager = inquiry.beamVehicle.forall { vehicle =>
-      zone.reservedFor.managerType == VehicleManager.TypeEnum.Default || zone.reservedFor.managerId == inquiry.reservedFor.managerId
-    }
-
-    hasAvailability & validParkingType & isValidTime & isValidVehicleManager
+    validParkingType && isValidTime
```
JustinPihony, line 233: Why were these other factors removed? Maybe they help performance, but they are also valid criteria.

src/main/scala/beam/utils/MathUtils.scala
```diff
@@ -162,6 +162,40 @@ object MathUtils {
...
+    * It is designed to be performant on not indexed collections (like Sets).
+    * @param xs the collection
+    * @param n number of elements to select
+    * @param random a Random
+    * @tparam T type of elements
+    * @return an array of selected elements
+    */
+  def selectRandomElements[T: ClassTag](xs: Iterable[T], n: Int, random: Random): Array[T] = {
+    val size = xs.size
+    val numToTake = Math.min(n, size)
```
JustinPihony, line 176: There is an optimization here, but it depends on the typical sizes versus selection. If size is the number to take then you can skip the entire loop and return the set as is. But if that is not often then it wouldn't be worth the extra check.

src/main/scala/beam/utils/MathUtils.scala
```diff
@@ -162,6 +162,40 @@ object MathUtils {
...
+    val result = Array.ofDim[T](numToTake)
+    val it = xs.iterator
+    var originalElementsLeft = size
+    var i = 0
+    while (i < numToTake) {
+      val elem = it.next()
+      val resultElementsLeft = numToTake - i
```
JustinPihony, line 183: This should be a var set on the outside of the while and updated as appropriate. There is no need to keep creating a value that is possibly static for long-ish periods of time.

src/main/scala/beam/utils/MathUtils.scala
```diff
@@ -162,6 +162,40 @@ object MathUtils {
...
+      if (resultElementsLeft < originalElementsLeft) {
+        val probability = resultElementsLeft.toDouble / originalElementsLeft
+        if (random.nextDouble() < probability) {
+          result(i) = elem
+          i += 1
+        }
+      } else {
+        result(i) = elem
```
JustinPihony, line 191: At this point it would be more efficient to take the rest of the sequence at this point - no need to iterate and perform any more checks.

src/main/scala/beam/utils/MathUtils.scala
```diff
@@ -162,6 +162,40 @@ object MathUtils {
...
+    */
+  def selectRandomElements[T: ClassTag](xs: Iterable[T], n: Int, random: Random): Array[T] = {
+    val size = xs.size
+    val numToTake = Math.min(n, size)
+    val result = Array.ofDim[T](numToTake)
+    val it = xs.iterator
+    var originalElementsLeft = size
+    var i = 0
+    while (i < numToTake) {
+      val elem = it.next()
```
JustinPihony, line 182: There is no need to extract the item until it is needed. You could `drop` when it comes time to get the actual element.

src/main/scala/beam/utils/MathUtils.scala
```diff
@@ -162,6 +162,40 @@ object MathUtils {
...
+  /**
+    * It selects random elements out of a collection.
+    * It is designed to be performant on not indexed collections (like Sets).
+    * @param xs the collection
+    * @param n number of elements to select
+    * @param random a Random
+    * @tparam T type of elements
+    * @return an array of selected elements
+    */
+  def selectRandomElements[T: ClassTag](xs: Iterable[T], n: Int, random: Random): Array[T] = {
```
JustinPihony, line 174: I made comments on areas of this particular implementation, however I would be interested in why not something akin to:

```
val indexedIterable = xs.toVector
val toSelect = scala.collection.mutable.Set[Int]() 
val maxSize = xs.size - 1
while(toSelect.size < n){
  toSelect += random.nextInt(maxSize)
}
toSelect.map(indexedIterable)
```

Although I would leave it as `scala.util.Random` instead of taking in a `random`, as it is already fairly random that way.

Also, you can optimize the edge cases here, but I think it is more efficient in most cases - you pay a one time index and then move on.

JustinPihony (commented):
I know this is a bit late. I wanted to point out that there is a condition omission that seems necessary still, and that the random algorithm can be simplified and improved at the same time.

dimaopen (commented):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519)* status: 22 of 24 files reviewed, 6 unresolved discussions

___
*[`src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala` line 233 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-N0H0Yo_X4o5DLa-NL6r-r1-233:-N1D9xjV7FitDbArc5zG:bwnvgc1) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/b588d57ce951605166fc4830611346d9bb0ff5f1/src/main/scala/beam/agentsim/infrastructure/ParkingFunctions.scala#L233)):*

Why were these other factors removed? Maybe they help performance, but they are also valid criteria.
</blockquote></details>

In the new parking search only parking zones with stallsAvailable > 0 is provided. The same for the second condition (valid vehicle manager).
___
*[`src/main/scala/beam/utils/MathUtils.scala` line 174 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-N0H0Yo_X4o5DLa-NL6y-r1-174:-N1DBgKU3daz1kt-QZpS:b-mqgz8b) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/b588d57ce951605166fc4830611346d9bb0ff5f1/src/main/scala/beam/utils/MathUtils.scala#L174)):*

I made comments on areas of this particular implementation, however I would be interested in why not something akin to:

```
val indexedIterable = xs.toVector
val toSelect = scala.collection.mutable.Set[Int]() 
val maxSize = xs.size - 1
while(toSelect.size < n){
  toSelect += random.nextInt(maxSize)
}
toSelect.map(indexedIterable)
```

Although I would leave it as `scala.util.Random` instead of taking in a `random`, as it is already fairly random that way.

Also, you can optimize the edge cases here, but I think it is more efficient in most cases - you pay a one time index and then move on.
</blockquote></details>

It looks good, except one thing: the condition in `while` might not be true for a lot of iterations especially if `n` is close to `xs.size`.
Generally I was trying to avoid creating a new collection to decrease number of Garbage Collections. We could use `scala.util.Random#shuffle` for getting random elements as a standard approach.
___
*[`src/main/scala/beam/utils/MathUtils.scala` line 176 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-N0H0Yo_X4o5DLa-NL6y-r1-176:-N1DEpDk0RigUymgXWOL:b-iu45bs) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/b588d57ce951605166fc4830611346d9bb0ff5f1/src/main/scala/beam/utils/MathUtils.scala#L176)):*

There is an optimization here, but it depends on the typical sizes versus selection. If size is the number to take then you can skip the entire loop and return the set as is. But if that is not often then it wouldn't be worth the extra check.
</blockquote></details>

I agree. But in this case we cannot have `Array` as a return type. And we also catch this in the `if (resultElementsLeft < originalElementsLeft) {` block.
___
*[`src/main/scala/beam/utils/MathUtils.scala` line 182 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-N0H0Yo_X4o5DLa-NL6y-r1-182:-N1DFTTL-N1uCtO_8sIC:b6lm23) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/b588d57ce951605166fc4830611346d9bb0ff5f1/src/main/scala/beam/utils/MathUtils.scala#L182)):*

There is no need to extract the item until it is needed. You could `drop` when it comes time to get the actual element.
</blockquote></details>

But it makes the code more complex: We have to either drop or extract. And drop is essentially next called n times.
___
*[`src/main/scala/beam/utils/MathUtils.scala` line 183 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-N0H0Yo_X4o5DLa-NL6y-r1-183:-N1DLUbL1tLcsWzNmQRY:b-z81rgj) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/b588d57ce951605166fc4830611346d9bb0ff5f1/src/main/scala/beam/utils/MathUtils.scala#L183)):*

This should be a var set on the outside of the while and updated as appropriate. There is no need to keep creating a value that is possibly static for long-ish periods of time. 
</blockquote></details>

Implemented
___
*[`src/main/scala/beam/utils/MathUtils.scala` line 191 at r1](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-N0H0Yo_X4o5DLa-NL6y-r1-191:-N1DNn5K0V7TAYwLJTDm:b-z81rgj) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/b588d57ce951605166fc4830611346d9bb0ff5f1/src/main/scala/beam/utils/MathUtils.scala#L191)):*

At this point it would be more efficient to take the rest of the sequence at this point - no need to iterate and perform any more checks.
</blockquote></details>

Implemented

PR #3522, opened 2022-04-25
Create and store network route for a person plan if it is empty

alex-vv:
In case `PersonAgent` doesn't have a route for next activity then store the result of the router response in `_experiencedBeamPlan` which means it will be available for use in plan strategies like `ClearRoutes`. 

This will help in case routes are not created during scenario loading, for example for urbansim_v2 csv loader.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3522)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
+++ b/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
@@ -747,7 +747,7 @@ trait ChoosesMode {
       for {
         activity <- nextActivity(choosesModeData.personData)
         leg      <- _experiencedBeamPlan.getTripContaining(activity).leg if leg.getRoute == null
-      } yield {
+      } {
         val links =
           response.itineraries
             .flatMap(_.beamLegs)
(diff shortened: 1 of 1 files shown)


src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -739,6 +742,29 @@ trait ChoosesMode {
           routingFinished = true
         )
       }
+
+      // If person plan doesn't have a route for an activity create and save it
+      for {
+        activity <- nextActivity(choosesModeData.personData)
+        leg      <- _experiencedBeamPlan.getTripContaining(activity).leg if leg.getRoute == null
+      } yield {
```
Xuan-1998: Is there any specific reason for using yield here? Will it be better to only use a for loop if nothing needs to be yielded for use?

Xuan-1998 (changes requested):

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
alex-vv: right, yield it not needed here

Xuan-1998 (approved):

Xuan-1998:
Test!

dimaopen (approved):

alex-vv:
test!

PR #3523, opened 2022-05-04
Implemented peer review comments

dimaopen:
Implemented Junstin's comments from here https://reviewable.io/reviews/lbnl-ucb-sti/beam/3519#-N0H0Yo_X4o5DLa-NL6r-r1-233

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3523)


dimaopen:
If we carefully consider the suggested changes then we can see that no performance improvement can be achieved.

PR #3531, opened 2022-05-17
Network relaxation scenario & spec

### LINKED ISSUE #3407: Test: testing relaxation of network.
To properly test network relaxation we need to create a new scenario with simple map and relatively high population (~2k) to create congestion on links.

Here is a picture of a map for scenario:
image

Links from 1 to 2 and from 3 to 4 are one-directional, the rest are bi-directional.

The scenario format should be urbansim_v2. 
The scenario name should be `relaxation-network-scenario`, the place - test/input/relaxation-network-scenario
The departure window for each actor and links capacity for map is a matter of tuning.
The goal of this scenario is to test network relaxation on links between nodes 1 and 2. With proper tuning and after 5 (or 10) iterations we expect to see almost even distribution of cars between all four ways from node 1 to node 2.

In order to create this test we need:
1. create a simple OSM map generator or an instruction of how to create an osm map by hand
2. create a scenario generator to create a scenario (because manually generating 2k population will be a problem)
3. calibrate the scenario (links capacities on the map / flowCapacity in the config / departure time of agens) in order to achieve equilibrium at last iterations of simulation

alex-vv:
This PR creates a new spec for testing network relaxation, see more details in #3407 

Fixes #3407

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3531)


alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

src/main/resources/beam-template.conf
```diff
@@ -631,6 +631,8 @@ beam.physsim.network.maxSpeedInference {
   type = "String | MEAN"
 }
 
+beam.physsim.network.removeIslands = "boolean | true"
```
nikolayilyin, line 636: Let's add some explanation for this new config key, just in a rows above in comment.

src/test/scala/beam/integration/NetworkRelaxationSpec.scala
```diff
@@ -0,0 +1,226 @@
...
+import com.google.inject.Injector
+import com.typesafe.config.ConfigFactory
+import org.matsim.core.scenario.MutableScenario
+import org.scalatest.wordspec.AnyWordSpecLike
+import org.scalatest.matchers.should.Matchers._
+
+import java.io.{Closeable, File}
+import scala.util.Random
+
+class NetworkRelaxationScenarioGenerator {
```
nikolayilyin: I suggest moving this generator into scripts as separate class/object. Otherwise no one will ever be aware of its existence.

nikolayilyin (changes requested):
Few nitpicks, the rest looks great

src/main/resources/beam-template.conf
alex-vv: done

src/test/scala/beam/integration/NetworkRelaxationSpec.scala
alex-vv: done

alex-vv:
test!

nikolayilyin (approved):

nikolayilyin:
test!

PR #3532, opened 2022-05-23
More control of EC2 instances and budgeting

JustinPihony:
Note that budgets are set up at https://us-east-1.console.aws.amazon.com/billing/home?#/budgets/overview (region doesn't matter as budgets are global). 

** Update: Setup and description are at https://github.com/LBNL-UCB-STI/beam/pull/3532/files#diff-64f6cd27a99e8fce7c818fc26efc5d8d13cc0b538f667bb1df6b96a22ecdd4f1 **

The suggested setup for this budget is a Cost Budget:
* Name = **SOME APPROPRIATE NAME (??Beam AWS Monthly Budget??)** (**This name should then be matched as the environment variable shown in the yaml**)
* Period = Monthly
* Recurring budget
* Budgeting method = Fixed
* Budget amount = **CHOOSE APPROPRIATE VALUE**
* Budget scope = All AWS Services
* First Alert
  * Threshold = 60% of budgeted amount and Trigger = Actual 
  * Email Recipients = All emails who should know when this threshold has occurred

Then there is an always on Event Rule (I believe this **is** region specific, so would need to be created on all regions for full security) named `instance_creation_notifier` with the below pattern. This will trigger anytime an EC2 instance changes into the `Running` state after the budget is both 60% over **and** over the monthly percent (ie. if the budget is 75% over but the month is 80% then it will NOT alert). It should be configured to go to an SNS topic that points to the above lambda, which will submit to an SNS notifier (instance_notifier - which emails out the key data points to what should be the same list as the budget alert). It will also **stop** any instance not tagged with `BudgetOverride`=`True` once the budget is over 100%. 

```
{
  "source": ["aws.ec2"],
  "detail-type": ["EC2 Instance State-change Notification"],
  "detail": {
    "state": ["running"]
  }
}
```

The event rule also uses a transformer to limit the json to what is needed:

```
{
  "account-id": "$.account",
  "instance-id": "$.detail.instance-id",
  "region": "$.region",
  "state": "$.detail.state"
}
```

transforms into

```
{
  "instance-id": <instance-id>,
  "state": <state>, 
  "region": <region>,
  "account-id": <account-id>
}
```

This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3532)

### DIFF
--- a/aws/build.gradle
+++ b/aws/build.gradle
@@ -149,6 +149,7 @@ task deployPilates(type: AWSLambdaInvokeTask) {
         defaultParamVals['dataRegion'] = defRegion
         defaultParamVals['beamBranch'] = getCurrentGitBranch()
         defaultParamVals['dataBranch'] = 'develop'
+        defaultParamVals['budgetOverride'] = True
 
         def paramMap = [:]
         def missing = []
@@ -156,7 +157,7 @@ task deployPilates(type: AWSLambdaInvokeTask) {
         // all possible parameters
         ['runName', 'pilatesScenarioName',
          'beamBranch', 'dataBranch', 'beamCommit',
-         'beamConfig',
+         'beamConfig', 'budgetOverride'
          'startYear', 'countOfYears', 'beamItLen', 'urbansimItLen',
          'initialS3UrbansimInput', 'initialS3UrbansimOutput', 'initialSkimPath',
          'pilatesImageVersion', 'pilatesImageName',
@@ -175,6 +176,7 @@ task deployPilates(type: AWSLambdaInvokeTask) {
         }
 
         paramMap['runName'] = paramMap['runName'] + '_' + getCurrentGitUserEmail()
+        paramMap['budget_override'] = paramMap['budgetOverride']
 
         println("parameters wasn't specified: " + missing.join(', '))
 
@@ -268,6 +270,7 @@ def buildPayload(command) {
     return """{
   "instance_ids": "${project.findProperty('instanceIds')}",
   "region": "${project.findProperty('region') ?: defaultRegion}",
+  "budget_override": ${ext.getParameterValue('budgetOverride') ?: true}
   "command": "$command"
 }"""
 }
--- a/aws/build.gradle
+++ b/aws/build.gradle
@@ -110,7 +110,7 @@ task deploy(type: AWSLambdaInvokeTask) {
   "max_cores": ${ext.getParameterValue('maxCores') ?: 0},
   "min_memory": ${ext.getParameterValue('minMemory') ?: 0},
   "max_memory": ${ext.getParameterValue('maxMemory') ?: 0},
-  "budget_override": ${ext.getParameterValue('budgetOverride') ?: true}
+  "budget_override": ${ext.getParameterValue('budgetOverride') ?: false}
 }"""
         payload = pload
         println "Using data branch $dataBranch"
@@ -149,7 +149,7 @@ task deployPilates(type: AWSLambdaInvokeTask) {
         defaultParamVals['dataRegion'] = defRegion
         defaultParamVals['beamBranch'] = getCurrentGitBranch()
         defaultParamVals['dataBranch'] = 'develop'
-        defaultParamVals['budgetOverride'] = True
+        defaultParamVals['budgetOverride'] = False
 
         def paramMap = [:]
         def missing = []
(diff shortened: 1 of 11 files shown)


aws/src/main/python/instanceNotifier/instance_notifier.py
```diff
@@ -0,0 +1,198 @@
...
+    days_in_current_month = monthrange(int(current_year), int(current_month))[1]
+    percent_of_month = (current_day / Decimal(days_in_current_month)) * 100
+
+    if percent_of_budget_used <= percent_of_month:
+        return "Budget does not exceed the current point of the month. Stopping check"
+
+    # We are over 60% and over budget AND above percent of month remaining, so notifications must go out
+    # LATER TODO: Add link to budget?
+    # LATER TODO: Send to slack also?
+    # LATER TODO: Add an override option that is not emptied (true_always?) or for # of starts (true_3? - decrement and update tag)
```
JustinPihony, line 189: Should there be a way to deploy an instance with this tag already set? Or would it be better to have this extra step if over budget no matter? In either case a notification would be sent.

JustinPihony:
@wrashid This is the instance limiter code. I do not have it fully in place because part of it still needs verified (ie. monthly budget amount, who to email, and if the override should be hard required to start). I can meet to discuss before turning it on more officially (I also have a markdown of the flow for documentation on my other computer - will update later). Also, if you have anybody who might be good to ping to review it from a code perspective then feel free to assign them. 

Also, I stuck with email notification to start - it wouldn't be hard to add Slack notification on top of - or in addition.

aws/build.gradle
```diff
@@ -109,7 +109,8 @@ task deploy(type: AWSLambdaInvokeTask) {
   "min_cores": ${ext.getParameterValue('minCores') ?: 0},
   "max_cores": ${ext.getParameterValue('maxCores') ?: 0},
   "min_memory": ${ext.getParameterValue('minMemory') ?: 0},
-  "max_memory": ${ext.getParameterValue('maxMemory') ?: 0}
+  "max_memory": ${ext.getParameterValue('maxMemory') ?: 0},
+  "budget_override": ${ext.getParameterValue('budgetOverride') ?: true}
```
JustinPihony: Should the default be true?

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -673,6 +679,7 @@ def get_param(param_name):
     run_grafana = event.get('run_grafana', False)
     run_helics = event.get('run_helics', False)
     profiler_type = event.get('profiler_type', 'null')
+    budget_override = event.get('budget_override', True)
```
JustinPihony: Same, should the default be true?

aws/src/main/python/instanceNotifier/instance_notifier.py
```diff
@@ -0,0 +1,226 @@
...
+    # LATER TODO: Add a link to trigger an automatic approval. Maybe link to instance is enough, since can tag there?
+    # LATER TODO: Create a report of all usages and cost and send as a daily/weekly? email
+    init_ec2_client(region)
+    instance_details = get_instance_details_for_(instance_id)
+
+    instance_stopped = False
+    if percent_of_budget_used >= 300:
+        stop_instance(instance_id)
+        instance_stopped = True
+    elif percent_of_budget_used >= 150 and instance_details.budget_override != "true":
```
dimaopen: In the `start_instance` method of the lambda we set the tag value 'True"
```
           Tags=[
                {
                    'Key': 'BudgetOverride',
                    'Value': 'True'
                }
            ]
```
But it is compared to 'true' here. Is this OK?

dimaopen (approved):
Please check the values of 'BudgetOverride' Tag. Otherwise it looks good.

aws/build.gradle
JustinPihony: Changed to false

aws/src/main/python/beam_lambda/lambda_function.py
JustinPihony: Same - changed to false

aws/src/main/python/instanceNotifier/instance_notifier.py
JustinPihony: Changing it to a lowercase so it handles all flavors of casing

PR #3534, opened 2022-05-30
New utility notebooks added into jupyter folder

nikolayilyin:
- a jupyter notebook to analyze aws resources usage by simulations
- a jupyter notebook to get latest closed BEAM PR's

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3534)


nikolayilyin:
test!

jupyter/utility/latest runs from spreadsheet analysis.py
```diff
@@ -0,0 +1,150 @@
...
+                     'r5.xlarge':0.252, 
+                     'r5.24xlarge':6.048,
+                     'r5.8xlarge':2.016, 
+                     'm5.24xlarge':4.608, 
+                     'r5.2xlarge':0.504, 
+                     'm4.16xlarge':3.20
+                    }
+
+missing_instance_types = set()
+def get_price(row):
```
dimaopen, line 92: Maybe we could extract this and other utility method to a python package? Then we could use it here and in the jupyter notebook.

dimaopen (commented):
I think that we need to reuse the same code or it would lead to some technical errors in the future (when someone forgot update the other file and it causes different results).

jupyter/utility/latest runs from spreadsheet analysis.py
nikolayilyin: In my opinion code duplication is very bad thing but jupyter with libraries might became too cumbersome to be used. 
I have tried something similar and it seems it did not work - no one use that library, even me (=

nikolayilyin:
test!

dimaopen (approved):

PR #3536, opened 2022-06-04
Tour mode choice

zneedell:
**EARLY ATTEMPT--NEEDS TO BE TESTED THOROUGHLY**

This PR introduces a `Tour Mode` category, which is distinct from a trip mode. The tour mode can either be `CAR_BASED`, `BIKE_BASED`, or `WALK_BASED`, and defines whether you are choosing to take a personal vehicle with you on the tour. Each Tour can have one Tour Mode in the same way that each Trip can have one Trip Mode. A relevant tour mode choice strategy is stored in the `BeamPlan`. 

If a person is on a tour with the tour mode already defined, the tour mode serves to limit the type of modes they can choose between (although in practice this doesn't really do anything, because e.g. if you already are on a `WALK_BASED` tour you shouldn't have access to any of your household vehicles until you get back home, and shared vehicle trips are still allowed on `WALK_BASED` tours). I've added the ability to assign a tour mode strategy an associated vehicle ID, which is intended to mimic the `tourPersonalVehicle` and potentially replace it eventually, but that isn't used yet.

The main change to simulation behavior happens when a person goes into `ChoosesMode` with no tour mode defined. Before getting routes from the router, the agent uses the skims to estimate the utility for every available trip on the upcoming tour and then choose which (if any) available household vehicle to use for the tour. This sets the TourMode for the tour and limits the available modes for all remaining trips.

One main goal of this is to eliminate the long walk trips that arise when we're using ActivitySim plans that e.g. look like `HOME -> short trip -> ACTIVITY 1 -> long trip -> ACTIVITY 2 ...`, where the agent chooses to walk or take transit for the first trip and is then stranded without a car for the second trip.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3536)


src/main/scala/beam/router/Modes.scala
```diff
@@ -245,3 +247,69 @@ object Modes {
...
+            HOV2_TELEPORTATION,
+            HOV3_TELEPORTATION
+          )
+        )
+
+    case object CAR_BASED
+        extends BeamTourMode(
+          "car_based",
+          Car,
+          Seq[BeamMode](CAR, CAR_HOV2, CAR_HOV3),
```
Xuan-1998, line 324: How about drive_transit? This might result drive_transit be categorized into walk_based tour
image

src/main/scala/beam/router/Modes.scala
```diff
@@ -245,3 +247,69 @@ object Modes {
...
+          Seq[BeamMode](CAR, CAR_HOV2, CAR_HOV3)
+        )
+
+    case object BIKE_BASED extends BeamTourMode("bike_based", Car, Seq[BeamMode](BIKE), Seq[BeamMode](BIKE))
```
Xuan-1998: Same as above, we don't have bike_transit?

src/main/scala/beam/router/Modes.scala
```diff
@@ -245,3 +247,69 @@ object Modes {
...
+      case _          => true
+    }
+  }
+
+  object BeamTourMode extends StringEnum[BeamTourMode] with StringCirceEnum[BeamTourMode] {
+
+    override val values: immutable.IndexedSeq[BeamTourMode] = findValues
+
+    val enabledModes: Map[BeamMode, Seq[BeamMode]] =
+      Map[BeamMode, Seq[BeamMode]](CAR -> Seq(DRIVE_TRANSIT), BIKE -> Seq(BIKE_TRANSIT))
```
Xuan-1998, line 291: why no WALK here for availableFirstAndLastLegModes?

Xuan-1998 (changes requested):

src/main/scala/beam/router/Modes.scala
```diff
@@ -245,3 +247,69 @@ object Modes {
...
+
+    // TODO: Also allow use of shared bikes/cars in walk based tours
+    case object WALK_BASED
+        extends BeamTourMode(
```
Xuan-1998, line 295: Think we could have 
`id: Id[BeamVehicleType]`
And then distinguish shared bike by
`def isSharedVehicle: Boolean = id.toString.startsWith("sharedVehicle")`
And then filter shared bike by
```
val bikeVehicleType = beamScenario.vehicleTypes.values
  .filterNot(_.isSharedVehicle)
  .find(_.vehicleCategory == VehicleCategory.Bike)
  .getOrElse(throw new RuntimeException("Bike not found in vehicle types."))
```
And then create a new object of shared bike to put in

Xuan-1998 (changes requested):

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -259,6 +252,49 @@ trait ChoosesMode {
...
+          val dummyVehicleType = beamScenario.vehicleTypes(dummyRHVehicle.vehicleTypeId)
+          val currentTour = _experiencedBeamPlan.getTourContaining(nextAct)
+          val tourModeCosts = beamServices.skims.od_skimmer.getTourModeCosts(
+            modesToQuery,
+            currentTour,
+            dummyRHVehicle.vehicleTypeId,
+            dummyVehicleType,
+            beamScenario.fuelTypePrices(dummyVehicleType.primaryFuelType)
+          )
+          val modeToTourMode =
```
zneedell, line 274: @Xuan-1998 So as I understand it you're suggesting that the problem here is that if the agents have access to a shared bike when they hit this logic, then they end up potentially choosing a `BIKE_BASED` tour, even though they need to leave the bike behind after the first trip? That definitely seems like it could be causing the problem. Perhaps we need to add some other logic that only allows a `BIKE_BASED` tour if they have access to a personal bike (and similar for personal vs shared cars)

src/main/scala/beam/router/Modes.scala
zneedell: I think you're on to something -- how about rather than giving this an `id` attribute we change the `allowedBeamModes` to a private attribute and give the class a new method like `def allowedBeamModes(vehicles: Vector[VehicleOrToken]): Seq[BeamMode]` that checks whether the vehicles are shared or not before returning the allowed beam modes, having `BIKE_BASED` not return `BIKE` if the only bikes available are shared bikes.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -259,6 +252,49 @@ trait ChoosesMode {
...
+            Some(firstAndLastTripModeToTourModeOption)
+          )
+      }
+
+      _experiencedBeamPlan.putStrategy(
+        _experiencedBeamPlan.getTourContaining(nextAct),
+        TourModeChoiceStrategy(chosenCurrentTourMode)
+      )
+
+      val availableModesGivenTourMode = getAvailableModesGivenTourMode(
```
zneedell, line 292: I think the one other change here is that if the person still chooses a `BIKE_BASED` tour, we cannot let them use a shared bike. Perhaps we need to add another function that is `getAvailableVehiclesGivenTourMode` before this one is called that filters our shared vehicles if we're on a `BIKE_BASED` or `CAR_BASED` tour

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
zneedell: And just thinking through a to do here -- we don't currently factor in walking distance to the available street vehicles into these costs. Perhaps eventually we should call the skimmer to factor in those access/egress costs as well. Don't think it's worth it for now, though.

src/main/scala/beam/router/Modes.scala
zneedell: Oh, wait a minute, this is definitely a typo and should be `BeamTourMode("bike_based", Bike, Seq[BeamMode](BIKE), Seq[BeamMode](BIKE))`. My bad!

zneedell: Although it looks like maybe this has no effect

Xuan-1998:
Let's open this one to be ready to get slack notification

Xuan-1998:
Let's open this one to be ready to get slack notification

Xuan-1998:
Let's merge to Dima's branch, and if Dima could do a last round check it would be wonderful

PR #3537, opened 2022-06-06
Move all utilities into an upper level folder for easier discovery

### LINKED ISSUE #3445: Move all utilities into an upper level folder for easier discovery
Utilities that use `extends App` or defines `main(args: String[])` to be run standalone should probably be moved to an upper folder so they are all easily discovered

alex-vv:
fixes #3445

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3537)


dimaopen (commented):
Generally it looks good. Though we have multiple files that moved without changes. I think that their package definition should be changed to the corresponding new folder.

alex-vv:
> Generally it looks good. Though we have multiple files that moved without changes. I think that their package definition should be changed to the corresponding new folder.

thanks, pushed a fix

alex-vv:
test!

dimaopen (commented):
Utils like these `beam.utils.gtfs.GTFSToShape`, `beam.utils.RouteRequester` are not going to be moved to scripts?

alex-vv:
> Utils like these beam.utils.gtfs.GTFSToShape, beam.utils.RouteRequester are not going to be moved to scripts?

yes, it looks I missed these, will move them as well

alex-vv:
test!

dimaopen (approved):

PR #3539, opened 2022-06-08
Exiting Beam if no progress

### LINKED ISSUE #3538: Kill beam if no simulation progress

dimaopen:
It introduces 2 new parameters:

initialiazationStuckThresholdMin
simulationStuckThresholdMin

The first one is used for the initialization phase (when time is 0). Actually the maximum of these 2 params is taken.

The second one is used during the simulation run.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3539)

### DIFF
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -765,8 +765,8 @@ beam.debug {
     writeRealizedModeChoiceFile = "boolean | false"
     messageLogging = "boolean | false"
     # the max of the next 2 values is taken for the initialization step
-    simulationStuckThresholdMin = "int | 999999"
-    initialiazationStuckThresholdMin = "int | 180"
+    maxSimulationStepTimeBeforeConsideredStuckMin = "int | 999999"
+    maxInitializationTimeBeforeConsideredStuckMin = "int | 180"
 }
 
 beam.logger.keepConsoleAppenderOn = "boolean | true"
--- a/src/main/resources/beam-template.conf
+++ b/src/main/resources/beam-template.conf
@@ -764,6 +764,9 @@ beam.debug {
 
     writeRealizedModeChoiceFile = "boolean | false"
     messageLogging = "boolean | false"
+    # the max of the next 2 values is taken for the initialization step
+    simulationStuckThresholdMin = "int | 999999"
+    initialiazationStuckThresholdMin = "int | 180"
 }
 
 beam.logger.keepConsoleAppenderOn = "boolean | true"
--- a/src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
+++ b/src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
@@ -47,6 +47,8 @@ object BeamAgentScheduler {
 
   case object Monitor extends SchedulerMessage
 
+  case object SimulationStuckCheck extends SchedulerMessage
+
   case object RequestCurrentTime extends SchedulerMessage
 
   case object SkipOverBadActors extends SchedulerMessage
@@ -146,6 +148,8 @@ class BeamAgentScheduler(
   private var idCount: Long = 0L
   private var startSender: ActorRef = _
   private var nowInSeconds: Int = 0
+  private var nowUpdateTime: Long = System.currentTimeMillis()
+  private var longestNowUpdateDelay: Long = 0
 
   private val maybeTriggerMeasurer: Option[TriggerMeasurer] = if (beamConfig.beam.debug.triggerMeasurer.enabled) {
     Some(new TriggerMeasurer(beamConfig.beam.debug.triggerMeasurer))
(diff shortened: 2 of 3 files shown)


src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -164,6 +168,10 @@ class BeamAgentScheduler(
   private val scheduledTriggerToStuckTimes: mutable.HashMap[ScheduledTrigger, Int] =
     mutable.HashMap.empty
 
+  import scala.concurrent.duration._
+
+  private val shutdownOnStuckTask: Cancellable =
+    context.system.scheduler.scheduleWithFixedDelay(10.minutes, 1.minute, self, SimulationStuckCheck)
```
JustinPihony, line 174: Should the checks be more frequent? This wouldn't be too expensive. Or make this also configurable.

src/main/resources/beam-template.conf
```diff
@@ -764,6 +764,9 @@ beam.debug {
 
     writeRealizedModeChoiceFile = "boolean | false"
     messageLogging = "boolean | false"
+    # the max of the next 2 values is taken for the initialization step
+    simulationStuckThresholdMin = "int | 999999"
+    initialiazationStuckThresholdMin = "int | 180"
```
JustinPihony: What would the suggested values be in a typical run? Wouldn't init time always be lower? As such, should it be the min of the two - otherwise, as max, it will always choose the sim stuck instead of the init stuck - rendering the latter pointless.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
```diff
@@ -393,19 +419,21 @@ class BeamAgentScheduler(
...
             log.info(
-              "Hour " + nowInSeconds / 3600.0 + " completed. " + math.round(
+              "Hour " + nowInSeconds / 3600.0 + " completed. Longest delay " + longestNowUpdateDelay + " ms. "
+              + math.round(
                 10 * (Runtime.getRuntime.totalMemory() - Runtime.getRuntime
                   .freeMemory()) / Math
                   .pow(1000, 3)
               ) / 10.0 + "(GB)"
             )
+            longestNowUpdateDelay = 0
```
JustinPihony, line 433: Just because an hour completed doesn't mean the longest update should reset? It would still be valid that the longest delay is the longest delay...

src/main/resources/beam-template.conf
```diff
@@ -764,6 +764,9 @@ beam.debug {
 
     writeRealizedModeChoiceFile = "boolean | false"
     messageLogging = "boolean | false"
+    # the max of the next 2 values is taken for the initialization step
+    simulationStuckThresholdMin = "int | 999999"
```
JustinPihony: After reading the implementation I propose a rename to something like `maxSimulationStepTimeBeforeConsideredStuck` and `maxInitializationTimeBeforeConsideredStuck`.

heyrutvik:
@dimaopen I re-produce the gemini stuckness locally and this changes does kill the BEAM if it stays stuck longer than configured value. :+1:

src/main/resources/beam-template.conf
heyrutvik: If we agree to change the names please add unit of the value as well.. like in the original names.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
dimaopen: This longest delay is just a hint for the user about how long his simulation is delayed on some simulation time. I think that it's more interesting to see this delay for each hour comparing to the whole simulation.

src/main/resources/beam-template.conf
dimaopen: That's correct. Some big simulation may have a very long init step (when the time is 0) because of a big number of agents that need to be initialized. It may last an hour or more. At the init step the probability of getting stuck is very low.
That value `simulationStuckThresholdMin = "int | 999999"` means that this feature (exiting beam in case of it got stuck) is disabled.
On medium size simulation a valid delay time during simulation may be a few minutes (1-3). I think that if the user set `simulationStuckThresholdMin` to 30 mins it would be effectively exiting beam if it is stuck.
On the big simulations I think the user could set it to 60 minutes. Then `initialiazationStuckThresholdMin` works fine.

src/main/scala/beam/agentsim/scheduler/BeamAgentScheduler.scala
dimaopen: I don't think that we need to have this more frequent. We just check once a minute if the delay is more then say 30 minutes. And if it is then we terminate Beam.

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

nikolayilyin (approved):
lgtm

PR #3551, opened 2022-07-05
CPU and RAM usage logging on aws

### LINKED ISSUE #3547: AWS: RAM and CPU usage statisticks.
We need some way to measure how much CPU and Memory were used during simulation on AWS instance.
It seems to be tricky to get that information from inside the JVM, so, another solution - to use Linux system apps to gather information.

The runtime statistic should be something ready to be analyzed - a csv file.
The statistic row should contain a timestamp, CPU and RAM usage of the whole system. 
If there is a simple way to get separately usage of CPU and RAM by BEAM - it should be included, otherwise we will be fine with total system statistics.

The statistic file should be copied into the output folder (and to s3 with the rest of the output)

nikolayilyin:
A shell script that collect CPU and RAM usage from system utilities and write out the collected data into a csv file

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3551)


nikolayilyin:
test!

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -119,11 +121,27 @@
             sleep 5s
             cd -
       path: /home/ubuntu/install-and-run-helics-scripts.sh
+    - content: |
+            #!/bin/bash
+            timeout=$1
+            echo "date,time,CPU usage,RAM used,RAM available"
+            while sleep $timeout
+            do
+                    timestamp_CPU=$(vmstat 1 3 -SM -a -w -t | python3 -c 'import sys; ll=sys.stdin.readlines()[-1].split(); print(ll[-2] + ", " + ll[-1] + ", " + str(100 - int(ll[-5])))')
```
alex-vv, line 130: it looks like `vmstat -t` will give enough information and you have just one line to parse

nikolayilyin: The issue is that it will give memory in megabytes. I wanted Gigabytes - because of that I used 'free'

alex-vv (approved):

PR #3552, opened 2022-07-05
Runs Jupyter at AWS, standalone or together with BEAM

### LINKED ISSUE #3540: Jupyter on AWS. A gradle command to start jupyter on aws instance
We need a simple solution in order to get jupyter running on aws instance.

Currently we have a gradle tasks 'jupyterstart' and 'jupyterstop' to start and stop jupyter image in local docker service.

I think we need some hard-coded instance type for jupyter instances (r5.2xlarge) by default with possibility to specify the instance type manually via parameter.

It seems we might also need a parameter for gradle 'deploy' task that would allow us to run jupyter as well as BEAM on the same aws instance.

alex-vv:
Fixes #3540

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3552)


alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -132,10 +132,27 @@
                     echo $timestamp_CPU, $ram_used_available
             done
       path: /home/ubuntu/write-cpu-ram-usage.sh
-
+    - content: |
+            #!/bin/bash
+            pip install setuptools
+            pip install jupyter
+      path: /home/ubuntu/install-jupyter.sh
```
nikolayilyin: Can we just run the same docker image without installing jupyter?

nikolayilyin (changes requested):

aws/src/main/python/beam_lambda/lambda_function.py
nikolayilyin: Otherwise, if using of docker image is impossible, we might install jupyter on instance and made a new beam AWS image to avoid installing jupyter on it every time. 
Personally I do not like installing things in lambda - it might suddenly break in future.

nikolayilyin: Another argument against installing jupyter and to use docker - we might enhance the docker image with required python packages and re-use that image everywhere, otherwise it will be required to configure instance every time, or, change lambda with additional installation packages, which might be a pain.

alex-vv: I am now using the same local gradle task to run Jupyter in lambda

nikolayilyin (approved):
lgtm

nikolayilyin:
test!

alex-vv:
test!

PR #3555, opened 2022-07-08
Storing the scheduler state in case of beam exists with an error

dimaopen:
1. Remoived maxInitializationTimeBeforeConsideredStuckMin param as we actually don't need it.
2. Enabled killing beam by default
3. Saving internal scheduler state

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3555)


nikolayilyin (approved):

nikolayilyin:
test!

PR #3559, opened 2022-07-13
External script at AWS EC2 instances that saves beam thread dump

### LINKED ISSUE #3550: Debugging: when simulation is terminated do a thread dump into a file.
Recently we introduced a functionality to terminate simulation if there was no progress for configured time (#3539)
In addition to this functionality it would be helpful to have a thread dump at the time of ending BEAM and the content of the scheduler queue.
We need that dump in the file in the output folder, so, it will be moved into s3 output folder with the rest of the files.

dimaopen:
1. Moved scheduler state files to the root of the output folder.
2. Save thread dump by scheduler.
3. Save thread dump and kill beam by an external process.

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3559)

### DIFF
--- a/aws/src/main/python/beam_lambda/lambda_function.py
+++ b/aws/src/main/python/beam_lambda/lambda_function.py
@@ -97,10 +97,9 @@ write_files:
           curl -X POST -H 'Content-type: application/json' --data-binary @/tmp/slack.json $SLACK_HOOK_WITH_TOKEN
       path: /tmp/slack.sh
     - content: |
-          0 * * * * curl -X POST -H "Content-type: application/json" --data '"'"'{"$(ec2metadata --instance-type) instance $(ec2metadata --instance-id) running... \\n Batch [$UID] completed and instance of type $(ec2metadata --instance-type) is still running in $REGION since last $(($(($(date +%s) - $(cat /tmp/.starttime))) / 3600)) Hour $(($(($(date +%s) - $(cat /tmp/.starttime))) / 60)) Minute."}'"'"
           */10 * * * * /home/ubuntu/beam_stuck_guard.sh
           
-      path: /tmp/slack_notification
+      path: /tmp/cron_jobs
     - content: |
             #!/bin/bash
             pip install setuptools
@@ -246,7 +245,7 @@ runcmd:
   - chmod +x /tmp/slack.sh
   - echo "notification sent..."
   - echo "notification saved..."
-  - crontab /tmp/slack_notification
+  - crontab /tmp/cron_jobs
   - crontab -l
   - echo "notification scheduled..."
 
--- a/aws/src/main/python/beam_lambda/lambda_function.py
+++ b/aws/src/main/python/beam_lambda/lambda_function.py
@@ -98,6 +98,8 @@ write_files:
       path: /tmp/slack.sh
     - content: |
           0 * * * * curl -X POST -H "Content-type: application/json" --data '"'"'{"$(ec2metadata --instance-type) instance $(ec2metadata --instance-id) running... \\n Batch [$UID] completed and instance of type $(ec2metadata --instance-type) is still running in $REGION since last $(($(($(date +%s) - $(cat /tmp/.starttime))) / 3600)) Hour $(($(($(date +%s) - $(cat /tmp/.starttime))) / 60)) Minute."}'"'"
+          */10 * * * * /home/ubuntu/beam_stuck_guard.sh
+          
       path: /tmp/slack_notification
     - content: |
             #!/bin/bash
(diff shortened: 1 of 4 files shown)


aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -98,6 +98,8 @@
       path: /tmp/slack.sh
     - content: |
           0 * * * * curl -X POST -H "Content-type: application/json" --data '"'"'{"$(ec2metadata --instance-type) instance $(ec2metadata --instance-id) running... \\n Batch [$UID] completed and instance of type $(ec2metadata --instance-type) is still running in $REGION since last $(($(($(date +%s) - $(cat /tmp/.starttime))) / 3600)) Hour $(($(($(date +%s) - $(cat /tmp/.starttime))) / 60)) Minute."}'"'"
+          */10 * * * * /home/ubuntu/beam_stuck_guard.sh
+          
       path: /tmp/slack_notification
```
nikolayilyin: I do not understand what this part do. 
Our slack notifications do not contain text from first row of file - I am not sure it is working at all, or, if it is - where to look for these notifications.
The added row should sent slack notifications too?
It seems the content of 'beam stuck guard' should kill beam if it is stuck, if so, this is not a place for such script.

nikolayilyin (changes requested):
I think the content of /tmp/slack_notification does not match it's name, either the file name should be changed, or the new script should be moved.

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -98,6 +98,8 @@
       path: /tmp/slack.sh
     - content: |
           0 * * * * curl -X POST -H "Content-type: application/json" --data '"'"'{"$(ec2metadata --instance-type) instance $(ec2metadata --instance-id) running... \\n Batch [$UID] completed and instance of type $(ec2metadata --instance-type) is still running in $REGION since last $(($(($(date +%s) - $(cat /tmp/.starttime))) / 3600)) Hour $(($(($(date +%s) - $(cat /tmp/.starttime))) / 60)) Minute."}'"'"
+          */10 * * * * /home/ubuntu/beam_stuck_guard.sh
```
JustinPihony, line 100: This should be stored somewhere else for backup, but overall it should be built into the image - not constantly added via a lambda. We could add a check for the appropriate files and setup in the image updating lambda (but that is likely not necessary)

JustinPihony: The biggest question mark would be if 10 minutes is long enough to get to the point of starting up beam from instance start - granted I don't think that changes with this impl or my suggested one

aws/src/main/python/beam_lambda/lambda_function.py
nikolayilyin: I suggest to remove old and not working part from slack-notification script and rename the script into something appropriate to its functionality - 'cron-tasks' or something like that

aws/src/main/python/beam_lambda/lambda_function.py
dimaopen: > This should be stored somewhere else for backup, but overall it should be built into the image

I generally agree that it's better to keep the script in the image. How can I do it? By the way if we need to fix/modify the script in the future would it be an easy task (comparing to changing the lambda)?
The script itself is at https://github.com/LBNL-UCB-STI/beam/blob/4cf7a9811d2dbf84b80a0b19ca3ccd1d723872b2/aws/src/main/python/beam_lambda/lambda_function.py#L138-L155

dimaopen: > The biggest question mark would be if 10 minutes is long enough to get to the point of starting up beam from instance start

If beam is not run that script just exists.
`pgrep -f RunBeam || exit 0 `

nikolayilyin (commented):
lgtm

nikolayilyin:
test!

nikolayilyin (approved):

aws/src/main/python/beam_lambda/lambda_function.py
JustinPihony: For future reference, the way to build it into the image is to take the AMI and start up an instance (usually you only need a small instance for rebasing). Then make the changes needed inside of that instance. Once done go back to the AWS console, choose the instance, and hit "Image and Templates > Create Image". Next you go to the AMIs and find the new AMI and hit Copy Image, changing the region to us-west-1 and us-east-1 to create the base image for those regions also. Finally you'll need to change the lambda environment variables to point to the new image ids for simulateBeam. Oh, and you also need to update jenkins at https://beam-ci.tk/configureClouds/ - changing the AMI IDs which reference the old one to the new one. It is more onerous than the current way, but good if steady. If expected to change, then yes, leaving in the lambda is better. I would assume this to be steady. Ultimately, it probably doesn't add much overhead, other than clutter to the lambda code, so fine to leave.

PR #3571, opened 2022-07-29
Refactor `jupyterEC2` task to use `simulateBeam` lambda

alex-vv:
Instead of using a custom lambda, re-use the deploy one and run jupyter in a standard way via gradle task

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3571)


aws/build.gradle
```diff
@@ -246,18 +246,26 @@ task terminateEC2(type: AWSLambdaInvokeTask) {
...
-            }
-            """
+        payload = """{
+          "title": "${(project.findProperty('title') ?: 'jupyter') + '_' + getCurrentGitUserEmail()}",
+          "git_user_email": "${getCurrentGitUserEmail()}",
+          "branch": "develop",
+          "storage_size": ${project.findProperty('storageSize') ?: 128},
+          "instance_type": "${project.findProperty('instanceType') ?: 'r5.xlarge'}",
+          "region": "${project.findProperty('region') ?: defaultRegion}",
+          "shutdown_behaviour": "terminate",
```
nikolayilyin: I think we need to be able to change branch, shutdown_behaviour and storage_size by changing gradle parameters, also, I think, the default storage size should be 256

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -158,7 +158,10 @@
 runcmd:
   - sudo chmod +x /home/ubuntu/install-and-run-helics-scripts.sh
   - sudo chmod +x /home/ubuntu/write-cpu-ram-usage.sh
-  - sudo chmod +x /home/ubuntu/beam_stuck_guard.sh
+  - if [ "$RUN_BEAM" = "True" ]
+  - then
+  -   sudo chmod +x /home/ubuntu/beam_stuck_guard.sh
+  - fi
```
nikolayilyin: It seems it is not require any changes here because crontab is used in later block only if RUN_BEAM is true

nikolayilyin (changes requested):
Few parameters needs to be variable, the rest LGTM.

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -214,114 +215,120 @@
...
+          \\"error\\":\\"$error\\",
+          \\"warning\\":\\"$warn\\",
+          \\"sigopt_client_id\\":\\"$SIGOPT_CLIENT_ID\\",
+          \\"sigopt_dev_id\\":\\"$SIGOPT_DEV_ID\\"
+        }
+      }" $(ec2metadata --instance-id) $(ec2metadata --instance-type) $(ec2metadata --public-hostname) $(ec2metadata --public-hostname) "${s3p#","}")
+  -   /tmp/slack.sh "$bye_msg"
+  -   curl -X POST "https://ca4ircx74d.execute-api.us-east-2.amazonaws.com/production/spreadsheet" -H "Content-Type:application/json" --data "$stop_json"
+  -   $END_SCRIPT
+  -   sudo shutdown -h +$SHUTDOWN_WAIT
```
nikolayilyin: Also giving it a second thought, let's keep the automatic shutdown at 8 hours by-default, and at any value if it was set as gradle parameter.

aws/build.gradle
alex-vv: done

aws/src/main/python/beam_lambda/lambda_function.py
alex-vv: done

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

nikolayilyin:
test!

nikolayilyin (approved):
Well done, LGTM

nikolayilyin:
test!

alex-vv:
test!

PR #3572, opened 2022-07-29
Upgrade gradle 7.1.1 to 7.5

frikit:
upgrade sphinx 1.0.3.Final to 2.10.1
upgrade to latest beam utils 2.17 to 2.19
upgrade scala 2.12.10 to 2.12.16(beam-utils has the same version now)
Fix deprecated warnings from scala for new format method
remove deprecated jcenter per gradle docs
upgradedocker compose plugin

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3572)


nikolayilyin (approved):
The only concern I have here is the minor scala upgrade, so, let's look at possible issues of our team after this PR is merged, the rest look good to me

Issue #3574, opened 2022-08-01
Potential fixes: Transit usage in PILATES


zneedell:
Running list of ideas to improve sensitivity to transit service with ActivitySim/BEAM integration:

- [x] Go back to recording experienced trip in skims (while having the option to include non-chosen alternatives as well)
- [x] Use real rather than generalized times and costs
- [x] Go back to merging in non-selected plans (after fix in #3560)
- [ ] Allow for specification of transit mode in input plans (e.g. if plans say `WLK_COM_WALK` choose a commuter rail route if possible)
- [x] Fix representation of not-used BEAM modes in skims (e.g. don't let people choose express bus, toll routes) [requires change to PILATES]
- [x] Add buffer time for drive transit so we don't have as many missed trips

zneedell:
- [x]  Locations aren't being mapped to the correct TAZ maybe? -- create a linkId -> TAZ map when loading the network

PR #3575, opened 2022-08-02
Use TAZ geoms for skim origin/destination

zneedell:
Build a map of linkId -> TAZ when loading the network and use that to look up the starting/ending TAZ of beam trips that get sent to the skimmer

Partially addresses #3574

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575)


src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -1452,14 +1468,33 @@ class PersonAgent(
...
+          (origGeo, destGeo)
+        } else {
+
+          beamScenario.exchangeGeoMap match {
+            case Some(geoMap) =>
+              val origGeo = geoMap.getTAZ(origCoord)
+              val destGeo = geoMap.getTAZ(destCoord)
+              logger.warn(
+                s"Could not find linkId->taz mapping so falling back on default: Trip from $origGeo to $destGeo"
+              )
```
nikolayilyin: How many of the following log entries there will be?
Maybe it is worth writing these coordinate pairs into a separate file and putting into beamLog just a warning about a number of these pairs?
Otherwise beamLog will be too big to work with it without any benefits for us.

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
```diff
@@ -67,6 +80,39 @@ class TAZTreeMap(val tazQuadTree: QuadTree[TAZ], val useCache: Boolean = false)
...
+      }
+      logger.info(
+        "Matched "
+        + linkIdToTAZMapping.size.toString +
+        " links to TAZs, failed to match "
+        + unmatchedLinkIds.size.toString +
+        " links"
+      )
+    }
+  }
```
nikolayilyin: How long will it take to process a big map such as sf-bay or new-york?
If it might take more than 10 minutes - let's change the log text by adding there the word 'Completed' for 'beam stuck guard'.

nikolayilyin (changes requested):
A few nitpicks, also for BeamConfig after generation it is required to run 'fmt', because the generation isn't doing it.

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
zneedell: It ran in a few minutes on my local machine for sf-bay so not too bad. I'd imagine this could be parallelized pretty easily, too. I tried building a parallel collection and then converting it to a map but that slowed things down, but I'm not too familiar with best practices for parallel collections -- do you have any ideas for making that work?

haitamlaarabi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @nikolayilyin and @zneedell)

___
*[`src/main/scala/beam/agentsim/infrastructure/taz/TAZ.scala` line 18 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575#-N8W0xyB4Gv4SdF7CXTn:-N8W0xyB4Gv4SdF7CXTo:bvdxmh4) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e390368a2e5f1ae3b30e729a7ccd86fdea9b6e2f/src/main/scala/beam/agentsim/infrastructure/taz/TAZ.scala#L18)):*
> ```Scala
>   * @param areaInSquareMeters area of TAZ
>   */
> class TAZ(val tazId: Id[TAZ], val coord: Coord, val areaInSquareMeters: Double, val geometry: Option[Geometry]) {
> ```

I have been wanting to add geometry in TAZ for so long but I tried to avoid that. Glad that you did it. It will help with matching and repositioning. 
___
*[`src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala` line 131 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575#-N8W0V654YASW0p8zeLt:-N8W0V654YASW0p8zeLu:bbpgt44) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e390368a2e5f1ae3b30e729a7ccd86fdea9b6e2f/src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala#L131)):*
> ```Scala
>         case (id, link) =>
>           val linkEndCoord = link.getToNode.getCoord
>           val foundTaz = TAZTreeMap.ringSearch(
> ```

ringSearch is using coordinates and not geometry. In case of large TAZ and the linkEndCoord at the edge, might lead to match the link to a nearby TAZ and not to the TAZ that contains the point, and later on that link might end up not matched to any TAZ.
___
*[`src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala` line 141 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575#-N8W07jQ-VIHsJ0o98b9:-N8W07jR2XqNH_zO-NrI:b56pxeh) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e390368a2e5f1ae3b30e729a7ccd86fdea9b6e2f/src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala#L141)):*
> ```Scala
>             else None
>           }
>           foundTaz match {
> ```

ringSearch returns a list of TAZs, I expect to see in foundTAZ a list of Options and one of them contains the actual TAZ. In this case "match" is not the right way to go.

zneedell:
*[`src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala` line 141 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575#-N8W07jQ-VIHsJ0o98b9:-N8W3yG469dgGeNG12aR:bc4xnx2) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e390368a2e5f1ae3b30e729a7ccd86fdea9b6e2f/src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala#L141)):*

ringSearch returns a list of TAZs, I expect to see in foundTAZ a list of Options and one of them contains the actual TAZ. In this case "match" is not the right way to go.
</blockquote></details>

My understanding (which might be wrong!) is that ringSearch first gets a list of tazs within the given distance of your point and then runs a filter on them and returns the first taz that passes the filter (and expands to a larger radius if not).

So anyway, the intended behavior is that ringSearch only returns an option for the single TAZ satisfying the condition ` (taz.geometry.exists(_.contains(GeometryUtils.createGeotoolsPoint(linkEndCoord))))` (which should be the single unique TAZ containing the point if the TAZ geoms are non overlapping)

So basically the goal of ringSearch is just to narrow down the set of possible TAZs that could contain the point so we don't need to do geom.contains on all of them

haitamlaarabi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @nikolayilyin and @zneedell)

___
*[`src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala` line 141 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575#-N8W07jQ-VIHsJ0o98b9:-N8WBJJZE4pUOZJzYD-0:b-md38ut) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e390368a2e5f1ae3b30e729a7ccd86fdea9b6e2f/src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala#L141)):*
<details><summary><i>Previously, zneedell (Zach Needell) wrote…</i></summary><blockquote>

My understanding (which might be wrong!) is that ringSearch first gets a list of tazs within the given distance of your point and then runs a filter on them and returns the first taz that passes the filter (and expands to a larger radius if not).

So anyway, the intended behavior is that ringSearch only returns an option for the single TAZ satisfying the condition ` (taz.geometry.exists(_.contains(GeometryUtils.createGeotoolsPoint(linkEndCoord))))` (which should be the single unique TAZ containing the point if the TAZ geoms are non overlapping)

So basically the goal of ringSearch is just to narrow down the set of possible TAZs that could contain the point so we don't need to do geom.contains on all of them 
</blockquote></details>

I did not notice that it takes an implicit filter function. It makes sense now. thanks

haitamlaarabi (approved):
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575)* status: 0 of 7 files reviewed, 5 unresolved discussions (waiting on @nikolayilyin and @zneedell)

___
*[`src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala` line 131 at r3](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3575#-N8W0V654YASW0p8zeLt:-N8WBeIEES9JVXhoKAYi:b-dm5ip0) ([raw file](https://github.com/lbnl-ucb-sti/beam/blob/e390368a2e5f1ae3b30e729a7ccd86fdea9b6e2f/src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala#L131)):*

ringSearch is using coordinates and not geometry. In case of large TAZ and the linkEndCoord at the edge, might lead to match the link to a nearby TAZ and not to the TAZ that contains the point, and later on that link might end up not matched to any TAZ.
</blockquote></details>

Please disregard this comment then

nikolayilyin (approved):
lgtm

zneedell:
Test!

nikolayilyin:
test!

nikolayilyin:
test!

PR #3578, opened 2022-08-02
Use [non-generalized] times and costs in asim skimmer

zneedell:
ActivitySim does its own conversion of time/cost to generalized time/cost so we should probably just be feeding it raw values

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3578)


zneedell:
Partially addressed #3574

nikolayilyin (approved):
lgtm

nikolayilyin:
test!

PR #3579, opened 2022-08-03
Latest code update from GEMINI-XFC work

haitamlaarabi:
Various code changes addressing different problems and capability enhancement

---
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3579)


haitamlaarabi:
Test!

haitamlaarabi:
Test!

src/main/python/gemini/helper.py
```diff
@@ -0,0 +1,7 @@
+import pandas as pd
+
+def read_file(filename):
+    compression = None
+    if filename.endswith(".gz"):
+        compression = 'gzip'
+    return pd.read_csv(filename, sep=",", index_col=None, header=0, compression=compression)
```
nikolayilyin: Pandas by-default understand compression from the path of the file, there is no need to do it manually.

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -53,25 +64,54 @@ class HouseholdFleetManager(
...
+          val (_, activityType, location) = homeAndStartingWorkLocations(personId)
+          val inquiry = ParkingInquiry.init(
+            SpaceTime(location, 0),
+            activityType,
+            VehicleManager.getReservedFor(vehicle.vehicleManagerId.get).get,
+            beamVehicle = Option(vehicle),
+            triggerId = triggerId,
+            searchMode = ParkingSearchMode.Init
+          )
+          // TODO Overnight charging is still a work in progress and might produce unexpected results
```
nikolayilyin: This comment is not adding anything useful - let's delete it

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -53,25 +64,54 @@ class HouseholdFleetManager(
...
+          if (vehicle.isEV && beamConfig.beam.agentsim.chargingNetworkManager.overnightChargingEnabled) {
+            logger.info(s"Overnight charging vehicle $vehicle with state of charge ${vehicle.getStateOfCharge}")
```
nikolayilyin: How many log entries this might produce?
Maybe instead we need to put this information somewhere else - into event file as events or collect by some info collector and plot as a graph?
Or change the log level to debug.

src/main/scala/beam/agentsim/agents/household/HouseholdFleetManager.scala
```diff
@@ -136,13 +176,20 @@ class HouseholdFleetManager(
...
     case Finish =>
       context.stop(self)
-
     case Success =>
-    case pir: ParkingInquiryResponse =>
-      logger.error(s"STUCK with ParkingInquiryResponse: $pir")
-
     case x =>
       logger.warn(s"No handler for $x")
   }
```
nikolayilyin: It wasn't changed in this PR, but probably this should be ERROR, warnings are ignored mostly

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -451,6 +424,7 @@ trait ChoosesMode {
       var requestId: Option[Int] = None
       // Form and send requests
 
+      var makeCurrentTourModeNone = false // to replan when personal vehicles are not available
```
nikolayilyin: Let's change the name of the flag according to reason of the value change, i.e. into 'householdVehiclesWereNotAvailable'.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -586,7 +575,8 @@ trait ChoosesMode {
           logDebug(m.toString)
       }
       val newPersonData = choosesModeData.copy(
-        personData = choosesModeData.personData.copy(currentTourMode = correctedCurrentTourMode),
+        personData = choosesModeData.personData
+          .copy(currentTourMode = if (makeCurrentTourModeNone) None else correctedCurrentTourMode),
```
nikolayilyin: Should we drop the flag into original value after it was used?

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -231,11 +242,33 @@ trait ChoosesParking extends {
...
+        vehicle.isRefuelNeeded(
+          remainingTourDistance + conf.home.refuelRequiredThresholdInMeters,
+          conf.noRefuelThresholdInMeters
+        )
+      case _ =>
+        vehicle.isRefuelNeeded(
+          remainingTourDistance + conf.refuelRequiredThresholdInMeters,
+          conf.noRefuelThresholdInMeters
+        )
+        true
```
nikolayilyin: Looks like some temporal hack, we need to leave here either the method call or the constant

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -53,6 +55,14 @@ class ChargingNetworkManager(
   private var timeSpentToPlanEnergyDispatchTrigger: Long = 0
   private var nHandledPlanEnergyDispatchTrigger: Int = 0
 
+  protected val chargingEndTimeInSeconds: Map[Id[Person], Int] =
+    beamServices.matsimServices.getScenario.getPopulation.getPersons.asScala.map { case (personId, person) =>
+      personId -> (person.getSelectedPlan.getPlanElements.asScala
+        .find(_.isInstanceOf[Activity])
+        .map(_.asInstanceOf[Activity].getEndTime)
+        .getOrElse(0.0) + (24 * 3600.0)).toInt
```
nikolayilyin, line 63: This block is confusing - it will return a map of a person -> the end time of the first activity plus 24 hours.
Maybe it would be better to name it like initialChargingEndTimeInSeconds to emphasize that it contains the end time of the first activity, but why it should be increased by 24 hours?

src/main/scala/beam/agentsim/infrastructure/ScaleUpCharging.scala
```diff
@@ -45,14 +43,30 @@ trait ScaleUpCharging extends {
...
+//    )
+//    getBeamServices.matsimServices.getScenario.getHouseholds.getHouseholds
+//      .values()
+//      .asScala
+//      .flatMap(_.getVehicleIds.asScala.map { vehicleId =>
+//        allVehicles = allVehicles - BeamVehicle.createId(vehicleId)
+//      })
+//    allVehicles
+//  }
+
```
nikolayilyin: Let's remove this comment or leave a comment explanation why we need it here

src/test/scala/beam/agentsim/agents/PersonWithPersonalVehiclePlanSpec.scala
```diff
@@ -680,6 +680,8 @@ class PersonWithPersonalVehiclePlanSpec
       personEntersVehicleEvents.expectMsgType[PersonEntersVehicleEvent]
 
       expectMsgType[CompletionNotice]
+
+      // TODO: Testing last resort vehicle creation
```
nikolayilyin, line 684: Is there an issue for that?
If not - please, create one.

nikolayilyin (changes requested):
Mostly nitpicks

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
haitamlaarabi: What do you mean by that?

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
haitamlaarabi: that was a mistake

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
haitamlaarabi: To stop charging in the next day of the simulation after 24h

src/test/scala/beam/agentsim/agents/PersonWithPersonalVehiclePlanSpec.scala
haitamlaarabi: This is related to another PR. I think it should be separate from this PR

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
nikolayilyin: Once the flag set to True it will always be True, I did not see any way to set the flag back to False.
Which means this agent will always reset current tour mode - is this behavior expected?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
nikolayilyin: Oh, that makes sense, ok

nikolayilyin (approved):
Apart from the behavior around the 'householdVehiclesWereNotAvailable' flag there is nothing else I can help with - approved

haitamlaarabi:
Test now!

haitamlaarabi:
Test!

haitamlaarabi:
Test!

PR #3580, opened 2022-08-05
Add buffer time for access to transit

zneedell:
We have very many drive transit trips end up needing to replan due to a missed pickup (it can be ~50%). I think this is because R5 only gives a ~60 sec buffer between the drive access and first transit legs when determining the trip departure time, so when the agent needs to find a parking place or runs into even moderately unexpected traffic they end up arriving late and potentially needing to wait a long time or take another mode. I think behaviorally it's pretty realistic to need to build in a buffer time for drive transit trips (which also serve to make drive transit trips less attractive than walk transit trips that are otherwise similar).

This PR introduces a new parameters allowing you to set a schedule buffer time for each access mode 

>     accessBufferTimeSeconds {
>        bike = "int | 60"
>        bike_rent = "int | 180"
>        walk = "int | 0"
>        car = "int | 300"
>    }

that is applied on top of the 61 second buffer that is always added to account for variable transit headways. also note that bike_rent may not work

Partially addressed #3574 

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3580)


zneedell:
Test!

zneedell:
Test!

nikolayilyin (approved):
lgtm

PR #3588, opened 2022-08-15
Update users.rst

norberto108:
The current **gradle classes** command trigger some errors, **gradlew classes** works without trigger an error

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3588)


nikolayilyin:
The base branch for changes is wrong - we need to use 'develop' instead of 'master'

PR #3590, opened 2022-08-16
update config template file path in docs

alvinlee001:
this is my first commit @heyrutvik , @nikolayilyin is it the correct way to create a pull request?

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3590)


nikolayilyin (approved):
Somehow my changes to src/main/scala/beam/sim/RunBeam.scala got their way here, probably because of 'git rebase' or something like that. 
Otherwise - things are fine.

PR #3593, opened 2022-08-19
added runtime validation (where household and plans are required in t…

alvinlee001:
added runtime validation (where household and plans are required in the old format) for legacy-beam-scenario-specific runs, while making these params optional in matsim.conf, so that we can reduce verbosity of config files

1) Scenario 1
- When: beam.conf or other legacy beam config not include household or plans
- Expected result: run fails, letting user know what config params is missing. Simulation did not proceed.

2) Scenario 2
- When: urbanism-v2/conf or other new beam config not include household or plans
- Expected result: run goes without error, simulation runs smoothly with expected result.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3593)


nikolayilyin:
test!

nikolayilyin (approved):
lgtm

PR #3595, opened 2022-08-22
Remove obsolete maven repositories

alex-vv:
central.maven.org doesn't exist for a long time

nexus.onebusaway.org seems to be not used based on a git history and now causing exception when resolving mockito dependency.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3595)


nikolayilyin (approved):
lgtm

PR #3596, opened 2022-08-22
Add an option to specify logback config for running tests

alex-vv:
It is useful to be able to optionally see detailed logs when running tests. I am adding an option to specify `-PlogbackCfg=logback.xml`  which enables logging configuration for tests. If no option specified it behaves as before.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3596)


nikolayilyin (approved):
lgtm

PR #3600, opened 2022-08-24
Fixes to tour mode choice strategy

zneedell:
Creating a new PR here to make it a little easier to identify the fixes before merging everything into develop

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3600)


src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1125,10 +1125,29 @@ trait ChoosesMode {
           if (isFirstOrLastTrip) WALK_BASED.allowedBeamModesForFirstAndLastLeg
           else WALK_BASED.allowedBeamModes
         walkBasedModes ++ enabledModes
+      case Some(BIKE_BASED) =>
+        if (BIKE_BASED.vehicleSharedOrNot(availablePersonalStreetVehicles)) {Seq[BeamMode]()}
+        else BIKE_BASED.allowedBeamModes
```
zneedell: So let's say the person has access to both a shared bike and a household bike -- if they are on a bike based tour we'd still want them to have access to the BIKE mode, but we'd want them to only be able to use the personal bike for that bike trip. Which makes me think that we might not want to make the change here but instead below

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1125,10 +1125,29 @@ trait ChoosesMode {
...
       case Some(tourMode) => tourMode.allowedBeamModes
       case None           => BeamMode.allModes
     })
   }
+//  def getAvailableVehiclesGivenTourMode(
```
zneedell: So the idea for this function would be a filter that depends on the current tour mode:

- **WALK_BASED** : filter out all household vehicles unless we're on the first or last trip (because then they can use the household vehicle for drive/bike transit)
- **BIKE_BASED** : filter out all shared bikes
- **CAR_BASED** : filter out all shared cars

Xuan-1998: Agree!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
Xuan-1998: True

zneedell: So to clarify -- we want to return an empty sequence of possible modes only if all available bikes are shared (rather than if any are)

Xuan-1998: Yes, I agree!

src/main/scala/beam/router/Modes.scala
```diff
@@ -261,6 +262,12 @@ object TourModes {
 
     import BeamTourMode._
 
+    def vehicleSharedOrNot(vehicles: Vector[VehicleOrToken]): Boolean = {
```
zneedell: There's some scala magic that could make this neater -- something like `vehicles.any(_.vehicle.isSharedVehicle)` I think. But either way, I think we should either use this function to filter the allowed beam modes here or just move this function over to ChoosesMode where it's used, probably no reason for it to be in the `Modes` class otherwise

Xuan-1998: Interesting! Thank you for letting me know!

zneedell: actually I think it's `forall` rather than `any`

Xuan-1998:
@zneedell Please review the latest changes. I'll clean the code up when we agree with the logic

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -385,9 +386,17 @@ trait ChoosesMode {
...
             requestId = None
           }
+          var availableVehicles = newlyAvailableBeamVehicles.map(_.streetVehicle) :+ bodyStreetVehicle
+          personData.currentTourMode match {
+
+            case Some(BIKE_BASED) =>  availableVehicles = newlyAvailableBeamVehicles.filterNot(v => BeamVehicle.isSharedVehicle(v.id)).map(_.streetVehicle) :+ bodyStreetVehicle
+            case Some(WALK_BASED) => availableVehicles = availableVehicles
+            case Some(CAR_BASED) => availableVehicles = newlyAvailableBeamVehicles.filterNot(v => BeamVehicle.isSharedVehicle(v.id)).map(_.streetVehicle) :+ bodyStreetVehicle
+          }
+
```
zneedell, line 397: Logic here looks good to me. Here's an attempt and putting it into cleaner Scala:

      val availableStreetVehiclesGivenTourMode = newlyAvailableBeamVehicles.map { vehicleOrToken =>
        chosenCurrentTourMode match {
          case Some(BIKE_BASED) if !vehicleOrToken.vehicle.isSharedVehicle => vehicleOrToken.streetVehicle
          case Some(CAR_BASED) if !vehicleOrToken.vehicle.isSharedVehicle  => vehicleOrToken.streetVehicle
          case _                                                           => vehicleOrToken.streetVehicle
        }
      } :+ bodyStreetVehicle

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -554,6 +555,9 @@ object BeamVehicle {
   def isSharedTeleportationVehicle(vehicleId: Id[BeamVehicle]): Boolean = {
     vehicleId.toString.startsWith(idPrefixSharedTeleportationVehicle)
   }
+  def isSharedVehicle(vehicleId: Id[BeamVehicle]): Boolean = {
+    vehicleId.toString.startsWith(idPrefixSharedVehicle)
+  }
```
zneedell: If we use `BeamVehicle.isSharedVehicle` as suggested above we won't need this any more

src/main/scala/beam/router/Modes.scala
zneedell: Here's a suggested way of doing all the filtering within `BeamTourMode`

    def allowedBeamModesGivenAvailableVehicles(
      vehicles: Vector[VehicleOrToken],
      firstOrLastLeg: Boolean
    ): Seq[BeamMode] = {
      val relevantModes = if (firstOrLastLeg) { allowedBeamModesForFirstAndLastLeg }
      else allowedBeamModes
      relevantModes.filter { mode =>
        vehicles.exists { vehOrToken =>
          !vehOrToken.vehicle.isSharedVehicle && (vehOrToken.streetVehicle.mode == mode)
        }
      }
    }

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1125,10 +1134,29 @@ trait ChoosesMode {
           if (isFirstOrLastTrip) WALK_BASED.allowedBeamModesForFirstAndLastLeg
           else WALK_BASED.allowedBeamModes
         walkBasedModes ++ enabledModes
+      case Some(BIKE_BASED) =>
+        if (BIKE_BASED.vehicleSharedOrNot(availablePersonalStreetVehicles)) {Seq[BeamMode]()}
+        else BIKE_BASED.allowedBeamModes
       case Some(tourMode) => tourMode.allowedBeamModes
```
zneedell: With the change below in `Modes` we could just replace this line with

      case Some(tourMode) => tourMode.allowedBeamModesGivenAvailableVehicles(availablePersonalStreetVehicles, isFirstOrLastTrip)

src/main/scala/beam/router/Modes.scala
zneedell: Actually scratch that, that won't keep the HOV modes for car. This would work better:

      if (
        vehicles.exists(vehOrToken =>
          !vehOrToken.vehicle.isSharedVehicle && vehOrToken.streetVehicle.mode.in(relevantModes)
        )
      ) { relevantModes }
      else { Seq.empty[BeamMode] }

Xuan-1998: addressed!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
Xuan-1998: Awesome! This is brilliant!

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
Xuan-1998: true, deleted

Xuan-1998:
@zneedell This version addressed your suggestions and merged origin/develop.

zneedell:
Lgtm!

PR #3603, opened 2022-08-26
Reads from csv file, skips vehicleAdjustment(No-Op), generates beam v…

### LINKED ISSUE #3179: Ability to load vehicles from file for urbansim inputs
Eventually we're going to want to assign individual vehicle types to households using the vehicles.csv input rather than probabilistically using vehicleTypes.csv. I'd imagine the best way to do it is just to add a householdId column to vehicles.csv and then assign each vehicle to a household that way.

alvinlee001:
…ehicles from vehicleInfo and binds to household

Able to run with:
gradle :run -PappArgs="['--config', 'test/input/beamville/beam-urbansimv2-static.conf']"

Config difference:
beam.agentsim.agents.vehicles.vehiclesFilePath = ${beam.inputDirectory}"/vehicles-test.csv"
beam.agentsim.agents.vehicles.vehicleAdjustmentMethod = "STATIC_FROM_FILE"

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3603)


src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -218,6 +230,20 @@ class UrbanSimScenarioLoader(
...
+        for {
+          vehicleInfo <- vehicles
+          vehicle = buildBeamVehicle(beamScenario.vehicleTypes, vehicleInfo, rand.nextInt)
+        } {
+          beamScenario.privateVehicles.put(vehicle.id, vehicle)
+          vehicleIds.add(vehicle.id)
+          vehicleCounter = vehicleCounter + 1
+        }
+      }
+
```
nikolayilyin: My suggestion to make things simpler is following:
1. create an implementation of VehiclesAdjustment - FromFileVehiclesAdjustment and read vehicle types once during initialization of the vehicle adjustment object (reading a file for each household is not a good thing - the file might be big and households plenty)
2. add an optional parameter into VehiclesAdjustment.sampleVehicleTypesForHousehold - householdId
3. for FromFileVehicleAdjustment use the householdId to return appropriate vehicle types, for the rest of vehicle adjustment classes - ignore householdId

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -229,6 +255,34 @@ class UrbanSimScenarioLoader(
...
+      VehicleManager.createOrGetReservedFor(info.householdId, VehicleManager.TypeEnum.Household).managerId
+    val powerTrain = new Powertrain(beamVehicleType.primaryFuelConsumptionInJoulePerMeter)
+    new BeamVehicle(
+      beamVehicleId,
+      powerTrain,
+      beamVehicleType,
+      new AtomicReference(vehicleManagerId),
+      randomSeed = randomSeed
+    )
+  }
```
nikolayilyin: Using the suggested approach we might be able to get rid of additional code here

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -57,6 +61,14 @@ class UrbanSimScenarioLoader(
...
     clear()
 
+//    for {
+//      vehicleInfo <- vehicles
+//      vehicle = buildBeamVehicle(beamScenario.vehicleTypes, vehicleInfo, rand.nextInt)
+//    } {
+//      beamScenario.privateVehicles.put(vehicle.id, vehicle)
+//      vehicleInfo.initialSoc.foreach(beamScenario.privateVehicleInitialSoc.put(vehicle.id, _))
+//    }
+
```
nikolayilyin: Comments like these are not very helpful in future, it is usually better to remove them

test/input/beamville/beam-urbansimv2-static.conf
```diff
@@ -0,0 +1,375 @@
...
+beam.agentsim.agentSampleSizeAsFractionOfPopulation = 1.0
+beam.agentsim.firstIteration = 0
+beam.agentsim.lastIteration = 0
+
+beam.agentsim.thresholdForWalkingInMeters = 100
+beam.agentsim.thresholdForMakingParkingChoiceInMeters = 100
+beam.agentsim.schedulerParallelismWindow = 30
+beam.agentsim.timeBinSize = 3600
+beam.agentsim.startTime = "00:00:00"
+beam.agentsim.endTime = "30:00:00"
```
nikolayilyin: In order to create a config file that has only few config values different there is no need to copy all the content, you migh just include the base config file in the beginning and then put into config file only the config keys you want to change.
For example - test/input/beamville/beam-csv.conf

nikolayilyin (changes requested):
Let's try to change the implementation to make it simpler.
Also, after the changes are made - let's run gradle command 'fmt', it will format the source code files.

src/main/scala/beam/sim/vehicles/FromFileVehiclesAdjustment.scala
```diff
@@ -0,0 +1,55 @@
...
+      beamScenario.beamConfig.beam.agentsim.agents.vehicles.vehiclesFilePath
+    )
+  }
+}
+
+object FromFileVehiclesAdjustment {
+
+  def apply(beamScenario: BeamScenario) = new FromFileVehiclesAdjustment(beamScenario) {
+    vehicles = readVechiclesFromFile()
+    vehicleTypesMap = beamScenario.vehicleTypes.map(i => i._1 -> i._2).toMap
```
nikolayilyin: I would put these rows as first rows (14-15) in the class and replace 'var' with 'val' to emphasize these values are not to be replaced.

nikolayilyin (changes requested):
Let's change the new vehicle adjustment class a bit, the rest looks good to me

src/test/scala/beam/utils/scenario/UrbanSimScenarioLoaderTest.scala
```diff
@@ -90,11 +98,45 @@ class UrbanSimScenarioLoaderTest extends AsyncWordSpec with Matchers with Before
...
 
+  def staticVehiclesConfig: com.typesafe.config.Config =
+    ConfigFactory
+      .parseString(s"""
+                      |beam.agentsim.simulationName = "beamville-urbansimv2_static"
+                      |beam.agentsim.agents.vehicles.vehiclesFilePath = $${beam.inputDirectory}"/vehicles-test.csv"
+                      |beam.agentsim.agents.vehicles.vehicleAdjustmentMethod = "STATIC_FROM_FILE"
+                    """.stripMargin)
+      .withFallback(testConfig("test/input/beamville/beam-urbansimv2.conf"))
+      .resolve()
```
nikolayilyin: let's put this config initialization inside the test, no need to create external variable for a test

src/test/scala/beam/utils/scenario/UrbanSimScenarioLoaderTest.scala
```diff
@@ -90,11 +98,45 @@ class UrbanSimScenarioLoaderTest extends AsyncWordSpec with Matchers with Before
...
+      val staticScenarioSource = buildUrbansimV2ScenarioSource(staticGeoUtils, staticVehicleBeamConfig);
+      val urbanSimScenario = new UrbanSimScenarioLoader(emptyScenario, staticBeamScenario, staticScenarioSource, geoUtils)
+      urbanSimScenario.loadScenario()
+
+      val vehicleIds = emptyScenario.getHouseholds.getHouseholds.get(Id.create(2, classOf[Household])).getVehicleIds
+      val vehicleMap = staticBeamScenario.privateVehicles.toMap
+      val vehiclesBelongToSpecificHousehold = vehicleIds.map(x => Id.create(x, classOf[BeamVehicle])).flatMap(x => vehicleMap get x);
+      vehiclesBelongToSpecificHousehold.count(x => x.beamVehicleType.id.compareTo(
+        Id.create("Bicycle", classOf[BeamVehicleType])
+      ) == 0) shouldBe 7
```
nikolayilyin: Let's use `equal` instead of `compareTo`

nikolayilyin (changes requested):
Let's change the test a bit, the rest look good

nikolayilyin (approved):
lgtm

nikolayilyin:
test!

nikolayilyin:
test!

PR #3606, opened 2022-08-29
Fix NPE when loading RideHail skims

alex-vv:
In case we get `UnmatchedRideHailRequestSkimmerEvent` for RideHail skims then  `waitTime` and `costPerMile` values are `Double.NaN` and they are written to skim csv file as empty values. When reading it back it causes NPE. The workaround is to assign `Double.NaN` in this case.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3606)


nikolayilyin (approved):
lgtm

PR #3620, opened 2022-09-09
Fixes issue #3457 of duplicate column names in passengerPerTripSubway.csv when the data contains only zeroes.

### LINKED ISSUE #3457: passengerPerTripSubway.csv contains duplicate columns
When the data for `0.passengerPerTripSubway.csv` is empty (meaning only zero numbers), then there are duplicate columns (header=1-0) in the file.

image

We should fix this and only print 1 column.

HoneyTauOverTwo:
Assuming that this is the [desired output](https://github.com/LBNL-UCB-STI/beam/issues/3457#issuecomment-1242211276).

IsDesirableOutput

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3620)


HoneyTauOverTwo:
As requested by @nikolayilyin, made changes to omit the columns that would be filled with zeroes.

This is what the empty data results looks like:
outputOmmited0

And I also modified the input data [Map<Integer, Map<Integer, Integer>>](https://github.com/LBNL-UCB-STI/beam/blob/develop/src/main/java/beam/analysis/plots/passengerpertrip/GenericPassengerPerTrip.java#L111) on debug to test the intermediate cases (with 1 - 3 non-null columns), sample result for two:

outputOmmited1

This code only affects the output for bus, subway and bike (at least for the beamville data set). The other transport modes (car and ride hail) remain unchanged.

HoneyTauOverTwo:
Removed columns filled with zeroes from passengerPerTripCar as well.

passengerPerTripCar

I figured the way I implemented for buses and subways can possibly have zeroed columns. I am only omitting the last X zeroed columns.

nikolayilyin:
test!

src/test/scala/beam/sim/output/IterationsPassengerPerTripTests.scala
```diff
@@ -0,0 +1,304 @@
...
+
+import scala.collection.mutable.ListBuffer
+import scala.util.control.Breaks.{break, breakable}
+
+class IterationsPassengerPerTripTests
+  extends AnyWordSpecLike
+    with Matchers
+    with BeamHelper
+    with BeforeAndAfterAllConfigMap
+    with Retries {
```
nikolayilyin: it seems there is no need in most of these extensions

nikolayilyin (approved):
looks good

nikolayilyin:
test!

src/test/scala/beam/sim/output/IterationsPassengerPerTripTests.scala
HoneyTauOverTwo: Oh, certainly... I copied BeamWarmStartRunSpec.scala and missed removing this.

nikolayilyin:
test!

src/test/scala/beam/utils/SnapCoordinateSpec.scala
```diff
@@ -67,7 +69,8 @@ class SnapCoordinateSpec extends AnyWordSpec with Matchers with BeamHelper {
   }
 
   "scenario plan" should {
-    "contain all valid locations" in {
+    // TODO: Investigate, There might be an issue hidden in this flaky test. Retryable tag was added to pass an unrelated PR
```
HoneyTauOverTwo, line 72: I added this comment to show this was not investigated. Just added the Retryable tag. I found two flaky tests on this file.

HoneyTauOverTwo:
test!

src/test/scala/beam/utils/SnapCoordinateSpec.scala
nikolayilyin: I think all the tests marked as Retryable deserves some investigation, no need to mark them individually

HoneyTauOverTwo:
test!

nikolayilyin (approved):
lgtm

PR #3625, opened 2022-09-21
Multiple RideHailManagers

### LINKED ISSUE #3616: Multiple RideHailManagers
For implementing it we could do the following:

1. Move the `rideHail` params to a list of objects (similar to what we have for Shared Vehicle Fleets).
2. Create a `RideHailManager` for each object of that list.
3. `ChargingNetworkManager` keeps separate charging network for each RHM. These networks are chosen by the `reservedFor` field from the inquiry.

It would require changing every config file we use. We can do it in several steps. As someone runs a config then them modify it to match the new config template.

**Handling RH requests**
We can create a new Actor (i.e. `RideHailMaster` maybe a better name). Which the agents sends the requests. That `RideHailMaster`  sends the requests to the actual RHMs. Then it chooses the best proposal from the responses and sends it back to the agent. In this case we avoid modifying the `PersonAgent` (ModeChoice).

dimaopen:
Now one can configure multiple rideHailManagers in the config parameter `beam.agentsim.agents.rideHail.managers`.

For backward compatibiltity of the configs a method `BeamHelper#updateConfigToCurrentVersion` is added. It also allow to run experiments with modifying RideHailManager parameters.

Introduced a new Actor: `RideHailMaster` which receives RH inquires and sends them to each RHM. When all the answer are received the best one (by the lowest price) is chosen and sent back to the requester.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3625)


dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailMaster.scala
```diff
@@ -0,0 +1,157 @@
...
+          logger.error(s"Cannot find originalResponse for $reserveRide")
+          sender() ! RideHailResponse.dummyWithError(UnknownInquiryIdError)
+      }
+
+    case Finish =>
+      rideHailResponseCache.clear()
+      rideHailManagers.values.foreach(_ ! Finish)
+
+    case _: Terminated =>
+      if (context.children.isEmpty) context.stop(self)
```
nikolayilyin, line 126: What if there are children?

dimaopen:
Test!

src/main/scala/beam/router/skim/ActivitySimSkims.scala
```diff
@@ -11,55 +11,4 @@ import beam.sim.BeamScenario
...
-      0,
-      0,
-      0,
-      0,
-      0,
-      0
-    )
-  }
-}
+case class ActivitySimSkims(beamConfig: BeamConfig, beamScenario: BeamScenario) extends AbstractSkimmerReadOnly
```
nikolayilyin, line 14: Why was this removed?

nikolayilyin (approved):

src/main/scala/beam/router/skim/ActivitySimSkims.scala
dimaopen: It is not used.

src/main/scala/beam/agentsim/agents/ridehail/RideHailMaster.scala
dimaopen: Then it's going to wait until all children stop themselves.

src/main/scala/beam/router/skim/ActivitySimSkims.scala
dimaopen: ActivitySimSkims are intended only for output (not for usage inside beam).

dimaopen:
Test!

dimaopen:
Test!

haitamlaarabi (approved):
> Reviewed 58 of 77 files at r1, 30 of 33 files at r2, 1 of 1 files at r3, 1 of 1 files at r4, 6 of 9 files at r5.
> *[Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3625)* status: 58 of 109 files reviewed, 2 unresolved discussions (waiting on @alex-vv, @JustinPihony, and @nikolayilyin)

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

dimaopen:
Test!

PR #3629, opened 2022-09-27
Memory usage optimization

### LINKED ISSUE #3609: Profiling of BEAM simulation - RAM
We need to do a general profiling of RAM usage:
1. Check if there are any memory leaks
2. Figure out which code areas are the main source of memory usage
3. Figure out which simulation steps are the main source of memory usage

In order to do so one needs to do profiling of small simulation locally (beamville, sf-light) and then work on profiling of big runs.
There is a tool to check if there are memory leaks is in beam, probably it is required to be upgraded.
The config key to turn the tool on: beam.debug.vmInformation.createGCClassHistogram

alex-vv:
This PR improves BEAM memory usage by around 20%. There are some unused variables dropped and some collections were replaced by `Array` in `BeamPath` and `BeamPlan`

Fixes #3609

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3629)


src/main/scala/beam/agentsim/agents/planning/BeamPlan.scala
```diff
@@ -92,43 +92,37 @@ class BeamPlan extends Plan {
...
-    if (nextTour.trips.nonEmpty) tours = tours :+ nextTour
+    if (nextTour.trips.nonEmpty) toursBuilder += nextTour
+    tours = toursBuilder.result()
     indexBeamPlan()
-    actsLegs.foreach {
-      case l: Leg =>
-        putStrategy(actsLegToTrip(l), ModeChoiceStrategy(BeamMode.fromString(l.getMode)))
-      case _ =>
-    }
   }
```
nikolayilyin, line 124: Why this part was removed?

nikolayilyin (approved):
Apart from piece of code being removed, looks good to me

src/main/scala/beam/agentsim/agents/planning/BeamPlan.scala
alex-vv: This was about updating `strategies` map in BeamPlan. I could not find any usage of this map. 

Previously it was removed from `ChoosesMode` here https://github.com/LBNL-UCB-STI/beam/commit/343672c43ef62281d842fd041c81e6983f126d15 and then from `PersonAgent` here https://github.com/LBNL-UCB-STI/beam/commit/d99b8cf09729e9c6d08a05ce86c118a385861b18. 
 
So we only inserting values there and don't use them. That is why I dropped this variable and related code for updates. Actually it looks that `beam.agentsim.agents.planning.Strategy` could be dropped as well after that.

nikolayilyin: It seems it is used only in test, looks like some forgotten old code, ok then, let's remove it

alex-vv:
test!

nikolayilyin:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

nikolayilyin (approved):
lgtm

PR #3635, opened 2022-10-04
Copy cloud-init if BEAM was unable to start on AWS s3

### LINKED ISSUE #3585: AWS deploy: save cloud-init if beam was unable to start.
If BEAM was unable to start and then the AWS instance was terminated we are loosing valuable information about the reasons behind this failure.
To solve this we need to upload to AWS bucket the cloud-init log file if BEAM output does not exist.
The name of the file should include the config file name and the path to the config file name and current date and time, for example: "cloud-init-output__08_13_2022_00:00:52__production_newyork_new-york-may2020-calibration-13.conf".
The file should be uploaded into the root of the output bucket and in the simulations spreadsheet the 'S3 Url' column should have a link to this file (or to the whole output folder as it is by-default if the BEAM output folder exist)

nikolayilyin:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3635)


alex-vv (approved):
LGTM

PR #3645, opened 2022-10-22
Lambda hotfix: a fix for 'not a string value'

nikolayilyin:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3645)

### DIFF
--- a/aws/src/main/python/beam_lambda/lambda_function.py
+++ b/aws/src/main/python/beam_lambda/lambda_function.py
@@ -95,6 +95,10 @@ EXPERIMENT_DEFAULT = 'test/input/beamville/calibration/experiments.yml'
 
 CONFIG_DEFAULT = 'production/application-sfbay/base.conf'
 
+DEFAULT_STUCK_GUARD_MIN_CPU_USAGE = "5"
+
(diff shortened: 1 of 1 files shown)


nikolayilyin:
test!

aws/src/main/python/beam_lambda/lambda_function.py
```diff
@@ -95,6 +95,10 @@
 
 CONFIG_DEFAULT = 'production/application-sfbay/base.conf'
 
+DEFAULT_STUCK_GUARD_MIN_CPU_USAGE = "5"
+
+DEFAULT_STUCK_GUARD_MAX_INACTIVE_TIME_INTERVAL = "10"
```
dimaopen, line 100: These default values are not the same as in the `aws/build.gradle`. Which values are more appropriate? CPU usage is measured in percent or fraction of 1.0?

nikolayilyin: These are just more safe values in case the branch did not get an update.

dimaopen (approved):

PR #3648, opened 2022-10-27
Do not generate duplicate values for WALK skims

### LINKED ISSUE #3239: Improve WALK skims generation for activity sim skims
Currently BackgroundSkimsCreatorApp generates duplicate rows for WALK skims because these are the same for each time period. We can optimize it and only generate WALK skims for one time period (EA).

alex-vv:
Fixes #3239

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3648)

### DIFF
--- a/src/main/scala/beam/router/skim/ActivitySimSkimmer.scala
+++ b/src/main/scala/beam/router/skim/ActivitySimSkimmer.scala
@@ -224,14 +224,18 @@ class ActivitySimSkimmer @Inject() (matsimServices: MatsimServices, beamScenario
     destination: GeoUnit,
     pathType: ActivitySimPathType
   ): Option[ExcerptData] = {
-    val individualSkims = timeBin.hours.flatMap { hour =>
-      getCurrentSkimValue(ActivitySimSkimmerKey(hour, pathType, origin.id, destination.id))
-        .map(_.asInstanceOf[ActivitySimSkimmerInternal])
-    }
-    if (individualSkims.isEmpty) {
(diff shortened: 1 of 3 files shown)


src/main/scala/beam/router/skim/ActivitySimSkimmer.scala
```diff
@@ -224,14 +224,18 @@ class ActivitySimSkimmer @Inject() (matsimServices: MatsimServices, beamScenario
     destination: GeoUnit,
     pathType: ActivitySimPathType
   ): Option[ExcerptData] = {
-    val individualSkims = timeBin.hours.flatMap { hour =>
-      getCurrentSkimValue(ActivitySimSkimmerKey(hour, pathType, origin.id, destination.id))
-        .map(_.asInstanceOf[ActivitySimSkimmerInternal])
-    }
-    if (individualSkims.isEmpty) {
+    if (pathType == ActivitySimPathType.WALK && timeBin != ActivitySimTimeBin.EARLY_AM) {
```
nikolayilyin, line 227: Is it possible to start generation of skims without 'EARLY_AM' time and completely miss the walk skims generation? 
If it is possible, maybe it would be better to keep a map of OD - walk skims and generate only if a value in a map is missing?

alex-vv: No, this code for ActivitySim skims is always used together with `ActivitySimTimeBin.values` which lists all possible time values.

nikolayilyin (approved):
lgtm

PR #3659, opened 2022-11-09
Fixed lots of similar log messages

dimaopen:
We still indication of the errors in the log

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3659)


nikolayilyin (approved):
I still don't like the approach of hiding the errors, but, let's gamble

PR #3675, opened 2022-11-17
Edward/#3612 aws storage analysis

edwazhao:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3675)

### DIFF
--- a/src/main/bash/analysis_storage_of_s3_bucket.sh
+++ b/src/main/bash/analysis_storage_of_s3_bucket.sh
@@ -1,0 +1,107 @@
+#!/bin/bash
+# Please use command below to start the shell script
+#   bash analysis_s3_bucket.sh beam > result.txt
+# The first parameter beam is the profile of aws configured to beam aws rescource,
+#   for example: aws s3 ls --profile=beam, then please pass beam as the first parameter to the shell
+# When set directory_depth to 3, the result would be:
+#beam_outputs/
+#    directoryA/ 10 GiB
+#        directoryA-1/   8 Gib
+#        FileA   1 Gib
+#        FileB   145 Mib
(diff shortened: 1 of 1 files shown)


src/main/bash/analysis_storage_of_s3_bucket.sh
```diff
@@ -0,0 +1,97 @@
+#!/bin/bash
+# Please use command below to start the shell script
+#   bash analysis_s3_bucket.sh beam beam-outputs/output/ > result.txt
+# The first parameter beam is the profile of aws configured to beam aws resource,
+#   for example: aws s3 ls --profile=beam, then please pass beam as the first parameter to the shell
+# The second parameter path is the profile of aws configured to beam aws resource,
+#   for example: aws s3 ls --profile=beam, then please pass beam as the first parameter to the shell
```
nikolayilyin, line 7: Seems to be a copy of a description of the first parameter, plus, let's add that this is an optional parameter and it's default value

src/main/bash/analysis_storage_of_s3_bucket.sh
```diff
@@ -0,0 +1,97 @@
...
+folder_name=${2:-'beam-outputs/'}
+if [ ${folder_name: -1} != "/" ]; then
+    folder_name=$folder_name'/'
+fi
+echo $folder_name
+
+alias aws_beam='aws --profile=$beam_profile'
+source /etc/bash.bashrc
+
+directory_depth=3
```
nikolayilyin, line 39: Let's add the depth as an input parameter

nikolayilyin (approved):
apart from few nitpicks looks good to me

PR #3676, opened 2022-11-17
Fixes the tick and trigger issue on SingleModeSpec test

HoneyTauOverTwo:
This issue happens because the test "let everybody take drive_transit when their plan says so" leaves some electric vehicles plugged into the charging station, then these vehicles are in an unexpected state when the next test "let everybody drive when their plan says so" starts.

Running ``beamScenario.privateVehicles.foreach(_._2.resetState())`` right at the begining of the aforementioned second test fixes the issue. I will be adding this code to SimRunnerForTest.beforeEach() and comparing the CI output before and after the change.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3676)

### DIFF
--- a/src/test/scala/beam/utils/SimRunnerForTest.scala
+++ b/src/test/scala/beam/utils/SimRunnerForTest.scala
@@ -1,13 +1,10 @@
 package beam.utils
 
-import java.io.File
-
 import akka.actor.ActorSystem
 import beam.agentsim.agents.modalbehaviors.ModeChoiceCalculator
 import beam.agentsim.events.eventbuilder.{ComplexEventBuilder, EventBuilderActor}
-import beam.api.{BeamCustomizationAPI, DefaultAPIImplementation}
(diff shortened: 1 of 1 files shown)


HoneyTauOverTwo:
test!

HoneyTauOverTwo:
test!

HoneyTauOverTwo:
Turns out, differently from what we thought, there is another test that was throwing this issue: ``BikeTransitModeSpec`` as can be seen here:
[CI output before change.txt](https://github.com/LBNL-UCB-STI/beam/files/10035455/CI.output.before.change.txt)

The output after these changes does not contain any instance of the issue:
[CI output after change.txt](https://github.com/LBNL-UCB-STI/beam/files/10035454/CI.output.after.change.txt)

I looked through all the error messages reported on "CI output after change" and, from what I can tell, there is no new kind of error being thrown when compared to "CI output before change".

Xuan-1998 (approved):
LGTM

PR #3681, opened 2022-11-21
Fix boxing issue

JustinPihony:
Anonymous functions with doubles create unnecessary boxing, so turning this into an explicit type

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3681)


Xuan-1998 (approved):
LGTM

PR #3683, opened 2022-11-23
Deployment spreadsheet improvements

### LINKED ISSUE #3387: Spreadsheet with deploy status and run data - bringing order.
We have more than 10k rows in the spreadsheet and it takes some time to scroll to the bottom of it.
I see two possible solutions for it:
1. reverse the order of existing rows and insert new rows at the beginning of the spreadsheet
2. move older simulations in the separate tab - "2019 - 2020"

 think we should try both solutions at the same time, but if the first approach is not possible, then the second will already make life easier.

alex-vv:
* There are two sheets now: current and archive
 * Archive sheet behaves as before
 * In the current sheet there are only a fixed number of recent runs
 * New run is added on top of the current sheet, the oldest one is deleted
 * New column is added in current sheet: 'Time Completed'
 * Once run is completed the relevant row in the current sheet is updated and completion timestamp is put in 'Time Completed' column

 Also did some code cleanup & dependencies updates.

 Some lambda parameters are moved to environment variables, in particular `event_id` is extracted to environment variable, so it is removed from `beam_lambda` as well.

 Fixes #3387

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3683)

### DIFF
--- a/aws/src/main/python/beam_lambda/lambda_function.py
+++ b/aws/src/main/python/beam_lambda/lambda_function.py
@@ -258,7 +258,6 @@ runcmd:
   -   start_json=$(printf "{
         \\"command\\":\\"add\\",
         \\"type\\":\\"beam\\",
-        \\"sheet_id\\":\\"$SHEET_ID\\",
         \\"run\\":{
           \\"status\\":\\"Run Started\\",
           \\"name\\":\\"$TITLED\\",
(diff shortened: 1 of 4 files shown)


alex-vv:
test!

alex-vv:
test!

alex-vv:
test!

nikolayilyin (approved):
lgtm

alex-vv:
test!

PR #3685, opened 2022-11-29
Spreadsheet lambda fixes & improvements

alex-vv:
A few follow-up fixes for https://github.com/LBNL-UCB-STI/beam/pull/3683

* Fix issue when S3 Url was not updated for completed runs
* Fix issue when lambda failed to add row in `latest 500` sheet in case of unexpected row count. Now it can automatically resize the grid adding or removing rows as needed.
* Automatically detect sheet id, so `current_sheet_id` env var is not needed
* Add an option for load credentials from local key file to simplify local testing
* Add deployment readme

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3685)

### DIFF
--- a/aws/src/main/python/spreadsheetData/README.md
+++ b/aws/src/main/python/spreadsheetData/README.md
@@ -1,0 +1,21 @@
+# Deployment to AWS
+
+Since this lambda function has external dependencies it has to be deployed in a zip archive.
+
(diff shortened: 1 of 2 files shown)


nikolayilyin (approved):
lgtm

PR #3686, opened 2022-11-29
Another follow-up lambda hotfix

alex-vv:
* Fix an edge case with row count
* Fix doc

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3686)


nikolayilyin (approved):
lgtm

PR #3689, opened 2022-12-03
Don't fail if we're missing a vehicle's fuel type

zneedell:
I was running into an error in the second iteration when an agent has a bike trip in their plan memory, which happens when the R5Wrapper is trying to look up the fuel cost for "food" and it isn't in the input table. I think it's safe to assume that if the fuel cost of a certain type of fuel isn't defined we can just use 0 by default

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3689)


Xuan-1998 (approved):
LGTM

PR #3703, opened 2022-12-23
Fixed Memory Leak that caused keeping EventsManagerImpl for the whole run

### LINKED ISSUE #3698: Memory leak investigation
It looks like there is a memory leak in the branch - freight-develop-merge.
I checked by running a [simulation](https://s3.us-east-2.amazonaws.com/beam-outputs/index.html#output/sfbay/sfbay-freight-calibration-3__2022-12-07_12-03-09_rao) 
Here is a growing amount of RAM from iteration to iteration:
image

Simulation was started at commit fe1904ad0290c746ecaf39b91b1a1fafa8d68986, probably worth checking if the leak is still present in the latest version of the branch.
The config - production/sfbay/freight/sfbay-freight-calibration-3-check-for-memory-leak.conf (data commit - 6b80a3460fdb91c83fa5520197c20fe9622d883c)

dimaopen:
1. Instead of adding `PhysSimEventWriter` which keeps a reference to the EventManager to the matsim controller listeners we add a small object that keeps a reference only to a terminated threadpool.
2. Do not init `PrallelEventManagerImpl` multiple times because it abandons blocked threads that are GC Roots. They keep their frames until the app exits.
3. Added a new parameter `vmInformation.writeHeapDump` in order to write a heap dump at each iteration so that we could debug memory leaks easier.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3703)

### DIFF
--- a/src/main/java/beam/physsim/jdeqsim/AgentSimToPhysSimPlanConverter.java
+++ b/src/main/java/beam/physsim/jdeqsim/AgentSimToPhysSimPlanConverter.java
@@ -96,8 +96,6 @@ public class AgentSimToPhysSimPlanConverter implements BasicEventHandler, Metric
     private final Random rand = MatsimRandom.getRandom();
     private final boolean agentSimPhysSimInterfaceDebuggerEnabled;
 
-    private final List<CompletableFuture> completableFutures = new ArrayList<>();
-
     final Map<String, Boolean> caccVehiclesMap = new TreeMap<>();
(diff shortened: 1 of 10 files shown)


dimaopen:
This is the run name to confirm memory leak is fixed `do/freight-memory-leak-fix_dimaopenATgmail_com`.

nikolayilyin (approved):
lgtm

Issue #3701, opened 2022-12-22
PWD in tests - a possible problem for windows users


nikolayilyin:
In few tests a `$pwd` variable is used and for windows users this variable is not working.
We need to either configure something (windows or IDEA) in order for this variable to work and create a documentation entry about the configuration or, if impossible, we need to get rid of this variable and replace it with something OS-independent.

edwazhao:
Please review https://github.com/LBNL-UCB-STI/beam/pull/3705

PR #3705, opened 2022-12-25
#3701 add hints for scala unit test

edwazhao:
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3705)

https://github.com/LBNL-UCB-STI/beam/issues/3701

$pwd is coming from the top of the Test file
`val pwd: String = System.getenv("PWD")`
So the way to fix the issue is to add the system environment variable PWD, for running BEAM simulation it is mentioned by https://beam.readthedocs.io/en/latest/users.html#running-beam-with-intellij-ide, so I added one more block for unit test, please review the changes.


nikolayilyin (approved):
lgtm

PR #3709, opened 2022-12-28
Enabling Multiple riders for shared bikes

Xuan-1998:
It's important to understand the impacts of bike capacities under the price schema we proposed so we want to do it in the following way:
1. several passengers (depending on the capacity of the shared bike) can access the shared bikes at the same time at the docked stations.
2. they should behave similarly with shared cars or carpooling, and leave the closer passengers at their destinations first and then carry on with other passengers.

- - -
This change is [Reviewable](https://reviewable.io/reviews/lbnl-ucb-sti/beam/3709)

### DIFF
--- a/src/main/scala/beam/agentsim/agents/PersonAgent.scala
+++ b/src/main/scala/beam/agentsim/agents/PersonAgent.scala
@@ -1078,6 +1078,7 @@ class PersonAgent(
       // Declaring a function here because this case is already so convoluted that I require a return
       // statement from within.
       // TODO: Refactor.
+
       def nextState: FSM.State[BeamAgentState, PersonData] = {
         val currentVehicleForNextState =
           if (currentVehicle.isEmpty || currentVehicle.head != nextLeg.beamVehicleId) {
@@ -1101,6 +1102,8 @@ class PersonAgent(
           }
         val legsToInclude = nextLeg +: restOfCurrentTrip.takeWhile(_.beamVehicleId == nextLeg.beamVehicleId)
         val newPassengerSchedule = PassengerSchedule().addLegs(legsToInclude.map(_.beamLeg))
+
+
 
         // Enroute block
         // calculate whether enroute charging required or not.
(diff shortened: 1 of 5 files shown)


src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -967,13 +967,23 @@ class PersonAgent(
     case Event(Boarded(vehicle, _), _: BasePersonData) =>
       beamVehicles.put(vehicle.id, ActualVehicle(vehicle))
       potentiallyChargingBeamVehicles.remove(vehicle.id)
+      // allowing shared bike to be used by multiple agents at the same time
+      if (vehicle.beamVehicleType.vehicleCategory == Bike) {
```
dimaopen, line 971: Shared vehicles could be not only bikes.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -967,13 +967,23 @@ class PersonAgent(
...
+      }
       goto(ProcessingNextLegOrStartActivity)
     case Event(NotAvailable(_), basePersonData: BasePersonData) =>
       log.debug("{} replanning because vehicle not available when trying to board")
       val replanningReason = getReplanningReasonFrom(basePersonData, ReservationErrorCode.ResourceUnavailable.entryName)
       eventsManager.processEvent(
         new ReplanningEvent(_currentTick.get, Id.createPersonId(id), replanningReason)
       )
+      // allowing shared bike to be used by multiple agents at the same time
+      if (basePersonData.currentVehicle.exists(_.toString.contains("shared"))) {
```
dimaopen, line 982: It's better to use `vehicle.isSharedVehicle` to detect the vehicle is shared.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -1121,6 +1132,7 @@ class PersonAgent(
         } else {
           false
         }
+        val isSharedVehicle = vehicle.isSharedVehicle
```
dimaopen, line 1135: Why do you need this variable? Could you just use `vehicle.isSharedVehicle` as they use it below?

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -1132,6 +1144,16 @@ class PersonAgent(
...
         }
 
+        if (isSharedVehicle) {
+          val tick = _currentTick.get
+          val triggerId = _currentTriggerId.get
+          scheduler ! CompletionNotice(
+            triggerId,
+            if (nextLeg.beamLeg.endTime > lastTickOfSimulation) Vector.empty
+            else Vector(ScheduleTrigger(StartLegTrigger(tick, nextLeg.beamLeg), self))
+          )
```
dimaopen, line 1154: Double CompletionNotice leads to errors in the scheduler.

dimaopen (changes requested):
Is it possible to use some configuration parameter to disable this functionality?

dimaopen:
We need to thoroughly test these changes before we merge it into `develop`.

PR #3720, opened 2023-01-30
Ridehail stops

dimaopen:
If a stop file is provided, then RH vehicles pick up/drop off passengers on the stops.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3720)


src/main/scala/beam/router/model/EmbodiedBeamTrip.scala
```diff
@@ -42,6 +42,7 @@ case class EmbodiedBeamTrip(legs: IndexedSeq[EmbodiedBeamLeg], router: Option[St
   lazy val replanningPenalty: Double = legs.map(_.replanningPenalty).sum
 
   val totalTravelTimeInSecs: Int = legs.lastOption.map(_.beamLeg.endTime - legs.head.beamLeg.startTime).getOrElse(0)
+  def totalDistanceInM: Double = legs.view.map(_.beamLeg.travelPath.distanceInM).foldLeft(0.0)(_ + _)
```
JustinPihony, line 45: Can this be a `val`? It seems that legs cannot change?

JustinPihony (approved):

src/main/scala/beam/router/model/EmbodiedBeamTrip.scala
dimaopen: Yes, but in this case it would be calculated every time when an instance is created. And that method is not used every time.

dimaopen:
Test!

dimaopen:
Test!

src/main/scala/beam/router/model/EmbodiedBeamTrip.scala
JustinPihony: How about `lazy val` then?

dimaopen: That might work. But `lazy val`s are not so cheap. They are generally for correctness not for optimization as I understood.

nikolayilyin:
test!

nikolayilyin:
test!

src/main/scala/beam/agentsim/agents/ridehail/RideHailManager.scala
```diff
@@ -1521,7 +1641,7 @@ class RideHailManager(
       if (currentlyProcessingTimeoutTrigger.isEmpty) {
         // We always use the request buffer even if we will process these requests immediately
         rideHailResourceAllocationManager.addRequestToBuffer(request)
-        findAllocationsAndProcess(request.departAt, triggerId)
+        findAllocationsAndProcess(request.requestTime, triggerId) //todome request.requestTime
```
nikolayilyin: please, elaborate about this TODO, or remove it completely

nikolayilyin (approved):
apart from one comment, the rest lgtm

nikolayilyin:
test!

nikolayilyin:
test!

PR #3722, opened 2023-02-02
Freight: using a separate optional vehicle types file

nikolayilyin:


nikolayilyin:
test!

src/main/scala/beam/agentsim/agents/freight/input/FreightReader.scala
```diff
@@ -205,7 +205,7 @@ object FreightReader {
     beamConfig: BeamConfig,
     geoUtils: GeoUtils,
     streetLayer: StreetLayer,
-    network: Network,
+    network: Option[Network],
```
dimaopen, line 208: It seems that network is always provided. Do we really need this Option?

src/main/resources/beam-template.conf
```diff
@@ -273,6 +273,7 @@ beam.agentsim.agents.freight {
   plansFilePath = ${beam.inputDirectory}"/freight/payload-plans.csv"
   toursFilePath = ${beam.inputDirectory}"/freight/freight-tours.csv"
   carriersFilePath = ${beam.inputDirectory}"/freight/freight-carriers.csv"
+  vehicleTypesFilePath = ""
```
dimaopen: Maybe we could use
`#@optional
carrierParkingFilePath = string`
in order to make it more clear that this is optional?

src/main/scala/beam/agentsim/agents/freight/input/FreightReader.scala
nikolayilyin: It is optional on the very bottom of the call tree: 
[val newLocIntUtm = networkMaybe.map(NetworkUtils.getNearestLink(_, locInUtm).getCoord).getOrElse(locInUtm)](https://github.com/LBNL-UCB-STI/beam/blob/791cc59547afdcc831d7a6ea6413f73bba7e614b/src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala#L336)
I think it is better to have optional type through all the way down there.

src/main/resources/beam-template.conf
nikolayilyin: good idea!

nikolayilyin:
test!

src/main/resources/beam-template.conf
dimaopen: If I understand correctly we need to run `generateConfig` task again. It generates an Option for this field. We also need to slightly modify the code (it may get even simpler).

src/main/scala/beam/agentsim/agents/freight/input/GenericFreightReader.scala
```diff
@@ -178,6 +179,13 @@ class GenericFreightReader(
     allPlans: Map[Id[PayloadPlan], PayloadPlan],
     vehicleTypes: Map[Id[BeamVehicleType], BeamVehicleType]
   ): IndexedSeq[FreightCarrier] = {
+    val freightVehicleTypes: Map[Id[BeamVehicleType], BeamVehicleType] = {
+      config.vehicleTypesFilePath match {
+        case Some(filePath) => vehicleTypes ++ readBeamVehicleTypeFile(filePath)
+        case None           => vehicleTypes
+      }
+    }
```
dimaopen, line 185: To improve readability we could remove the outer curly braces. Though it's up to you.

dimaopen (approved):

nikolayilyin:
test!

PR #3725, opened 2023-02-09
Execute BEAM on Google Cloud

### LINKED ISSUE #3652: Amazon vs Google cloud: comparison
There are three parts of the issue:
1. to figure out the cost difference between google cloud and amazon cloud
2. to figure out if it is possible to use google cloud instead of amazon cloud
3. to figure out how much effort it will cost us to move from one cloud to another

dimaopen:
In the project a new module (gcp) is created. It contains cloud functions that deployed to Google Cloud functions.

GCP project is "beam-core"

beam-box images is used to create VMs. It's already has all the required infrastructure to run BEAM (java 8, git, git-lfs)


nikolayilyin (approved):
If we urgently need this functionality in develop, then I think we are good to go.
Otherwise It needs to be merged with aws code: we will benefit from having only one cloud-init script and such, probably even deploy commands might be united by adding a parameter 'cloudPlatform' with values 'amazon' or 'google'.

dimaopen:
Test!

PR #3727, opened 2023-02-15
Adds tests and bugfixes related to EV charging and parking selection for private vehicles

### LINKED ISSUE #3714: EV charging unit testing
Create a test to cover electric vehicles charging behavior.

HoneyTauOverTwo:
### Low level tests

The tests reside in `ChargingNetworkSpec.scala` and are done by creating a known and orderly arrangement of charging stalls, where for each test, just one characteristic of the charging stall is different from each other. This allows it to separate different behavior for each search parameter. The parameters with a checkmark have related tests implemented.

`ParkingMNL.Parameters.`:
- `WalkingEgressCost` ✔️
- `ParkingTicketCost` ✔️
- `HomeActivityPrefersResidentialParking` ✔️
- `EnrouteDetourCost` ✔️
- `RangeAnxietyCost` ❌

Each test aims to influence just one of these parameters at a time. The expected probability is calculated based on the intended influence of the parameter being tested. Then several calls to `ChargingNetwork.processParkingInquiry` are made to obtain an observed probability for each charging stall. The expected and observed probabilities are compared, if they are close enough the test passes.

The method used to test the parameters have proven to be unsuitable to test the parameters with an ❌.

### High level tests

These are in `ElectricVehicleChargingBehaviorTest.scala` and aim to observe parking selection and charging behaviors on an actual beam simulation instead of just observing the result of parking inquiries to the ChargingNetwork.

Another PR will be created to add similar tests to ride hail electric vehicles.


HoneyTauOverTwo:
test!

HoneyTauOverTwo:
Test now!

HoneyTauOverTwo:
Test now!

HoneyTauOverTwo:
test!

HoneyTauOverTwo:
test!

HoneyTauOverTwo:
test!

HoneyTauOverTwo:
test!

HoneyTauOverTwo:
test!

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
```diff
@@ -97,30 +97,66 @@ class ChargingFunctions(
...
-  def ifHomeWorkOrLongParkingDurationThenSlowChargingOnly(zone: ParkingZone, inquiry: ParkingInquiry): Boolean = {
-    zone.chargingPointType.forall(chargingPointType =>
-      inquiry.beamVehicle.forall {
-        case vehicle
-            if !vehicle.isRideHail && (isHomeWorkOrOvernight(inquiry) || hasLongParkingDurationButNotCharge(inquiry)) =>
-          !ChargingPointType.isFastCharger(chargingPointType)
-        case _ => true
-      }
-    )
+  def ifHomeWorkOrLongParkingDurationThenSlowChargingOnlyUnlessEnrouting(
```
JustinPihony: The unless is unnecessary and confuses the overall. How about
```suggestion
  def ifEnrouting_Home_Work_Or_LongParkingDurationThenSlowChargingOnly(
```

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
```diff
@@ -97,30 +97,66 @@ class ChargingFunctions(
...
+          case _ =>
+            inquiry.parkingActivityType match {
+              case Charge  => true
+              case EnRoute => true
+              case _       => false
+            }
+        }
+      } else {
+        false
+      }
```
JustinPihony:
```suggestion
        val verifyCharging = (
          (inquiry.beamVehicle.isDefined && inquiry.beamVehicle.get.beamVehicleType.chargingCapability.isDefined
      ) 
          && (inquiry.searchMode == ParkingSearchMode.EnRouteCharging || inquiry.parkingActivityType == Charge || inquiry.parkingActivityType == EnRoute) 
```

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
```diff
@@ -182,7 +219,7 @@ class ChargingFunctions(
           origin.time + travelTime1,
           beamVehicle.beamVehicleType
         )
-        ((travelTime1 + travelTime2) / ZonalParkingManager.HourInSeconds) * inquiry.valueOfTime
+        (travelTime1 + travelTime2) * inquiry.valueOfTime / ZonalParkingManager.HourInSeconds
```
JustinPihony: The parentheses should be put back in place. They provide definitive clarity on order of operations. Without them  then future additions could cause problems

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
HoneyTauOverTwo: I think your suggestion implies that if enrouting then slow charging only, which is the opposite behavior.

I added the "UnlessEnrouting" to reflect that the filter now just returns true if we are enrouting, regardless of the other conditions.

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
HoneyTauOverTwo: The reason I changed this was to solve a bug, `travelTime1`, `travelTime2` and `ZonalParkingManager.HourInSeconds` are all integers, and the desired effect is not an integer division, so I moved `inquiry.valueOfTime` (double) to fix it.

I guess I could just do this instead then:

`((travelTime1 + travelTime2) / ZonalParkingManager.HourInSeconds.toDouble) * inquiry.valueOfTime`

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
JustinPihony: But isn't that the end functionality anyway? True means to use slow charging unless I am misreading it.

JustinPihony (approved):
I still disagree on the naming, but not worth blocking

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
HoneyTauOverTwo: This function is part of a filter used to decide if we will allow a particular charging station to be considered as a choice. True means we allow it to be chosen.
when we are enrouting, we only allow fast chargers;
when we are parking for a long time, at home or at work, we only allow slow chargers.

But originally, there was a bug where this filter would return false for everything if we needed to enroute on our way to work or home, because another filter says we need fast chargers, and this would only allow slow ones so they were mutually exclusive.

Basically the true, in this case for enrouting, means we do not care if it is fast or slow because the other filters will take care of it, as opposed to, if we are not searching for a stall to charge during enrouting but we are going to park for a long time, or park at home or work, then we only allow slow chargers.

HoneyTauOverTwo:
test!

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
HoneyTauOverTwo: Perhaps `unlessEnrouting_IfHomeWorkOrLongParkingDurationThenSlowChargingOnly` is a bit clearer? That if you agree that it is the functionality of this.

HoneyTauOverTwo:
test!

src/main/scala/beam/agentsim/infrastructure/ChargingFunctions.scala
JustinPihony: Ultimately this is a really convoluted function. Based on what you said then this should be renamed entirely - to indicate that it selects the charger if true. I broke it down to be a simple boolean expression (feel free to use it if you want - but the `case` might be more readable):

```
lazy val notRideHail_and_homeWorkOvernightOrLongDuration= !vehicle.isRideHail && (isHomeWorkOrOvernight(inquiry) || hasLongParkingDurationButNotCharge(inquiry))
    inquiry.searchMode == ParkingSearchMode.EnRouteCharging || !notRideHail_and_homeWorkOvernightOrLongDuration || 
    	!(notRideHail_and_homeWorkOvernightOrLongDuration && ChargingPointType.isFastCharger(chargingPointType))
```

So, that makes it read to me as super convoluted no matter how I try to word it...so my opinion is to leave the implementation details internal to the method: `allowChargingStationBasedOn(zone, inquiry)`. Otherwise to capture the full implementation it must become something like: `skipChargingStartionIfNotEnroute_notRideHailAndHomeWorkOvernightLongDuration_andIsFastCharger`

Because when I read the implementation further then there is only **1** case where it is false.

HoneyTauOverTwo: I'm changing this to `filterBasedOnChargerSpeed`. I think it is a good middle ground between explaining too much (possibly being convoluted) and explaining too little.

HoneyTauOverTwo:
test!

PR #3734, opened 2023-03-13
Hl/jl/restrict road access by freespeed

jlaz:
Restricts ridehail vehicles from using links with freespeed above a defined speed limit using a link weight multiplier in r5Wrapper. Has been tested using urbansim-1k.


zneedell (approved):
Lgtm!

PR #3735, opened 2023-03-14
Send routing failures to skims

zneedell:
Several fixes to make sure that the skimmer keeps track of situations where no route is possible from an origin to a destination (particularly relevant for transit) so that we don't keep sending agents on that trip without realizing it's impossible. A couple of changes:

- Create a new replanning error code `RouteNotAvailableForChosenMode` for when someone has a mode already chosen but an itinerary of that mode doesn't show up as possible in mode choice. Also added origin (and sometimes destination) location to the replanning event to make it easier understand what happened after the fact. Also create a failedODSkimmerEvent in this situation.
- Renamed `observations` in the skims to clarify the difference between a `completedTrip` and a `failedTrip`
- Also create a FailedActivitySimSkimmerEvent so that we record failed trips there too, where this event increases the number of failedTrips that show up in the skims but don't change the averages for the other values at all. Record this value for all possible asim path types for a given trip (e.g. all possible transit modes for a WALK_TRANSIT trip) because if no transit route was returned that means that R5 couldn't find a transit route across all possible transit paths

~~To do:~~
- ~~Add test coverage~~


zneedell:
Test!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1258,14 +1262,6 @@ trait ChoosesMode {
 
       def isAvailable(mode: BeamMode): Boolean = combinedItinerariesForChoice.exists(_.tripClassifier == mode)
```
dimaopen: This method can also be removed

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -1456,6 +1494,14 @@ class PersonAgent(
       }
   }
 
+//  def createActivitySimSkimmerEvent(
+//    currentActivity: Activity,
+//    nextActivity: Activity,
+//    ODSkimmerEvent: ODSkimmerEvent
+//  ): Unit = {
+//    println("Look!!!")
+//  }
```
dimaopen: Maybe remove?

dimaopen (approved):

zneedell:
Test!

zneedell:
Test now!

zneedell:
Test now!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

PR #3737, opened 2023-03-21
Jupyter notebook for getting beam folders within s3 location

dimaopen:
1. Python script for getting beam output entries from an S3 bucket.
2. Bash scripts to delete/archive s3 entries.

The python script stores the beam locations to a csv file.
It may install boto3 in your python env if it's not there yet.


dimaopen:
Test!

dimaopen:
Test!

aws/src/main/bash/archive_s3.sh
```diff
@@ -0,0 +1,20 @@
...
+    if [[ ${#in} -lt 1 ]];
+    then
+      echo "error: $in"
+    elif [[ "$in" == */ ]];
+    then
+      aws s3 cp --recursive "s3://$bucket/$in" "s3://$bucket/$in" --storage-class DEEP_ARCHIVE
+    else
+      aws s3 cp "s3://$bucket/$in" "s3://$bucket/$in" --storage-class DEEP_ARCHIVE
+    fi
+done < ./archive.csv
```
nikolayilyin: few nitpicks:
1. variable name 'in' should be changed to something like 'input_line' (I had to google to understand if it is a part of shell syntax or not because of the name)
2. a small comment to each `if check` would be great
3. it is required to provide csv as input, but it seems the script process the input file as text lines, is it intended? 
4. providing a short example of input file content somewhere in the script would be great

aws/src/main/bash/delete_s3.sh
```diff
@@ -0,0 +1,17 @@
...
+    if [[ ${#in} -lt 1 ]];
+    then
+      echo "error: $in"
+    elif [[ "$in" == */ ]];
+    then
+      aws s3 rm --recursive "s3://beam-outputs/$in"
+    else
+      aws s3 rm "s3://beam-outputs/$in"
+    fi
+done < ./approved_to_delete.csv
```
nikolayilyin: the same nitpicks as for archieve_s3 script

aws/src/main/bash/move_s3.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/bash
+# get all objects for each entry from archive_test.csv, tar them
+# and put to glacier-beam-outputs bucket with storage class DEEP_ARCHIVE
+
+bucket="beam-outputs"
+arch_bucket="glacier-beam-outputs"
+while read in; do
+    in=$(echo "$in" | xargs)
+    echo "archiving: s3://$bucket/$in"
+    if [[ ${#in} -lt 1 ]];
```
nikolayilyin: the same nitpicks as for archieve_s3 script

jupyter/utility/s3_buckets_analysis.py
```diff
@@ -0,0 +1,331 @@
...
+        pass
+
+
+def import_or_install(package):
+    try:
+        __import__(package)
+        print(f'{package} is already installed')
+    except ImportError:
+        print(f'Installing {package}')
+        pip.main(['install', package])
```
nikolayilyin, line 24: if one runs the notebook from the scratch one cells after another - will it work?

nikolayilyin (changes requested):

aws/src/main/bash/archive_s3.sh
dimaopen: 3. csv file should contain a single colum (no header). csv extension is used because the file is usually is an export result from a spreadsheet or pandas dataframe.

jupyter/utility/s3_buckets_analysis.py
dimaopen: Yes it will Though the cells are more or less independent of each other. They contain comments about what they do.

dimaopen:
Test!

nikolayilyin (approved):

PR #3740, opened 2023-04-12
Jl/update ridehail surge

jlaz:
Updating surge pricing to work for multiple ridehail managers.

### DIFF
--- a/src/main/java/beam/analysis/plots/RideHailRevenueAnalysis.java
+++ b/src/main/java/beam/analysis/plots/RideHailRevenueAnalysis.java
@@ -27,7 +27,7 @@ public class RideHailRevenueAnalysis implements ControlerListener, IterationEnds
     private final Logger log = LoggerFactory.getLogger(RideHailRevenueAnalysis.class);
 
     private final RideHailSurgePricingManager surgePricingManager;
-    static final String fileBaseName = "rideHailRevenue";
+    static final String fileBaseName = "rideHailRevenue_%s";
 
     private OutputDirectoryHierarchy outputDirectoryHiearchy;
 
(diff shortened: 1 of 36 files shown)


src/main/java/beam/analysis/plots/RideHailRevenueAnalysis.java
```diff
@@ -27,7 +27,7 @@ public class RideHailRevenueAnalysis implements ControlerListener, IterationEnds
     private final Logger log = LoggerFactory.getLogger(RideHailRevenueAnalysis.class);
 
     private final RideHailSurgePricingManager surgePricingManager;
-    static final String fileBaseName = "rideHailRevenue";
+    static final String fileBaseName = "rideHailRevenue_%s";
```
nikolayilyin, line 30: let's avoid putting formatting into the variable, instead let's use it in the string.format method directly, this will improve readability

src/main/scala/beam/agentsim/agents/ridehail/RideHailSurgePricingManager.scala
```diff
@@ -10,10 +11,16 @@ import scala.collection.JavaConverters._
 import scala.collection.mutable.ArrayBuffer
 import scala.util.Random
 
-class RideHailSurgePricingManager @Inject() (val beamServices: BeamServices) {
+class RideHailSurgePricingManager @Inject() (
+  val beamConfig: BeamConfig,
+  val beamScenario: BeamScenario,
+  val managerName: String
+) {
```
nikolayilyin, line 18: Using constant link to beamConfig will prevent automatic updates of config in case it was changed during a simulation, instead it is better to get the latest version of config from the beamServices object each time or manually subscribe to changes of beamConfig and updating it each time.

There is the same issue for the any config values that are stored in variables.

jlaz: I see. The issue here is that beamServices is initialized after RideHailSurgePricingManager. And since I am initializing a map of multiple RideHailSurgePricingManager objects, I needed to actually initialize each one, which would error without access to beamServices. @nikolayilyin do you have any suggestions for alternative approaches?

PR #3744, opened 2023-04-24
Link based parking sampling

zneedell:
Rather then generating parking stall locations randomly within a TAZ, this PR adds the functionality of sampling road links (weighted by length) within a TAZ and then generating a stall along one of those. If there is high parking availability in a TAZ, a stall gets created along the nearest link as close as possible to the request point. If there is low parking availability, only a random subsample of links are considered as available, and then a stall is created as close as possible along the closest available link.

This has been partially tested in some of the PILATES and tour mode choice runs -- consolidating all of the changes here and adding tests.

### DIFF
--- a/src/main/scala/beam/agentsim/infrastructure/parking/ParkingStallSampling.scala
+++ b/src/main/scala/beam/agentsim/infrastructure/parking/ParkingStallSampling.scala
@@ -35,7 +35,7 @@ object ParkingStallSampling {
     Some(filteredLinks)
       .filter(_.nonEmpty)
       .map(
-        _.map(lnk => getClosestPointAlongLink(lnk, requestLocation)).minBy(loc =>
+        _.map(lnk => getClosestPointAlongLink(lnk, requestLocation, distanceFunction)).minBy(loc =>
           distanceFunction(loc, requestLocation)
         )
       )
@@ -47,19 +47,48 @@ object ParkingStallSampling {
 //      .getOrElse(requestLocation)
   }
 
-  private def getClosestPointAlongLink(link: Link, requestLocation: Location): Location = {
+  private def getClosestPointAlongLink(
+    link: Link,
+    requestLocation: Location,
+    distanceFunction: (Coord, Coord) => Double
+  ): Location = {
     val (p1, p2) = (link.getToNode.getCoord, link.getFromNode.getCoord)
-    val (dx, dy) = (p2.getX - p1.getX, p2.getY - p1.getY) // vector between p1 and p2
-    val (diffx, diffy) = (requestLocation.getX - p1.getX, requestLocation.getY - p1.getY) // vector between p1 and p3
-    val c1 = (dx * diffx + dy * diffy) / (pow(dx, 2) + pow(dy, 2)) // projection of w onto v
-    if (c1 < 0) { p1 }
-    else if (c1 > 1) { p2 }
-    else {
-      val x = p1.getX + c1 * dx // closest point on line to p3
-      val y = p1.getY + c1 * dy // closest point on line to p3
-      new Coord(x, y)
+    val (closestPointOption, closestDistance) = List(p1, p2).foldLeft(None: Option[Coord], Double.PositiveInfinity) {
+      (accumulator, endPoint) =>
+        val dist = distanceFunction(endPoint, requestLocation)
+        if (dist < accumulator._2) {
+          (Some(endPoint), dist)
+        } else {
+          accumulator
+        }
     }
 
+    closestPointOption match {
+      case Some(closestPoint) =>
+        if (closestDistance < 100) {
+          closestPoint
+        } else {
+          val (dx, dy) = (p2.getX - p1.getX, p2.getY - p1.getY) // vector between p1 and p2
+          val (diffx, diffy) =
+            (requestLocation.getX - p1.getX, requestLocation.getY - p1.getY) // vector between p1 and p3
+          val c1 = (dx * diffx + dy * diffy) / (pow(dx, 2) + pow(dy, 2)) // projection of w onto v
+          if (c1 < 0) {
+            p1
+          } else if (c1 > 1) {
+            p2
+          } else {
+            val x = p1.getX + c1 * dx // closest point on line to p3
+            val y = p1.getY + c1 * dy // closest point on line to p3
+            if ((!x.isNaN) && (!y.isNaN)) {
+              new Coord(x, y)
+            } else {
+              closestPoint
+            }
+          }
+        }
+      case None =>
+        requestLocation
+    }
   }
 
   /**
--- a/src/main/scala/beam/agentsim/infrastructure/parking/ParkingStallSampling.scala
+++ b/src/main/scala/beam/agentsim/infrastructure/parking/ParkingStallSampling.scala
@@ -3,6 +3,7 @@ package beam.agentsim.infrastructure.parking
 import scala.util.Random
 import beam.agentsim.infrastructure.taz.TAZ
 import beam.router.BeamRouter.Location
+import com.typesafe.scalalogging.LazyLogging
 import org.matsim.api.core.v01.network.Link
 import org.matsim.api.core.v01.{Coord, Id}
 import org.matsim.core.utils.collections.QuadTree
@@ -13,7 +14,7 @@ import scala.math.pow
 /**
   * sampling methods for randomly generating stall locations from aggregate information
   */
-object ParkingStallSampling {
+object ParkingStallSampling extends LazyLogging {
 
   val maxOffsetDistance = 600.0 // TODO: Make this a config parameter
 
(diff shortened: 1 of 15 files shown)


src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
```diff
@@ -41,6 +41,9 @@ class TAZTreeMap(val tazQuadTree: QuadTree[TAZ], val useCache: Boolean = false)
   val idToTAZMapping: mutable.HashMap[Id[TAZ], TAZ] = mutable.HashMap()
   private val cache: TrieMap[(Double, Double), TAZ] = TrieMap()
   private val linkIdToTAZMapping: mutable.HashMap[Id[Link], Id[TAZ]] = mutable.HashMap.empty[Id[Link], Id[TAZ]]
+
+  val TAZtoLinkIdMapping: mutable.HashMap[Id[TAZ], QuadTree[Link]] =
```
JustinPihony: Nit: lowercase naming = `tazToLinkIdMapping`

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
```diff
@@ -177,10 +209,10 @@ object TAZTreeMap {
     val quadTreeBounds: QuadTreeBounds = quadTreeExtentFromShapeFile(features)
 
     val tazQuadTree: QuadTree[TAZ] = new QuadTree[TAZ](
-      quadTreeBounds.minx,
-      quadTreeBounds.miny,
-      quadTreeBounds.maxx,
-      quadTreeBounds.maxy
+      quadTreeBounds.minx - 2e4,
```
JustinPihony: You should turn this into a constant value with a name to better understand its purpose. As it stands I do not understand the reasoning for the addition myself - beyond buying some space.

JustinPihony: Same with the 2e3 above

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
```diff
@@ -125,9 +128,22 @@ class TAZTreeMap(val tazQuadTree: QuadTree[TAZ], val useCache: Boolean = false)
 
   def mapNetworkToTAZs(network: Network): Unit = {
     if (tazListContainsGeoms) {
+      val extent =
+        (tazQuadTree.getMinEasting, tazQuadTree.getMinNorthing, tazQuadTree.getMaxEasting, tazQuadTree.getMaxNorthing)
```
JustinPihony: The 2e3 buffer should be done here. I see no need to spin things up continuously. Also, the updating of the `tazToLinkIdMapping` can be done in the `match` directly. No need to toss variables around that aren't utilized other than that.

src/main/scala/beam/agentsim/infrastructure/parking/ParkingStallSampling.scala
```diff
@@ -3,15 +3,100 @@ package beam.agentsim.infrastructure.parking
...
+      case Some(linkQuadTree) =>
+        val allLinks = linkQuadTree.getDisk(requestLocation.getX, requestLocation.getY, maxDist).asScala
+        val totalLength = allLinks.foldRight(0.0)(_.getLength + _)
+        var currentLength = 0.0
+        val filteredLinks = rand.shuffle(allLinks).takeWhile { lnk =>
+          currentLength += lnk.getLength
+          currentLength <= totalLength * availabilityRatio
+        }
+        Some(filteredLinks)
+          .filter(_.nonEmpty)
```
JustinPihony: This can be pushed out and filtered earlier.

JustinPihony: In fact, this is the only reason you would do a `getOrElse` as near as I can tell. I think an if/else would be fine here:

```
if(filteredLinks.isEmpty) //logger.warn....etc
else //filteredLinks.map....etc
```

src/main/scala/beam/agentsim/infrastructure/parking/ParkingStallSampling.scala
```diff
@@ -3,15 +3,100 @@ package beam.agentsim.infrastructure.parking
...
+
+    closestPointOption match {
+      case Some(closestPoint) =>
+        if (closestDistance < 100) {
+          closestPoint
+        } else {
+          val (dx, dy) = (p2.getX - p1.getX, p2.getY - p1.getY) // vector between p1 and p2
+          val (diffx, diffy) =
+            (requestLocation.getX - p1.getX, requestLocation.getY - p1.getY) // vector between p1 and p3
+          val c1 = (dx * diffx + dy * diffy) / (pow(dx, 2) + pow(dy, 2)) // projection of w onto v
```
JustinPihony: Please wrap in parens...I hate relying on order of operations. If somebody adds something without paying attention then it could screw the whole thing up. Also, make sure all values are doubles or else you'll end up rounding inadvertently (unless that is what you want)

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
zneedell: I don't quite follow here. I think one reason to loop over all TAZs first is to ensure that `idToTAZMapping` has an entry for each TAZ (even if the network doesn't have any links in that TAZ) so that I can do a `.get(tazId)` later without worrying about a missing key. Any suggestions for simplifying taking that into account?

I added constants distinguishing between the buffers for each TAZ and for the entire map. I think the map buffer should be included in e.g. `tazQuadTree.getMinEasting`, but I want to be sure to also include a buffer in the TAZ-specific treeMap I'm creating to keep track of link locations

JustinPihony: You actually changed the logic in this recent change. Previously you had subtracted/added 2e3 to the min/max no matter if it was `extent` or from the `env`. In the latest change you only do it if it is from `env`. In the new code my suggestion is something more like this (note that `extent` is reused):

```
val extent =
       new QuadTree[Link](tazQuadTree.getMinEasting, tazQuadTree.getMinNorthing, tazQuadTree.getMaxEasting, tazQuadTree.getMaxNorthing)
    idToTAZMapping.toList.foreach { case (id, taz) =>
        taz.geometry.map(_.getEnvelope.getEnvelopeInternal) match {
          case Some(env) => tazToLinkIdMapping(id) = new QuadTree[Link](
              env.getMinX - tazBoundingBoxBufferMeters,
              env.getMinY - tazBoundingBoxBufferMeters,
              env.getMaxX + tazBoundingBoxBufferMeters,
              env.getMaxY + tazBoundingBoxBufferMeters
            )
          case _ => tazToLinkIdMapping(id) = extent
        }
      }
```

In the old code it would be more like this (same as above, but with the buffer added to the extent):

```
val extent =
       new QuadTree[Link](tazQuadTree.getMinEasting - tazBoundingBoxBufferMeters, tazQuadTree.getMinNorthing - tazBoundingBoxBufferMeters, tazQuadTree.getMaxEasting + tazBoundingBoxBufferMeters, tazQuadTree.getMaxNorthing + tazBoundingBoxBufferMeters)
      idToTAZMapping.toList.foreach { case (id, taz) =>
        taz.geometry.map(_.getEnvelope.getEnvelopeInternal) match {
          case Some(env) =>tazToLinkIdMapping(id) = new QuadTree[Link](
              env.getMinX - tazBoundingBoxBufferMeters,
              env.getMinY - tazBoundingBoxBufferMeters,
              env.getMaxX + tazBoundingBoxBufferMeters,
              env.getMaxY + tazBoundingBoxBufferMeters
            )
          case _         => tazToLinkIdMapping(id) = extent
        }
    }
```

Ultimately my original comment was that extent could be initialized once and then reused over and over, rather than being built, extracted, and rebuilt.

zneedell:
Test!

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
```diff
@@ -163,8 +163,9 @@ class TAZTreeMap(val tazQuadTree: QuadTree[TAZ], val useCache: Boolean = false)
                   logger.warn(e.toString)
               }
               linkIdToTAZMapping += (id -> taz.tazId)
-            case _ =>
+            case None =>
               unmatchedLinkIds += id
+            case _ =>
```
JustinPihony, line 168: Matched, but not car/walk are just ignored then?

JustinPihony (approved):

src/main/scala/beam/agentsim/infrastructure/taz/TAZTreeMap.scala
zneedell: Yep, exactly -- just ignoring them for the purposes of creating parking stalls

PR #3747, opened 2023-05-05
Lawrencium cluster: base functionality

nikolayilyin:
- a beam-environment docker image files - to be able to run any branch with any data without installing any beam dependency
- a shell script to run a job on lawrencium cluster + readme.txt with instruction how to do that
- a small bugfix for our python post-processing script to reduce number of useless output lines in the beamLog


nikolayilyin:
test!

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+    vval="${!vval1}"
+    echo "$v -> '$vval'  (using default value)"
+  else
+    echo "$v -> '$vval'"
+  fi
+done
+
+
+function send_slack_notification() {
+  if [[ -z "$SLACK_HOOK_WITH_TOKEN" ]]; then
```
JustinPihony, line 59: This is listed as required - should it be optional?

JustinPihony: I see other areas - maybe required is a wrong name?

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+
+
+##
+## either pull the code or use the code from mounted folder
+##
+if [ "$PULL_CODE" = true ]; then
+  BEAM_NAME="beam"
+  if [ -z "$BEAM_CODE_URL" ]; then BEAM_CODE_URL="https://github.com/LBNL-UCB-STI/beam.git"; fi
+  echo "Pulling the code from github (PULL_CODE set to '$PULL_CODE'), cloning '$BEAM_CODE_URL' into $(pwd)/$BEAM_NAME"
+  git clone --single-branch --branch "$BEAM_BRANCH_NAME" "$BEAM_CODE_URL" "$BEAM_NAME"
```
JustinPihony: What if BEAM_BRANCH_NAME is not set? Maybe add a check for all required parameters - rather than just printing them out. This is seen in other parts where a "required" variable would crash out in ways that might be harder to debug as compared to an explicit error message.

docker/beam-environment/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM openjdk:8
+RUN apt-get update
+RUN wget -O git-lfs.deb https://packagecloud.io/github/git-lfs/packages/debian/stretch/git-lfs_2.3.4_amd64.deb/download
+RUN dpkg -i git-lfs.deb && rm git-lfs.deb
+RUN apt-get -y install awscli build-essential python3 python3-pandas
+RUN pip install plotly kaleido psutil
+WORKDIR /app
+VOLUME /app/sources
```
JustinPihony: Based on he entrypoint code it seems that you do not want `VOLUME`, but instead a `RUN mkdir -p ./sources` followed by a `COPY ../.. /app/sources` (might need to verify my pathing) - as you are expecting beam to be pre-loaded in the image. `VOLUME` afaik only creates a VOLUME link:
> a volume is a persistent storage location that exists outside of the container

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+
+  if [[ $BEAM_COMMIT_SHA ]]; then
+    echo "Resetting the branch '$BEAM_BRANCH_NAME' to commit '$BEAM_COMMIT_SHA'"
+    git reset --hard "$BEAM_COMMIT_SHA";
+  else
+    BEAM_COMMIT_SHA=$(git log -1 --pretty=format:%H)
+    echo "Using the latest commit in the branch '$BEAM_BRANCH_NAME' - '$BEAM_COMMIT_SHA'"
+  fi
+else
+  echo "Pulling the code is disabled (PULL_CODE set to '$PULL_CODE'), using code mounted to '$PATH_TO_PROJECT_PARENT'."
```
JustinPihony: See my comment in the Dockerfile - I do not believe there will be any source code a this point at all (in the current setup). Also, for parity with the PULL version, should the code be at `/app/sources` or should it be `/app/sources/beam`? If so then you will need to `cd` to that directory here. Although thinking on it more, the git clone in the PULL version would fail if a beam directory already exists...so maybe it is fine? As long as it is considered I am fine with whatever solution. Maybe so that it is the same path(less confusing) for both options, you could either remove the directory in the PULL path and then clone - or cd into it and just make sure it is on the develop branch (or whatever is expected) and then do a git pull? Again, all possible design decisions - I am fine with either as long as they are done purposefully

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+## remember the BEAM working folder location
+##
+BEAM_PATH=$(pwd)
+echo "Working from '$BEAM_PATH'"
+
+
+##
+## Required to avoid configuring ssh keys because by-default all our git data submodules configured to use ssh.
+##
+if [ "$ENFORCE_HTTPS_FOR_DATA_REPOSITORY" = true ]; then
```
JustinPihony, line 140: This is not marked in the required nor optional variable listing

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+
+##
+## pulling data from github if enabled or checking if data location was mounted separately
+##
+if [ "$PULL_DATA" = true ]; then
+  echo "Pulling the data from github (PULL_DATA set to '$PULL_DATA')."
+  production_data_submodules=$(git submodule | awk '{ print $2 }')
+  for i in $production_data_submodules;
+    do
+      if [[ $BEAM_CONFIG  == $i* ]];
```
JustinPihony, line 159: This is not listed in the required env vars - as well as I am not sure how this check works. It does not seem useful to check the beam config path against the names of production modules? Maybe because it is expected to be part of the path, however that would require a double sided wildcard I think.

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+        fi
+        cd - || echo "ERROR: Can't move to the previous location"
+      fi;
+    done
+
+  echo "Doing lfs pull"
+  git lfs pull
+else
+  echo "Pulling the data from github is disabled (PULL_DATA set to '$PULL_DATA')."
+  COMBINED_CONFIG_PATH="/app/data/${BEAM_CONFIG#*/}"
```
JustinPihony: This seems to me to be another case where this would not be filled in under any circumstances given the way the Dockerfile is built. I guess the question is how much pre-loaded scenarios should be put in place? Should we maybe have different images for each main production branch?

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+  else
+    echo "File does not exist, going to use data from inside beam folder."
+  fi
+fi
+
+
+##
+## notification
+##
+if [ "$SEND_NOTIFICATION" = true ]; then
```
JustinPihony: nit - no quotes really needed.

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+  send_slack_notification "Run Started
+    Run Name $TITLED
+    Instance ID $INSTANCE_ID
```
JustinPihony: Some of these seem less useful if being run from docker - as it could in theory even be run locally. That said, maybe just make parts optional instead of the required set?

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+    \"sigopt_dev_id\":\"$SIGOPT_DEV_ID\""
+else
+  echo "Sending notifications is disabled (SEND_NOTIFICATION set to '$SEND_NOTIFICATION')."
+fi
+
+
+##
+## calculating a location for gradle cache
+##
+if [[ -z "$GRADLE_CACHE_PATH" ]]; then
```
JustinPihony: This is another variable that is not noted anywhere else - so maybe add to the optional set?

JustinPihony: Also, should this not also check that the supplied path exists?

docker/beam-environment/entrypoint.sh
```diff
@@ -0,0 +1,362 @@
...
+      final_status="Run Failed"
+  fi
+fi
+echo "The final status of simulation is '$final_status'"
+
+
+##
+## calculating the health of simulation
+##
+warn="?"
```
JustinPihony, line 273: What is the point of these variables other than to fill the spreadsheet? If that is it then it might be better to fill the question marks directly in this case.

docker/beam-environment/run-example-1.sh
```diff
@@ -0,0 +1,29 @@
...
+  --network host \
+  --env MAX_RAM=$MAX_RAM \
+  --env BEAM_CONFIG="$BEAM_CONFIG" \
+  --env BEAM_BRANCH_NAME=$BEAM_BRANCH_NAME \
+  --env PULL_CODE=true \
+  --env PULL_DATA=false \
+  --env S3_PUBLISH=false \
+  --env SEND_NOTIFICATION=false \
+  --mount source="$OUT_PATH",destination=/app/sources,type=bind \
+  "beammodel/beam-environment:latest"
```
JustinPihony, line 28: Going back to another comment - should we have tags for different data branches?

docker/beam-environment/run-example-3.sh
```diff
@@ -0,0 +1,29 @@
...
+
+docker run \
+  --network host \
+  --env MAX_RAM=$MAX_RAM \
+  --env BEAM_CONFIG=$BEAM_CONFIG \
+  --env PULL_CODE=false \
+  --env PULL_DATA=false \
+  --env S3_PUBLISH=false \
+  --env SEND_NOTIFICATION=false \
+  --mount source="$EXISTING_CODE_PATH",destination=/app/sources,type=bind \
```
JustinPihony, line 23: I see here that you are mounting to the volume - but isn't the point to have an image that does not require the local dependencies? And some examples do not set these - but would expect them...

docker/beam-environment/write_cpu_ram_usage.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
JustinPihony, line 1: As this is part of the beam lambda maybe we should save this to the src folder directly and everything can run it instead of now having 2 or 3 places to modify should it ever be needed.

lawrencium/src/main/bash/run_job.sh
```diff
@@ -0,0 +1,144 @@
...
+# By-default it expects 2 input arguments: <run name> <expected execution time>,
+# though both might be manually filled in the script body.
+
+CODE_PHRASE="Execute the body of the job."
+
+# Doing shell magic - the script will send itself as a job to the cluster,
+# and first argument will tell what to do - start a job or execute the job body.
+# This way we could avoid using two shell scripts - one to start a job and another as job body.
+# if the first argument is not what we are looking for - then this shell script is used to start a job
+if [[ "$1" != "$CODE_PHRASE" ]]; then
```
JustinPihony, line 13: This is not documented in the readme - and seems to conflict with `$1` being the run name - not some magic phrase?

lawrencium/src/main/bash/run_job.sh
```diff
@@ -0,0 +1,144 @@
...
+  echo "Starting the job .."
+
+  # what code, data and config to use for simulation
+  export BEAM_BRANCH_NAME="develop"
+  export BEAM_COMMIT_SHA
```
JustinPihony: Why set these to nothing? Just to leave as an option and more visible to the users? Maybe comment it out instead? More of a nit I guess, though.

lawrencium/src/main/bash/run_job.sh
```diff
@@ -0,0 +1,144 @@
...
+    # using the current memory limit as MAX RAM for BEAM simulation
+    export MAX_RAM="$MEMORY_LIMIT"
+  fi
+
+  if [[ -z "$RUN_NAME" ]]; then
+    echo "Error: RUN_NAME is not set."
+    exit 1
+  else
+    # adding current user name as part of simulation title
+    export TITLED="$USER/$RUN_NAME"
```
JustinPihony: $USER is expected to be set in lawrencium?

lawrencium/src/main/bash/run_job.sh
```diff
@@ -0,0 +1,144 @@
...
+
+  RANDOM_PART=$(tr -dc A-Z0-9 </dev/urandom | head -c 8)
+  DATETIME=$(date "+%Y.%m.%d-%H.%M.%S")
+  NAME_SUFFIX="$DATETIME.$RANDOM_PART.$PARTITION.$QOS.$MEMORY_LIMIT"
+
+  # The TEMP directory for this simulation, there will be stored code and data.
+  # The simulation output will be there as well.
+  # /global/scratch will be cleaned after some inactive time, so, it is no advised to be used long-term.
+  BEAM_DIR="/global/scratch/users/$USER/out_beam_$NAME_SUFFIX"
+  mkdir "$BEAM_DIR"
```
JustinPihony: Should this be -p?

lawrencium/src/main/bash/run_job.sh
```diff
@@ -0,0 +1,144 @@
...
+  export SEND_NOTIFICATION="false"
+  export SLACK_HOOK_WITH_TOKEN
+  export SIMULATIONS_SPREADSHEET_UPDATE_URL
+
+  ACCOUNT="pc_beamcore"
+
+  # INPUT Argument #1 - run name
+  RUN_NAME="$1"
+  # INPUT Argument #2 - expected simulation duration
+  EXPECTED_TIME="$2" # D-HH:MM:SS, i.e. for 1 day, 2 hours and 30 minutes => 1-02:30:00
```
JustinPihony: Even in the readme this is not clear. Should it be when the job is to run or how long it is expected to run. I don't understand the use of the latter, but the former would be better named as `SCHEDULED_RUN_TIME` or something more about how it is being scheduled - expected does not convey that.

lawrencium/src/main/bash/run_job.sh
```diff
@@ -0,0 +1,144 @@
...
+  JOB_NAME="$RANDOM_PART.$DATETIME"
+
+  # Two SLURM commands to run a job:
+  #   srun    - run a job directly without returning control
+  #   sbatch  - queue a job and return control to user
+
+  set -x
+  # The last row in this command is the script name itself with a special argument.
+  # See comments to if-else blocks.
+  sbatch --partition="$PARTITION" \
```
JustinPihony, line 110: I assume this will end up triggering the else side at a later time?

JustinPihony (changes requested):
A few areas where I see blocking problems. A number are nits or can be argued against.

For the readme (since it is uncommentable)

MEMORY_LIMIT="480"  ## in G - should be GB to be more standard and explicit
..
export S3_PUBLISH="false" - should be true?
..
Why some exports have examples, but others empty = all should have something if even "ADD_VALUE_HERE" - minor nit

JustinPihony (changes requested):
A few areas where I see blocking problems. A number are nits or can be argued against.

For the readme (since it is uncommentable)

MEMORY_LIMIT="480"  ## in G - should be GB to be more standard and explicit
..
export S3_PUBLISH="false" - should be true?
..
Why some exports have examples, but others empty = all should have something if even "ADD_VALUE_HERE" - minor nit

lawrencium/src/main/bash/run_job.sh
nikolayilyin: Yes, it is set by cluster login node

lawrencium/src/main/bash/run_job.sh
nikolayilyin: Base part of the path (/global/scratch/users/$USER/) should exist and a user won't have permissions to create anything in the base path anyway

lawrencium/src/main/bash/run_job.sh
nikolayilyin: sbatch - is a cluster command which accept a shell script as a parameter, and it is sent there with a special parameter a bit later than this line in order to trigger the else part.

nikolayilyin:
> export S3_PUBLISH="false" - should be true?

I left it as `false` because if set to true then two more variables should contain keys, I will mark it in comment

docker/beam-environment/entrypoint.sh
nikolayilyin: I agree there might be some confusion in paths when the code is pulled and when not - there will be an additional level of folders when git creates a new directory to pull in there.
I was thinking about that and haven't found anything simple and more intuitive for user than what I already did because:
1. removing folders does not sound like a good idea - there might be something inside,
2. existing beam code might be inside of a folder with any name, not exactly `beam`
3. one of requirements was to be able to mount data and code separately and they both might be in the same parent folder.

There will be also some kind of confusion with path to config when data folder was mounted because parent folder name is lost when mounted, but again - I see no clear and intuitive solution for that.

docker/beam-environment/entrypoint.sh
nikolayilyin: fixed

docker/beam-environment/entrypoint.sh
nikolayilyin: Good thing you pointed at this, it appears there is a better solution - to git clone into current folder, which is simplifying paths and reduces confusion.

docker/beam-environment/entrypoint.sh
nikolayilyin: fixed

docker/beam-environment/entrypoint.sh
nikolayilyin: It is a check to find if beam config contains a part of production module paths. It seems double sided wildcard does not required because we want to check if config starts from the same path, not just contains it somewhere in the middle

docker/beam-environment/entrypoint.sh
nikolayilyin: The docker file is built with one or two mounted folders in mind!
I.E. in order to run the image a user has to mount at least an empty folder for pulled code or a folder with checked out BEAM code and \ or a folder with checked out data repository.
It is possible to run the image without mounting anything, but in that case BEAM output won't be easily accessible.

docker/beam-environment/entrypoint.sh
nikolayilyin: This time the quotes are needed to reduce IDEA warnings

docker/beam-environment/entrypoint.sh
nikolayilyin: fixed

docker/beam-environment/entrypoint.sh
nikolayilyin: good catch!

docker/beam-environment/entrypoint.sh
nikolayilyin: The question marks are here as defaults and to mark the possible problem if something went wrong, the real values are calculated after execution of simulation_health_analysis.py by this command - `export "$metric=$count"`

docker/beam-environment/run-example-1.sh
nikolayilyin: The beam-environment image does not contain any data\code at all, everything will be used from mounted folder | pulled to mounted folders during execution

docker/beam-environment/run-example-3.sh
nikolayilyin: We are going to have two docker images - beam-environment and regular beam. The regular version will contain data and compiled code, when environment version will contain only dependency.

docker/beam-environment/write_cpu_ram_usage.sh
nikolayilyin: My Idea to actually have it only here and use the image in lambdas - this way we will have less things in lambdas and only one place with this script and other included in the image things

lawrencium/src/main/bash/run_job.sh
nikolayilyin: This is expected, the script is executed twice - once with the first parameter equal to something else and second time with the first parameter equal to the code phrase.

Here is a simplified example:

`#!/bin/bash`
`if [[ "$1" != "run a job" ]]; then`
	`echo "[$$] Running a job"`
	`export TEST_VAR_1="Parent PID is $$"`
	`bash $0 "run a job"`
`else`
	`echo "[$$] The body of a job"`
	`echo "[$$] TEST_VAR_1 value is '$TEST_VAR_1'"`
`fi`

If you run it, you will see something like that:

`[29307] Running a job`
`[29308] The body of a job`
`[29308] TEST_VAR_1 value is 'Parent PID is 29307'`

lawrencium/src/main/bash/run_job.sh
nikolayilyin: `export BEAM_COMMIT_SHA`  (if environment variable wasn't set before)
is equal to 
`BEAM_COMMIT_SHA=""` 
`export BEAM_COMMIT_SHA`
and equal to 
`export BEAM_COMMIT_SHA=""`

But the latest version is indeed better

lawrencium/src/main/bash/run_job.sh
nikolayilyin: Wrong naming indeed, the expected time should be expected execution time. This is cluster specific thing needed to understand order in which jobs will be executed if there are not enough nodes around.

nikolayilyin:
test!

nikolayilyin:
test!

nikolayilyin:
test!

docker/beam-environment/entrypoint.sh
nikolayilyin: We don't need to check if the folder exist - the ability to provide gradle cache path is for being able to name new folder as required in order to reuse existing folder or to know beforehand where the cache will be.

docker/beam-environment/Dockerfile
nikolayilyin: fixed!

docker/beam-environment/entrypoint.sh
nikolayilyin: fixed

docker/beam-environment/run-example-3.sh
JustinPihony: I do not see a Dockerfile for building the regular version, though. Should that be part of this PR?

docker/beam-environment/entrypoint.sh
JustinPihony: This still doesn't check whether there is something existing in the source and data directory as far as I can tell. So, if you are relying on it being there then it might not..especially since there is no dockerfile for a builtin scenario

docker/beam-environment/entrypoint.sh
JustinPihony: If that is the spec then sounds good to me

docker/beam-environment/entrypoint.sh
JustinPihony: Consolidating this to one (the other) comment - which would be addressed by the creation of a regular image (as noted in that comment)

docker/beam-environment/entrypoint.sh
JustinPihony: There are still notification items that are probably not useful, but they are in the optional field, so I guess it would just be empty - and that is probably fine

docker/beam-environment/run-example-1.sh
JustinPihony: Should there be a check that the folder is filled in? Or it will just crash anyway - so leave it to that?

docker/beam-environment/write_cpu_ram_usage.sh
JustinPihony: Should we create an issue to track this for the future?

JustinPihony (approved):
The open comments should still be addressed, but not worth blocking, so approving

docker/beam-environment/run-example-1.sh
nikolayilyin: It will crash if there is nothing in the mounted folder

docker/beam-environment/entrypoint.sh
nikolayilyin: I expect the mounted folders will either contain code\data or any of `pull` flags to be true, if neither - then the running of the image will fail

docker/beam-environment/run-example-3.sh
nikolayilyin: It is already in develop actually, In this PR I changed the base image and removed unnecessary dependency:
https://github.com/LBNL-UCB-STI/beam/blob/fd0e9c43602bf00d0fd98c7e32b0206c0d7cb521/build.gradle#L737-L761

docker/beam-environment/write_cpu_ram_usage.sh
nikolayilyin: Good idea!
https://github.com/LBNL-UCB-STI/beam/issues/3757

PR #3750, opened 2023-05-19
Scripts: seattle network helpers

heyrutvik:
1) Python script to remove duplicate links
2) Scala script to remove redundant nodes

Related doc: https://ec2-18-116-117-142.us-east-2.compute.amazonaws.com/doku.php?id=home:shapefile_to_osm_convertor

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3750)

### DIFF
--- a/src/main/python/scripts/remove_duplicate_links.py
+++ b/src/main/python/scripts/remove_duplicate_links.py
@@ -17,12 +17,12 @@ writer = shapefile.Writer(args.outputshapefile)
 inbound_links = dict()
 outbound_links = dict()
 
-def adjacent_links(d, key, value):
-    if key not in d:
-        d[key] = set([value])
+def adjacent_links(links_dict, node_id, link_speed):
+    if node_id not in links_dict:
+        links_dict[node_id] = set([link_speed])
     else:
-        s = d[key]
-        s.add(value)
+        s = links_dict[node_id]
+        s.add(link_speed)
 
 for record in reader.iterRecords():
     [from_node, to_node] = record['ID'].split('-')
--- a/src/main/python/scripts/remove_duplicate_links.py
+++ b/src/main/python/scripts/remove_duplicate_links.py
@@ -1,0 +1,33 @@
+import sys
+import shapefile
+
+if len(sys.argv) != 3:
+    print("python remove_duplicate_links.py <input shapefile> <output location>")
+    exit()
+
+reader = shapefile.Reader(sys.argv[1])
+
+writer = shapefile.Writer(sys.argv[2])
+
+writer.field('ID', 'C', 20, 0)
+writer.field('MODES', 'C', 64, 0)
+writer.field('LANES', 'N', 35, 7)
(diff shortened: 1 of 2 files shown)


heyrutvik:
test!

heyrutvik:
test!

src/main/python/scripts/remove_duplicate_links.py
```diff
@@ -0,0 +1,76 @@
...
+
+def adjacent_links(d, key, value):
+    if key not in d:
+        d[key] = set([value])
+    else:
+        s = d[key]
+        s.add(value)
+
+for record in reader.iterRecords():
+    [from_node, to_node] = record['ID'].split('-')
```
nikolayilyin, line 28: is it possible there will be more than two pieces after the split?

src/main/python/scripts/remove_duplicate_links.py
```diff
@@ -0,0 +1,76 @@
...
+
+reader = shapefile.Reader(sys.argv[1])
+writer = shapefile.Writer(sys.argv[2])
+
+# START
+# accumulate adjacent links max_speed
+inbound_links = dict()
+outbound_links = dict()
+
+def adjacent_links(d, key, value):
```
nikolayilyin: If I understood the function correctly, let's give variables sensible names:
d -> links_dict
key -> node_id
value -> link_speed

nikolayilyin (approved):
A nitpick, otherwise looks good

src/main/python/scripts/remove_duplicate_links.py
heyrutvik: No. 

Note: the script is specifically written for Seattle emme shapefile.

src/main/python/scripts/remove_duplicate_links.py
heyrutvik: Ah, sure. I'll update the PR. Thanks for reviewing the PR, @nikolayilyin .

heyrutvik: Done.

src/main/python/scripts/remove_duplicate_links.py
heyrutvik: But I'll add safe guard in subsequent commit. Thanks.

heyrutvik: But on the second thought, lets keep as it is. Becuase if developer will try to use the script other than Seattle emme shapefiles, they will get notified with "not enough (or too many) value to unpack" and they can decide what chages they want to make. Thoughts? @nikolayilyin

nikolayilyin: Giving it a second thought - it is fine to keep it as is, because with another network we might need to change other places in this script as well.

nikolayilyin (approved):
lgtm

PR #3754, opened 2023-06-01
Added storage class attribute to s3 objects

dimaopen:
We need a storage class to filter out the s3 folders that we need to retrieve before we are able to get the data from them.


nikolayilyin (approved):
lgtm

PR #3758, opened 2023-06-23
Jupyter: map analysis for osm.pbf and physsim network

nikolayilyin:


nikolayilyin:
test!

jupyter/map_analysis/comparing_osm_physsim.py
```diff
@@ -0,0 +1,271 @@
...
+pd.set_option('display.max_columns', 500)
+pd.set_option('display.width', 1000)
+
+
+# In[3]:
+
+
+# reading osm file
+
+osm_map = pyrosm.OSM("../beam_root/test/input/sf-light/r5/sflight_muni.osm.pbf")
```
dimaopen, line 29: What if we create an utility method that would check for existence beam_root, and if not found then it would use the project root? That allows not to run jupyter notebooks in a container but directly on the local machine.

dimaopen (approved):
A minor note wichi is up to you to implement.

jupyter/map_analysis/comparing_osm_physsim.py
nikolayilyin: I see no point in that. 
In my understanding the script will be used purely for inspiration and as a source for copy-pasting code.
Once the new code is created the paths will be the once required and not the paths from the script.

PR #3763, opened 2023-07-17
Upgrade Java and related

JustinPihony:
Starting this to be able to use CI

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3763)


JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
Currently stuck on https://github.com/scala/bug/issues/12340 ... Bumping to Scala 2.13 is a hefty task also....

JustinPihony:
Test11!

JustinPihony:
Test11!

JustinPihony:
Test11!

JustinPihony:
Notes for anybody who is looking to continue this work while I am on vacation. Most of the test errors seem to revolve around the new use of `OptionalTime` instead of a simple `double`. So, the first iteration of the upgrade called `.seconds()` directly, which would return the `double` that our code was built around. However, the scenarios where this was undefined no longer yields a `Double.NEGATIVE_INFINITY` by default as it used to. Now it will throw an exception, so must be handled more explicitly. My second attempt for a simple, quick fix was to explicitly call `.orElse(Double.NEGATIVE_INFINITY)` so that it would not throw. However it not only checks on the read, but also the set. So, many of these are cases where this is being set. As such, to avoid the exception, instead of using the pattern `SOMEACTIVITYOBJECT.setTIMEPROPERTY(ENDTIMEOBJECT)` you have to call a specialized set undefined method: `ENDTIMEOBJECT.ifDefinedOrElse(SOMEACTIVITYOBJECT.setTIMEPROPERTY(_), () => SOMEACTIVITYOBJECT.setTIMEPROPERTYundefined())`. There are different iterations of this, but realistically you will likely need to check the tests and fix those AND anywhere that is currently using `orElse(Double.NegativeInfinity)` (or the Java equivalent as above) needs changed to use the more roundabout check for undefined in the set and do it more explicitly.

src/main/java/beam/utils/gtfs/MATSim2GTFS.java
```diff
@@ -143,8 +143,8 @@ public static void main(String[] args) {
                         gtfsStopTime.trip_id = gtfsTrip.trip_id;
                         gtfsStopTime.stop_sequence = stopWithinTrip;
                         gtfsStopTime.stop_id = trs.getStopFacility().getId().toString();
-                        gtfsStopTime.arrival_time = (int) (firstDepartureTime + trs.getArrivalOffset());
-                        gtfsStopTime.departure_time = (int) (firstDepartureTime + trs.getDepartureOffset());
+                        gtfsStopTime.arrival_time = (int) (firstDepartureTime + trs.getArrivalOffset().orElse(0));
```
JustinPihony, line 146: This is a breaking change as it had previously returned a `Double.NEGATIVE_INFINITY`, but it feels like now that we have this `Optional` mechanism we should treat empty as 0 instead of tanking the arrival time. That seems more correct, but should be verified.

src/main/scala/beam/agentsim/agents/PersonAgent.scala
```diff
@@ -499,17 +499,22 @@ class PersonAgent(
...
-      } else {
+    def activityEndTime: Double = {
+      def fallbackActivityEndTime: Double = {
         // logWarn(s"Activity endTime is negative or infinite ${activity}, assuming duration of 10 minutes.")
         // TODO consider ending the day here to match MATSim convention for start/end activity
         tick + 60 * 10
       }
-    val endTime = beamServices.beamScenario.fixedActivitiesDurations.get(activity.getType) match {
+      val endTime = activity.getEndTime
+      var returnVal: Double = fallbackActivityEndTime //Because OptionalTime doesn't have a method which returns - given an fn
```
JustinPihony: Cleaned this up and took the intent, but it is subtly different if anybody wants to double check.

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -143,8 +143,9 @@ object ChoosesParking {
       case (Some(PricingModel.Block(costInDollars, intervalSeconds)), _) =>
         costInDollars / intervalSeconds * SECONDS_IN_HOUR
       case (Some(PricingModel.FlatFee(costInDollars)), Some(activity))
-          if activity.getEndTime - activity.getStartTime > 0 =>
-        costInDollars / (activity.getEndTime - activity.getStartTime) * SECONDS_IN_HOUR
+          if activity.getEndTime.isDefined && activity.getStartTime.isDefined &&
```
JustinPihony, line 146: This is a breaking change, as if either side was undefined then it would be double negative infinity, so an empty start time would really mess with things. So, it may be breaking, but I think it is more correct

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
```diff
@@ -171,7 +172,9 @@ trait ChoosesParking extends {
     val lastLeg = vehicleTrip.last.beamLeg
     val activityType = nextActivity(data).get.getType
     val remainingTripData = calculateRemainingTripData(data)
-    val parkingDuration = nextActivity(data).map(_.getEndTime - lastLeg.endTime).getOrElse(0.0)
+    val parkingDuration =
+      nextActivity(data).map(_.getEndTime.orElse(beam.UNDEFINED_TIME) - lastLeg.endTime).getOrElse(0.0)
```
JustinPihony: This is how the old code would work - but maybe it is worth re-thinking. If endtime is empty then we could yield a 0.0 also? Or maybe this is a scenario where we SHOULD throw an exception?

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -59,7 +59,7 @@ class ChargingNetworkManager(
     beamServices.matsimServices.getScenario.getPopulation.getPersons.asScala.map { case (personId, person) =>
       personId -> (person.getSelectedPlan.getPlanElements.asScala
         .find(_.isInstanceOf[Activity])
-        .map(_.asInstanceOf[Activity].getEndTime)
+        .flatMap(x => Option(x.asInstanceOf[Activity].getEndTime.orElse(null)))
```
JustinPihony: This is another breaking change, but seems more accurate. Of course we can go back to using a negative infinity

src/main/scala/beam/package.scala
```diff
@@ -1,4 +1,3 @@
 package object beam {
-  import beam.agentsim._
-  import beam.utils.OptionalUtils.JavaOptionals._
+  val UNDEFINED_TIME = Double.NegativeInfinity
```
JustinPihony, line 2: This is so that we have a central location in case the backing value is changed.

src/main/scala/beam/replanning/SupplementaryTripGenerator.scala
```diff
@@ -321,9 +337,9 @@ class SupplementaryTripGenerator(
             fuelPrice
           )
         val startingOverlap =
-          (altStart - (additionalActivity.getStartTime - accessTripSkim.time)).max(0)
+          (altStart - (additionalActivity.getStartTime.orElse(beam.UNDEFINED_TIME) - accessTripSkim.time)).max(0)
```
JustinPihony, line 341: This and the ending overlap are the same as they were before, but seem like bugs to me - as the negative cancelation would result in an enormous positive....so maybe use 0 if it is undefined here?

src/main/scala/beam/replanning/utilitybased/ChainBasedTourVehicleAllocator.scala
```diff
@@ -172,27 +167,21 @@ object ChainBasedTourVehicleAllocator {
...
             Option(leg)
               .flatMap(leg => Option(leg.getRoute))
-              .filterNot(route =>
-                Time.isUndefinedTime(Try {
-                  route.getTravelTime
-                }.getOrElse(Double.NegativeInfinity))
-              )
-              .map(_.getTravelTime)
-              .filterNot(Time.isUndefinedTime)
+              .map(_.getTravelTime.orElse(0.0))
```
JustinPihony, line 180: Noting that this is a lot of lines squashed, but I believe this is the intent.

src/main/scala/beam/sim/BeamHelper.scala
```diff
@@ -193,6 +194,10 @@ trait BeamHelper extends LazyLogging with BeamValidationHelper {
           bind(classOf[TerminationCriterion]).toProvider(classOf[TerminationCriterionProvider])
 
           bind(classOf[PrepareForSim]).to(classOf[BeamPrepareForSim])
+
```
JustinPihony, line 197: Noting this in comments as we should remove, right?

src/main/scala/beam/sim/BeamSim.scala
```diff
@@ -524,7 +524,9 @@ class BeamSim @Inject() (
     val activityEndTimesNonNegativeCheck: Iterable[Plan] = persons.toList.flatMap(_.getPlans.asScala.toList) filter {
       plan =>
         val activities = plan.getPlanElements.asScala.filter(_.isInstanceOf[Activity])
-        activities.dropRight(1).exists(_.asInstanceOf[Activity].getEndTime < 0)
+        activities
+          .dropRight(1)
+          .exists(_.asInstanceOf[Activity].getEndTime.orElse(beam.UNDEFINED_TIME) < 0)
```
JustinPihony, line 529: Right now `UNDEFINED_TIME` **is** negative, but it might be worth checking for undefined also. Really this seems to me to be a piece of code that needs optimized as it is only there for logging - and can be simplified into an overall exists or something that will short-circuit if negative - or whatever the preference is? I'll likely turn this into an issue for later

src/main/scala/beam/sim/ConsecutivePopulationLoader.scala
```diff
@@ -141,7 +131,6 @@ class ConsecutivePopulationLoader(
     val peopleToRemove = population.getPersons.keySet().asScala.toVector
     peopleToRemove.foreach { personId =>
       population.removePerson(personId)
-      population.getPersonAttributes.removeAllAttributes(personId.toString)
```
JustinPihony, line 144: This was a piece of the original code changes - I would like to double back to it and find why it was removed though...

src/main/scala/beam/sim/population/DiffusionPotentialPopulationAdjustment.scala
```diff
@@ -151,7 +151,9 @@ object DiffusionPotentialPopulationAdjustment {
   }
 
   def activityDuration(activities: List[Activity]): Double = {
-    if (activities.size < 2) 0 else activities(1).getEndTime - activities.head.getEndTime
+    if (activities.size < 2) 0
+    else
+      activities(1).getEndTime.orElse(beam.UNDEFINED_TIME) - activities.head.getEndTime.orElse(beam.UNDEFINED_TIME)
```
JustinPihony, line 156: This is another area where this keeps the original implementation to what it was, but you could be adding a double negative also.

src/main/scala/beam/utils/csv/writers/PlansCsvWriter.scala
```diff
@@ -96,7 +96,7 @@ object PlansCsvWriter extends ScenarioCsvWriter {
           legRouteType = route.map(_.getRouteType),
           legRouteStartLink = route.map(_.getStartLinkId.toString),
           legRouteEndLink = route.map(_.getEndLinkId.toString),
-          legRouteTravelTime = route.map(_.getTravelTime),
+          legRouteTravelTime = route.flatMap(x => Option(x.getTravelTime.orElse(null))),
```
JustinPihony: This is a breaking change but seems more correct

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -122,7 +122,6 @@ class UrbanSimScenarioLoader(
 
   private def clear(): Unit = {
     scenario.getPopulation.getPersons.clear()
-    scenario.getPopulation.getPersonAttributes.clear()
```
JustinPihony, line 125: Another I'd like to double check

JustinPihony:
test11!

JustinPihony:
test11!

JustinPihony:
test11!

JustinPihony:
test11!

src/test/scala/beam/sim/BeamWarmStartRunSpec.scala
```diff
@@ -42,7 +43,6 @@ class BeamWarmStartRunSpec
       zipIn.close()
 
       val expectedFiles = List(
-        "output_personAttributes.xml.gz",
```
JustinPihony, line 45: This was removed as part of a change that didn't seem that it would be removed - https://github.com/matsim-org/matsim-libs/commit/6bd549399871f2d9b1e9437e9eb343e8009b65b4#diff-ba5f0f7de8618dfd5a5a6791ca473a358b7f7abee88c3ac5f12dbb6c5cccd365L84 Should we re-add it or keep the test without?

JustinPihony: It appears it was removed as an output here - https://github.com/matsim-org/matsim-libs/pull/543 - This should be reviewed in itself

JustinPihony:
test11!

JustinPihony:
test11!

src/test/scala/beam/sim/population/PopulationAdjustmentSpec.scala
```diff
@@ -31,12 +32,14 @@ class PopulationAdjustmentSpec extends AnyWordSpec with Matchers with BeforeAndA
 
   override def afterEach(): Unit = {
     appLogger.detachAppender(appender)
+    persons.foreach(_._2.getAttributes.clear())
```
JustinPihony: The change to things like `...Utils` means that it might be harder to reset things...I am fearful of things breaking without realizing due to this..

JustinPihony:
test11!

JustinPihony (commented):
Closing off this current round of reviews as this comment is important:

`SpatialIndexForRideHailAgentLocationTest` is failing because there [was a breaking change](https://github.com/matsim-org/matsim-libs/pull/394)  - which could be fixed by changing the check on `.value` in `put` and `remove` with a `.equals` instead of `==`. The problem I realized is that this was kind of broken all along - if you look at the values in the old implementation of QuadTree it would not actually update to the new values. So we should discuss what we actually want here - it is also admittedly odd that we are testing a library that is not "ours" - although it helped flag something here.

src/test/scala/beam/sflight/SfLightRouterSpec.scala
```diff
@@ -72,10 +72,10 @@ class SfLightRouterSpec extends AbstractSfLightSpec("SfLightRouterSpec") with In
       assert(response.itineraries.exists(_.tripClassifier == WALK_TRANSIT))
       val transitOption = response.itineraries.find(_.tripClassifier == WALK_TRANSIT).get
       assertMakesSense(transitOption.toBeamTrip)
-      assert(transitOption.totalTravelTimeInSecs == 1116)
+      assert(transitOption.totalTravelTimeInSecs == 1117)
```
JustinPihony, line 75: This is seemingly from a change in the NetworkReader. I did not chase it the whole way, but found that the states changed:

`State{vertex=34422, weight=82, backEdge=93357, durationSeconds=49, distance=62266, idx=4}`

versus 

`State{vertex=34421, weight=80, backEdge=93355, durationSeconds=48, distance=62198, idx=4}`

This comes from the TransportNetwork as loaded via the NetworkReader code.

JustinPihony:
test11!

JustinPihony:
test11!

JustinPihony:
Tests are passing, but there are a number of unhandled events and the following exception:

> 22:52:05 java.lang.NullPointerException: null
> 22:52:05 	at org.matsim.core.population.routes.RouteUtils.calcDistance(RouteUtils.java:167)
> 22:52:05 	at org.matsim.core.scoring.EventsToLegs.handleEvent(EventsToLegs.java:338)
> 22:52:05 	at org.matsim.core.scoring.EventsToLegsAndActivities.handleEvent(EventsToLegsAndActivities.java:68)
> 22:52:05 	at org.matsim.core.events.EventsManagerImpl.callHandlerFast(EventsManagerImpl.java:299)
> 22:52:05 	at org.matsim.core.events.EventsManagerImpl.handle(EventsManagerImpl.java:242)
> 22:52:05 	at org.matsim.core.events.EventsManagerImpl.processEvent(EventsManagerImpl.java:143)
> 22:52:05 	at org.matsim.core.events.ProcessEventThread.run(ProcessEventThread.java:91)
> 22:52:05 	at java.base/java.lang.Thread.run(Thread.java:829)

Which I have traced to be due to `networkRoute.getStartLinkId()` in matsim's `calcDistance` returning Double.Negative_Infinity. The current tests do not fail on this, and they would have had the same problem I would think (most of the code is the same in this regard) - so this could be another effect of the network handling with r5 changing. One thing I could try is to back r5 to what we had it as and see if it resolves things --- if it can work with Java 11+

build.gradle
```diff
@@ -186,14 +201,14 @@ dependencies {
...
+    implementation group: 'org.matsim.contrib', name: 'multimodal', version: '12.0'
+    implementation group: 'org.matsim.contrib', name: 'bicycle', version: '12.0'
 
-    implementation(group: 'org.matsim.contrib', name: 'decongestion', version: '0.11.0-2018w44') {
+    implementation(group: 'org.matsim.contrib', name: 'decongestion', version: '12.0') {
         exclude group: 'org.matsim', module: 'matsim'
     }
 
-    implementation(group: 'com.github.wrashid.matsim', name: 'matsim', version: '0.10.1-beam-15') {
+    implementation(group: 'com.github.LBNL-UCB-STI.matsim', name: 'matsim', version: '12.0-beam-5') {
```
JustinPihony: This is from branch [beam-patches-12.x](https://github.com/LBNL-UCB-STI/matsim/tree/beam-patches-12.x)

build.gradle
```diff
@@ -138,10 +155,8 @@ configurations.all {
 
 dependencies {
 
-    implementation(group: 'com.github.LBNL-UCB-STI', name: 'beam-utilities', version: 'v0.2.19') {
-        exclude group: 'com.github.LBNL-UCB-STI', module: 'r5'
-        exclude group: 'org.matsim', module: 'matsim'
-    }
+    // TODO FIXME
+    implementation(group: 'com.github.LBNL-UCB-STI', name: 'beam-utilities', version: '0.3.4')
```
JustinPihony, line 159: These changes are based off of the master branch - https://github.com/LBNL-UCB-STI/beam-utilities

build.gradle
```diff
@@ -202,7 +217,7 @@ dependencies {
     implementation "com.typesafe.scala-logging:scala-logging_${scalaBinaryVersion}:3.9.0"
     implementation "org.slf4j:log4j-over-slf4j:${slf4jVersion}"
 
-    implementation(group: 'com.github.LBNL-UCB-STI', name: 'r5', version: 'v4.5.3_linkRadiusMeters') {
+    implementation(group: 'com.github.LBNL-UCB-STI', name: 'r5', version: 'v4.8.1_linkRadiusMeters') {
```
JustinPihony, line 220: This is based on changes from [beam-patches-4.8](https://github.com/LBNL-UCB-STI/r5/tree/beam-patches-4.8)

build.gradle
```diff
@@ -381,7 +383,7 @@ tasks.withType(ScalaCompile) {
                                 "-P:scapegoat:disabledInspections:OptionGet:UnsafeTraversableMethods:TryGet:EitherGet:AsInstanceOf:NullAssignment:VariableShadowing:NullParameter:MethodReturningAny:IsInstanceOf:EmptyIfBlock:CatchException:EmptyInterpolatedString:UnnecessaryConversion:CatchThrowable:CollectionIndexOnNonIndexedSeq:AvoidSizeNotEqualsZero:ComparingFloatingPointTypes:UnsafeContains:VarCouldBeVal:CatchThrowable:SuspiciousMatchOnClassObject:VarClosure:RepeatedIfElseBody:LooksLikeInterpolatedString",
                                 "-P:scapegoat:verbose:false",
                                 "-P:scapegoat:consoleOutput:true"]
-    def minimumCompilationParameters = ["-unchecked", "-deprecation", "-feature"]
```
JustinPihony, line 385: Note to self - we should probably add this back in

src/main/java/beam/utils/gis/Network2Shapefile.java
```diff
@@ -13,7 +13,7 @@ public static void main(String[] args) {
          * [3] output network filename
          */
         Network n = NetworkUtils.createNetwork();
-        NetworkReaderMatsimV2 reader = new NetworkReaderMatsimV2(n);
+        MatsimNetworkReader reader = new MatsimNetworkReader(n);
```
JustinPihony, line 16: This is the new MatsimNetworkReader - which goes off the DTD - which is V2 from what I can tell...

src/main/scala/beam/agentsim/infrastructure/ChargingNetworkManager.scala
```diff
@@ -59,7 +60,7 @@ class ChargingNetworkManager(
     beamServices.matsimServices.getScenario.getPopulation.getPersons.asScala.map { case (personId, person) =>
       personId -> (person.getSelectedPlan.getPlanElements.asScala
         .find(_.isInstanceOf[Activity])
-        .map(_.asInstanceOf[Activity].getEndTime)
+        .flatMap(x => x.asInstanceOf[Activity].getEndTime.toOption)
```
JustinPihony, line 62: This another place of possible change (where an empty becomes None instead of Double Negative Infinity)

src/main/scala/beam/utils/OptionalUtils.scala
```diff
@@ -30,4 +32,8 @@ object OptionalUtils {
       */
     def toOption: Option[T] = if (opt.isPresent) Some(opt.get()) else None
   }
+
+  implicit class OptionalTimeExtension(optionalTime: OptionalTime) {
+    def toOption: Option[Double] = if (optionalTime.isDefined) Option(optionalTime.seconds()) else Option.empty[Double]
```
JustinPihony, line 37: We could opt to make this return a Some(Double.NegativeInfinity) for backwards compat.

src/main/scala/beam/sim/ConsecutivePopulationLoader.scala
JustinPihony: I still am not sure as to why this was removed - I am going to re-add it unless @alex-vv recalls the reason for removal.

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
JustinPihony: Also another one I will re-add unless @alex-vv can recall why removed?

JustinPihony:
This is a large PR - so feel free to focus on code that you worked on more recently. Also, the comments I made along the way should be reviewed.

src/main/scala/beam/agentsim/agents/choice/logit/DestinationChoiceModel.scala
```diff
@@ -259,10 +259,11 @@ class DestinationChoiceModel(
   private def getRealStartEndTime(
     activity: Activity
   ): (Double, Double) = {
-    val start = if (activity.getStartTime > 0) { activity.getStartTime }
-    else { 0 }
-    val end = if (activity.getEndTime > 0) { activity.getEndTime }
-    else { 3600 * 24 }
+    val start =
+      if (activity.getStartTime.isDefined && activity.getStartTime.seconds() > 0) activity.getStartTime.seconds() else 0
```
dimaopen: We could use Scala implicits to pimp OptionalTime and add fold/filter methods to it to simplify these expressions. Though it can be done  as the next step.

src/main/scala/beam/agentsim/agents/parking/ChoosesParking.scala
dimaopen: I think it should be 0.0 because the last Activity is usually a Home activity with undefined end time.

src/test/scala/beam/utils/SpatialIndexForRideHailAgentLocationTest.scala
```diff
@@ -30,17 +31,23 @@ class SpatialIndexForRideHailAgentLocationTest extends AnyFunSuite with Matchers
         0.1,
         vehicleCategory = VehicleCategory.Car
       ),
-      SpaceTime(1, 1, 1)
+      origLatestUpdatedLocationUTM
     )
 
     spatialIndex.put(10, 10, rideHailAgentLocation)
     spatialIndex.size() shouldBe 1
+    //Cannot compare the RideHailAgentLocation since it overrides the equals - must check the updated portion
```
JustinPihony, line 40: This is a breaking change started at https://github.com/LBNL-UCB-STI/matsim/commit/978538d5f5d06cf9e6914f3ecadcc2a9b575c126 - we can revert to the previous behavior where it ignored the update altogether - meaning the object would never ACTUALLY be updated.

build.gradle
```diff
@@ -47,13 +48,27 @@ version = '0.8.6.13'
...
+//        languageVersion.set(JavaLanguageVersion.of(11))
+//    }
+//}
+
 
 compileScala.options.encoding = 'UTF-8'
-compileScala.scalaCompileOptions.additionalParameters = ["-Xmx1g"]
+compileScala.scalaCompileOptions.additionalParameters = ["-Xmx1g", "-W"]
+// TODO temporary, drop this after matsim 0.10 -> 12.0 migration is fully complete
+//compileScala.scalaCompileOptions.deprecation = false
```
dimaopen, line 68: Shouldn't we remove this todo? or replace it with a comment why someone want enabling this option?

src/main/scala/beam/replanning/SupplementaryTripGenerator.scala
dimaopen: As I understand these additionalActivities always have start/end time defined because they are in the middle of the person plan.

src/main/scala/beam/sim/BeamHelper.scala
dimaopen: I think you are right.

src/main/scala/beam/utils/csv/writers/PlansCsvWriter.scala
```diff
@@ -96,7 +97,7 @@ object PlansCsvWriter extends ScenarioCsvWriter {
           legRouteType = route.map(_.getRouteType),
           legRouteStartLink = route.map(_.getStartLinkId.toString),
           legRouteEndLink = route.map(_.getEndLinkId.toString),
-          legRouteTravelTime = route.map(_.getTravelTime),
+          legRouteTravelTime = route.flatMap(_.getTravelTime.toOption),
```
dimaopen: Wouldn't it break the contract that all the legRoute* attributes should be defined in case the route is defined?

dimaopen (approved):
That OptionalTime seems to bring a different behavior in Beam. I think for now we have a good try to adopt it. We can later rethink all the places where we meet undefined times at runtime.

src/main/scala/beam/utils/csv/writers/PlansCsvWriter.scala
JustinPihony: I can move it back to what it was in this case, although it is an `Option` and I would think it is acceptable to be empty or else it wouldn't be defined that way?

JustinPihony:
test!

JustinPihony:
test!

JustinPihony:
test11!

JustinPihony:
test11!

JustinPihony:
test11!

JustinPihony:
test11!

dimaopen:
Test!

dimaopen:
test11!

dimaopen:
test11!

JustinPihony:
test11!

JustinPihony:
test11!

dimaopen:
test11!

PR #3764, opened 2023-08-07
Vehicle energy rate basing on the weight of the truck + payload

dimaopen:
* Introduced a jupyter notebook that converts a RouteE data csv to a beam vehicle energy file that is defined in `vehicleTypes.csv`.
* Added possibility to define the vehicle mass range in the vehicle energy file.
* Added `curbWeight` column to vehicle types csv.
*  Introduced `DoubleTypedRange` for more precise double ranging.
* All classes extending `DriveVehicle` should be able to provide the current leg payload.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3764)


src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
```diff
@@ -131,41 +144,54 @@ class ConsumptionRateFilterStoreImpl(
...
-                      s"Keeping first rate of $rawFirstRate and ignoring new rate of $rawRate.",
-                      speedInMilesPerHourBin,
-                      gradePercentBin,
-                      numberOfLanesBin
-                    )
-                  case None => numberOfLanesFilter += numberOfLanesBin -> rate
+              case Some(weightKgFilter) => {
+                weightKgFilter.get(weightKgBin) match {
+                  case Some(numberOfLanesFilter) => {
+                    numberOfLanesFilter.get(numberOfLanesBin) match {
```
haitamlaarabi: @dimaopen Are we always assuming that a value of weightKg exists? even for other vehicles? Same For grade? Because from this code, if there is no weightKgFilter it will ignore the rest of the code.

src/main/scala/beam/utils/BeamVehicleUtils.scala
```diff
@@ -59,6 +59,15 @@ object BeamVehicleUtils {
     }
   }
 
+  def vehcileCategoryToWeightInKg(vehicleCategory: VehicleCategory.VehicleCategory): Double = vehicleCategory match {
```
haitamlaarabi, line 68: @dimaopen can you make this value modifiable through with a config file. The reason is that some places might have different sizes of vehicles.

src/main/scala/beam/utils/BeamVehicleUtils.scala
```diff
@@ -135,6 +148,7 @@ object BeamVehicleUtils {
       seatingCapacity = 4,
       standingRoomCapacity = 0,
       lengthInMeter = 4.5,
+      curbWeightInKg = 2000,
```
haitamlaarabi, line 159: same here, using value from config would be better

haitamlaarabi (changes requested):

src/main/scala/beam/agentsim/agents/vehicles/VehicleEnergy.scala
dimaopen: In this code they are checking that a combination (speed, grade, weight, # of lanes) is unique. If it's not then they log an error (line 156). If the file doesn't contain weight column then a single weight range is created (0, 50000] which should cover any vehicle (line 132).

src/main/scala/beam/utils/BeamVehicleUtils.scala
dimaopen: That's a fallback for missing types. I think it's better not to blow up the config. One could always create a missing type in vehicleTypes.csv.

src/main/scala/beam/utils/BeamVehicleUtils.scala
dimaopen: One can always put vehicle weight value to the vehicle type. Though we could add these params as config values. If you have some appropriate parameter names in mind please let me know.

haitamlaarabi: fair enough. Can you add a comment on top of those values that says these are hard coded values, to overwrite them, point them to the vehicle type file and column.

dimaopen: Done

dimaopen:
Test!

haitamlaarabi (approved):
good to go

dimaopen:
Test!

PR #3794, opened 2023-10-09
DEBUGGING: reverting the fix for RH transit

nikolayilyin:
Remove the branch after checking with @wrashid and @nikolayilyin

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3794)


nikolayilyin:
the feature was fixed in #3807

PR #3806, opened 2023-11-09
Include minimum speed in non-cacc BPR travel time calculations

zneedell:
Previously it was only being used when cacc was enabled.

Also includes some config simplification and variable renaming to make things clearer

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3806)


zneedell:
Test!

haitamlaarabi:
Test!

PR #3813, opened 2023-12-18
Only drive time in car skims

zneedell:
This addresses a bug in the PILATES integration with ActivitySim. For non-car modes, ActivitySim expects the `TIME` field to include both in-vehicle and out-of-vehicle time, but for car modes it expects only in-vehicle time (so, not walking to and from parking spots). This addresses that discrepancy.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3813)


dimaopen (approved):

dimaopen:
Test!

zneedell:
Test!

PR #3814, opened 2023-12-18
Read in experienced plans from previous beam runs and SelectExpBeta from them

zneedell:
Only used in PILATES for now -- allows for BEAM to read in agents' plan memories during a PILATES warmstart and select from possible daily plans based on their utility. Also add a replanning penalty to experienced plans to make it less likely that agents will select plans that involved replanning.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3814)


zneedell:
Test!

zneedell:
Test!

src/main/scala/beam/utils/scenario/PreviousRunPlanMerger.scala
```diff
@@ -127,6 +183,13 @@ object LastRunOutputSource extends LazyLogging {
     Some(filePath).filter(Files.exists(_))
   }
 
+  private def findLatestOutputDirectory(outputPath: Path, dirPrefix: String) = {
```
dimaopen, line 186: We can reuse code from lines 196-199

src/main/scala/beam/utils/scenario/PreviousRunPlanMerger.scala
```diff
@@ -94,24 +124,50 @@ object PreviousRunPlanMerger extends LazyLogging {
...
+    }
   }
 }
 
 object LastRunOutputSource extends LazyLogging {
 
-  def findLastRunOutputPlans(outputPath: Path, dirPrefix: String): Option[Path] = {
+  def findLastRunOutputPlans(outputPath: Path, dirPrefix: String): (Option[Path], Option[Path]) = {
     val plansPaths = for {
       (itDir, itNumber) <- findAllLastIterationDirectories(outputPath, dirPrefix)
```
dimaopen, line 147: The following 7 lines is the same as 157-164. Could we extract a method with a self-explanatory name?

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -43,7 +45,7 @@ class UrbanSimScenarioLoader(
 
   private val rand: Random = new Random(beamScenario.beamConfig.matsim.modules.global.randomSeed)
```
dimaopen, line 47: Maybe remove this val completely? or we need a comment why we keep it.

src/main/scala/beam/utils/scenario/UrbanSimScenarioLoader.scala
```diff
@@ -55,6 +57,31 @@ class UrbanSimScenarioLoader(
     if (wereCoordinatesInWGS) coord else geo.utm2Wgs(coord)
   }
 
+  private def buildAndAddLegToPlan(currentPlan: Plan, planElement: PlanElement): Leg = {
+    val leg = PopulationUtils.createAndAddLeg(currentPlan, planElement.legMode.getOrElse(""))
+    planElement.legDepartureTime.foreach(v => leg.setDepartureTime(v.toDouble))
+    planElement.legTravelTime.foreach(v => leg.setTravelTime(v.toDouble))
+    planElement.legMode.foreach(v => leg.setMode(v))
+    leg.getAttributes.putAttribute("trip_id", planElement.tripId)
```
dimaopen, line 65: This is the only line it differs from `BeamScenarioLoader.buildAndAddLegToPlan`. I think we can reuse the same code.

src/main/resources/beam-template.conf
```diff
@@ -691,7 +691,9 @@ beam.physsim.network.maxSpeedInference {
 beam.physsim.network.removeIslands = "boolean | true"
 
 # Replanning
+beam.replanning.replanningPenaltyInDollars = "double | 100.0"
```
dimaopen, line 694: If we want to make beam more or less an internation product we can use something like `replanningMonetaryPenalty`. Though in this case we should change other references to dollar.

dimaopen (changes requested):
Most of these are up to the developer. But the big pieces of the same code should be reused.

zneedell:
Merged in with #3829

PR #3818, opened 2024-01-18
allow walk access or egress for ride_hail_transit

zneedell:
Previously we only allowed itineraries of rh -> transit -> rh when the mode `ride_hail_transit` was selected. This should allow for either the access or egress leg to be walk. There's nothing explicitly preventing the router from constructing a `walk_transit` itinerary as well, but that one should be filtered out in ChoosesMode if it is not allowed based on the agent's input plans.

This PR also captures RH cost in router so that we don't have as many routes with really long ride-hail legs and short transit legs

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3818)


zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

zneedell:
Test!

src/main/scala/beam/router/r5/R5Wrapper.scala
```diff
@@ -1249,10 +1291,12 @@ class R5Wrapper(workerParams: R5Parameters, travelTime: TravelTime, travelTimeNo
           }
         } else 1f
 
-      (traversalTimeSeconds + (timeValueOfMoney * tollCalculator.calcTollByLinkId(
+      val fare: Double = traversalTimeSeconds / 60.0 * perMinuteCost + edge.getLengthM / METERS_IN_MILE * perMileCost
```
jlaz, line 1308: The fare should include the baseCost; getRideHailManagerCosts already returns the baseCost but getVehicleCosts is not using it/passing it to the travelCostCalculator.

jlaz (changes requested):
See comment on r5Wrapper.travelCostCalculator

src/main/scala/beam/router/r5/R5Wrapper.scala
zneedell: I can't figure out a way to tell R5 to add a fixed cost to the itinerary when starting a RH access/egress leg, if anyone can figure that out it would be great to include.

I _think_ it won't make a big difference, though. R5 is already returning a bundle of options that should include each of RH->TRANSIT->RH, RH->TRANSIT->WALK, and WALK->TRANSIT->RH if they are available. Giving the per-mile and per-minute cost to R5 will let it find the true best alternative between each of these three categories, and within each of these categories there would be the same total baseCost. So as long as R5 isn't preemptively choosing between the different access/egress mode options I think it should be moot.

zneedell:
Test!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1038,77 +1082,89 @@ trait ChoosesMode {
...
-        accessTransitEgress.dropRight(1)
-      }
-      Some(surroundWithWalkLegsIfNeededAndMakeTrip(fullTrip))
+  ): Option[EmbodiedBeamTrip] = {
+    val transitLegs = driveTransitTrip.legs
+      .span(leg => !leg.beamLeg.mode.isTransit)
+      ._2
+      .reverse
+      .span(leg => !leg.beamLeg.mode.isTransit)
+      ._2
```
dimaopen: Technical note `span(...)._2` can be replaced with `dropWhile(...)`

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1038,77 +1082,89 @@ trait ChoosesMode {
     rideHail2TransitAccessResult: RideHailResponse,
     rideHail2TransitEgressResult: RideHailResponse,
     driveTransitTrip: EmbodiedBeamTrip
-  ): Vector[EmbodiedBeamTrip] = {
-    if (
-      rideHail2TransitAccessResult.error.isEmpty &&
+  ): Option[EmbodiedBeamTrip] = {
```
dimaopen: This method was supposed to return multiple RH2T itineraries including pooling/solo for access/egress combinations.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -627,8 +636,29 @@ trait ChoosesMode {
...
+            .asInstanceOf[AttributesOfIndividual]
+
+        modeChoiceCalculator(
+          theRouterResult.itineraries.filter(_.tripClassifier == RIDE_HAIL_TRANSIT).toIndexedSeq,
+          attributesOfIndividual,
+          nextActivity(choosesModeData.personData),
+          Some(currentActivity(choosesModeData.personData)),
+          Some(matsimPlan.getPerson)
+        )
+      } else { None }
```
dimaopen: We could filter the RHT itineraries here and send them to `modeChoiceCalculator`. It returns None if the alternatives are empty.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -627,8 +636,29 @@ trait ChoosesMode {
           routeHistory.rememberRoute(leg.travelPath.linkIds, leg.startTime)
         }
       }
+      // HACKY FIX: The old behavior just chose the first RH_TRANSIT itinerary from all the possible RH_TRANSIT
+      // itineraries produced by R5. But different itineraries may have different combinations of walk/rh for
+      // access/egress, and the egress RH legs in different itineraries may have different starting locations. So we'd
+      // really need to send separate RH requests for each RH_TRANSIT itinerary if we wanted to do it correctly. This
+      // sounds too complicated and costly, so instead we just run a route choice on the RH_TRANSIT itineraries returned
+      // by R5. These don't necessarily have correct wait times, but R5 has been updated to give them appropriate costs.
+      // Once we've chosen the best itinerary we can send requests to the RHM to fill in true costs and wait times
```
dimaopen, line 645: If we generally have the same response from the router (except it may contain WALK as access or egress) then the rest of the `ChoosesMode` should be the same (except we don't need to send RH requests for WALK legs).

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -740,11 +771,25 @@ trait ChoosesMode {
...
         case _ =>
-          choosesModeData.copy(rideHailResult = Some(theRideHailResult))
+          val routingFinished =
+            choosesModeData.rideHail2TransitRoutingResponse.nonEmpty &&
+            choosesModeData.routingResponse.nonEmpty &&
+            choosesModeData.rideHail2TransitAccessResult.nonEmpty &&
+            choosesModeData.rideHail2TransitEgressResult.nonEmpty
+          choosesModeData.copy(
+            rideHailResult = Some(theRideHailResult),
+            routingFinished = choosesModeData.routingFinished || routingFinished
```
dimaopen, line 786: I remeber introducing this `routingFinished` but I don't see it is used now. Am I missing something?

dimaopen (commented):
Maybe it's time to refactor ChoosesModeData. Instead of multiple requestIds and Optional responses that also may contain a virtual "Error" not requested, we could have a map similar to `parkingResponses` map. That map contains all the responses from the router and the RHM. And when all the required keys are presented in the map then we can do our caclulation and finish the mode choice?

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
zneedell: Aah, that makes sense, thanks. Hopefully the updates should capture that behavior now

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
zneedell: I'm imagining a situation where the router returns two separate WALK->TRANSIT->RH itineraries, where the two egress RH legs pick up in different locations (e.g. at different transit stops). As I, understand it doing this would allow for only one RH request for the egress leg in this case, even though it could start from multiple locations. I'm not actually sure what the consequences of that would be (only having a maybe incorrect estimated wait time for the egress leg?), but it seems to me like we'd still be missing something.

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
zneedell: It seems like it's still used here, although I haven't fully understood the logic

https://github.com/LBNL-UCB-STI/beam/blob/e8acab10604ccf6a7c1f258b98e7c22bddea1dc4/src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala#L851

zneedell:
Test now!

zneedell:
Test now!

zneedell:
Test now!

src/main/scala/beam/router/r5/R5Wrapper.scala
zneedell: Resolving this after confirming that it does work as expected -- R5 only uses this travelCostCalculator to determine the best routes given a sequence of legs (e.g. RH->TRANSIT->WALK or RH->TRANSIT->RH), so assuming that the base fare is constant geographically (which it is for now) it should work as intended

zneedell:
Test!

zneedell:
Test!

beambot2:
test now!

PR #3858, opened 2024-04-08
Fuel consumption fix

### LINKED ISSUE #3842: BEAM is not calculating energy consumption for rail-based services
BEAM is not calculating energy consumption for rail-based services like subway, light rail, train, cable car.

I think I already discussed that with Rashid, but I don't remember if we fixed that. The problem was related to the fact we are only calculating the energy consumption on the streets - or overriding the off-street links calculation somehow.

Thank you !

dimaopen:
Now PTE has consumed energy data.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3858)


cristian-poliziani:
Sounds great, thank you !

cristian-poliziani (approved):

PR #3880, opened 2024-06-21
OD Skimmer for vehicle types

### LINKED ISSUE #3877: OD Skimmer for Freight vehicle types
We need to add a new Skimmer for freight, it should be an OD skimmer but the keys instead of mode, we need to use vehicle types (or vehicle category/class + powertrain)

dimaopen:
Created a new Origin-Destination skimmer that also uses the vehicle type as a key part. Agent sends the skimmer event for all the car trips. One may filter skims by vehicle type to get only freight vehicles.
In case the parameter `beam.router.skim.sendNonChosenTripsToSkimmer` is set to true (which is default) then non-choosen trip data is also sent to the skimmer.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3880)


dimaopen:
Test!

dimaopen:
Test!

haitamlaarabi (changes requested):
> *[Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3880)* status: 0 of 19 files reviewed, 1 unresolved discussion (waiting on @dimaopen)

___
*[`src/main/scala/beam/agentsim/agents/PersonAgent.scala` line 1627 at r1](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3880#-O0H0SBs9Frgb-kS9L5C:-O0H0SBt38uURqrNdUNw:b-7ovpfn) ([raw file](https://github.com/LBNL-UCB-STI/beam/blob/95499227754149a3fefe7c01d6f040687bc33c0f/src/main/scala/beam/agentsim/agents/PersonAgent.scala#L1627)):*
> ```Scala
>       eventsManager.processEvent(DriveTimeSkimmerEvent(tick, beamServices, carLeg))
>     }
>     if (!failedTrip && correctedTrip.tripClassifier == BeamMode.CAR) {
> ```

Can you make two things configurable:
1- 3 options: Freight, Passenger or Freight & Passenger
If freight, then skimmer only generate freight skims

2- 3options: VehicleTypeId, VehicleCategory, or VehicleCategory+Powertrain
If vehicle type, then key is vehicle type ID, if vehicle category then VehicleCategory is the key, if VehicleCategory+Powertrain, then both Vehicle Category and fuel type.
Fuel type is PrimaryFuelType + SecondaryFuelType

dimaopen:
Test!

dimaopen:
Test!

haitamlaarabi (approved):
> Reviewed 13 of 19 files at r1, 2 of 6 files at r3, 9 of 9 files at r4, all commit messages.
> *[Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3880)* status: all files reviewed, 1 unresolved discussion (waiting on @dimaopen)

PR #3885, opened 2024-07-19
Hl/#3839 BEAM Emissions 

haitamlaarabi:
In this PR we preprocess frism and emfac population, map vehicle types and then assign emissions rates the same way we did for energy rate with RouteE.
We also create interface similar to energy calculation but for emissions in BEAM. We add emissions to PathTraversals and ParkingLeaveEvents. We also produce link level skims for emissions.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3885)


haitamlaarabi:
test!

src/main/resources/beam-template.conf
```diff
@@ -471,6 +471,16 @@ beam.exchange {
...
   }
   output {
+    emissions {
+        # this is the list of emissions to filter out among
+        # "CH4", "CO", "CO2", "HC", "NH3", "NOx", "PM", "PM10", "PM2_5", "ROG", "SOx", "TOG"
+        #@optional
+        emissionsToFilterOut = [string] | []
+        # this will embed emissions profiles in the following events: PathTraversalEvent, PersonEntersVehicleEvent, LeavingParkingEvent
+        events = "boolean | true"
+        # this will produce link levek skims through emissions-skimmer
```
nikolayilyin: typo

src/main/resources/beam-template.conf
```diff
@@ -471,6 +471,16 @@ beam.exchange {
...
+        skims = "boolean | true"
```
nikolayilyin: might there be any performance issues if this settings is on? would it worth turning it off by-default to increase speed of simulations that not need new skims?

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
```diff
@@ -836,7 +835,6 @@ class RideHailAgent(
       }
       isOnWayToParkAtStall match {
         case Some(stall) =>
-          currentBeamVehicle.useParkingStall(stall)
```
nikolayilyin, line 837: I do not see anything calling useParkingStall nearby.
So, just to confirm this row was not removed by accident

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
```diff
@@ -300,11 +286,29 @@ class BeamVehicle(
...
+      beamServices.beamScenario.vehicleEmissions.getEmissionsProfileInGram(
+        vehicleActivityData,
+        vehicleActivity,
+        beamVehicleType,
+        beamServices
+      )
+    emissionsRWLock.write(emissions.map(_.values.foreach { case (source, rates) =>
+      emissionsProfileInGramInternal.values.get(source).map(_ + rates)
+    }))
+    if (beamServices.beamConfig.beam.exchange.output.emissions.events) emissions else None
```
nikolayilyin: we could move this check before emissions calculation and avoid calculation and RWLock altogether if configuration not set and emissions calculation not needed elsewhere

src/main/scala/beam/agentsim/agents/vehicles/VehicleEmissions.scala
```diff
@@ -0,0 +1,637 @@
...
+    value: Double,
+    getDefaultInterval: Boolean = true
+  ): Option[(DoubleTypedRange, T)] = {
+    Try {
+      map.view
+        .filter(_._1.has(value))
+        .maxBy { case (range, _) =>
+          (range.upperBound - range.lowerBound) * (if (getDefaultInterval) 1 else -1)
+        }
+    }.toOption
```
nikolayilyin: if this function is called often and the try-catch block most of the time catch exception then the accumulated execution time of this function will be up to 2 times longer than a similar function with if-else instead of try.toOption

nikolayilyin (approved):
few nitpicks, lgtm otherwise

src/main/resources/beam-template.conf
haitamlaarabi: Done

src/main/scala/beam/agentsim/agents/ridehail/RideHailAgent.scala
haitamlaarabi: the function after it: parkAndStartRefueling, already include currentBeamVehicle.useParkingStall(stall). It was redundant.

src/main/scala/beam/agentsim/agents/vehicles/BeamVehicle.scala
haitamlaarabi: Actually, we have two outputs: skims and events. I adjust this function to make it clearer

src/main/scala/beam/agentsim/agents/vehicles/VehicleEmissions.scala
haitamlaarabi: Good point

haitamlaarabi:
test now!

haitamlaarabi:
test now!

haitamlaarabi:
Test now!

nikolayilyin:
test!

nikolayilyin:
test!

haitamlaarabi:
test now!

dimaopen:
Test!

haitamlaarabi:
Test!

beambot2:
test now!

haitamlaarabi:
Test now!

beambot2:
test!

beambot2:
test!

beambot2:
test!

beambot2:
test!

PR #3904, opened 2024-10-31
copy of Tour Mode Clean Develop 202403

beambot2:
A copy created in an attempt to fix strange CI behavior

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3904)


beambot2:
test!

src/main/scala/beam/agentsim/agents/modalbehaviors/ChoosesMode.scala
```diff
@@ -1704,6 +1944,504 @@ trait ChoosesMode {
...
+          case Some(tripMode) =>
+            // If trip mode is already set, determine tour mode from that and available vehicles (sticking
+            // with walk based tour if the only available vehicles are shared)
+            val chosenTourModeAndVehicle =
+              getTourModeAndVehicle(
+                firstLegItineraries.filter(_.tripClassifier == tripMode),
+                availableVehicles,
+                choosesModeData.personData.currentTourPersonalVehicle
+              )
+            val (chosenTourMode, chosenTourVehicleMap) = chosenTourModeAndVehicle.headOption match {
```
wrashid, line 2377: We are using headOption on a map here. Maps usually have random order. This means, we are returning a random value. If different behaviour is needed, some sorting might be needed.

src/main/scala/beam/router/Modes.scala
```diff
@@ -241,3 +251,237 @@ object Modes {
...
+
+  object BeamTourMode extends StringEnum[BeamTourMode] with StringCirceEnum[BeamTourMode] with ExponentialLazyLogging {
+
+    override val values: immutable.IndexedSeq[BeamTourMode] = findValues
+
+    def getTourModeAndVehicle(
+      trips: Vector[EmbodiedBeamTrip],
+      availableVehicles: Vector[VehicleOrToken] = Vector.empty[VehicleOrToken],
+      currentTourPersonalVehicle: Option[Id[BeamVehicle]] = None
+    ): mutable.Map[Option[BeamTourMode], mutable.Map[EmbodiedBeamTrip, Option[BeamVehicle]]] = {
```
wrashid, line 310: We are returning a map here. The keys of this Map is Option[BeamTourMode]. But we are only putting Some objects there. So Option is not really needed here.

beambot2:
test!

beambot2:
original tour mode branch was merged, no need for this one

PR #3912, opened 2025-04-04
Read Parquet inputs from ActivitySim

zneedell:
The existing "parquet" input mode only worked for the (depreciated) "urbansim" input type, not the "urbansim_v2" input currently used in PILATES. (side note, we should maybe update the names here). This repeats the csv functionality for parquet inputs in urbansim_v2

Also added a test to confirm that the scenarios generated by the parquet and csv readers are identical for identical inputs

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3912)


zneedell:
Test!

PR #3915, opened 2025-04-29
New Freight features and PILATES improvements

zneedell:
This PR includes several updates/fixes that have been happening in parallel

-- General stickiness fixes for rare situations for person agents
*  Includes some fixes for situations with missed transit connections
*  Also includes some improvements for tracking vehicles across complicated tours

-- Some refactoring of the mode choice process
* Better encapsulation of the process of building/sending routing requests
* Ability to re-use routing responses in some cases when the preferred mode isn't available, rather than needing to re-start routing from scratch
* Configuration parameter allowing agents to request routes for all modes regardless of planned mode (useful for filling in transit skims early in a PILATES run)

-- Performance improvements
* Array rather than map lookup for link travel times
* Stop storing path traversal event characteristics in duplicate data structures

-- Some improved ActivitySim coupling
* Read ActivitySim plans in parquet format, which is natively produced by ActivitySim 1.x
* Ability to read expected travel time and cost of planned trips from plans and then to score daily experienced plans based on the *difference* between experienced and expected utility rather than just the experienced utility

-- Freight parking improvements

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3915)


zneedell:
Test now!

PR #3925, opened 2026-03-04
Fixes for drive transit and SingleModeSpec

zneedell:
## Summary
This PR improves drive-transit routing robustness/performance in the R5 integration and updates `SingleModeSpec` to be faster and more regression-sensitive.

## Why
`SingleModeSpec` was slow/flaky and was not strongly guarding against regressions in drive-transit viability.  
We also found drive-transit access stop search behavior that could prematurely stop and reduce feasible itineraries.

### *Old issue with drive-transit routes*
The old heuristic was effectively treating “I can reach the destination link” as “stop searching for transit access stops now.” Previously:

1. For transit requests, BEAM first runs an R5 street search to collect candidate access stops (this is what `StopVisitor` is doing).
2. The old logic had a hard stop at `destinationSplit` (the two street vertices of the destination link).
3. As soon as the search frontier touched that split, the search terminated, even if it had collected zero (or too few) transit stops for drive access.

Why that was a problem for drive-transit:

1. Drive-transit often needs a stop that is not on the “direct-to-destination” street path.
2. The search is graph expansion, not “enumerate all useful stops within a conceptual radius.”
3. So reaching destination early on one branch could cut off exploration of other branches that would have found viable park-and-ride access stops.
4. Result: many requests looked like “no access stops,” not because no good drive-transit existed, but because we stopped searching too early.

The fix changed destination-split from a hard stop to a soft stop:
- after first hitting destination split, keep exploring until either:
1. we found at least `N` access stops (`continue_if_stops_below`, default `1`), or
2. we spent an extra vertex budget (`extra_vertices_after_hit`, default `5000`).

### Routing logic and performance
- Improved `StopVisitor` and related access-stop search behavior for drive-transit in:
  - `src/main/scala/beam/router/RoutingWorker.scala`
  - `src/main/scala/beam/router/r5/R5Wrapper.scala`
- Added configurable destination-split continuation controls for drive-transit access search (env-driven, defaulting to values that improve coverage while keeping runtime reasonable).
- Added/kept lightweight, env-gated diagnostics for drive-transit outcomes in `R5Wrapper`/`RoutingWorker`.
- Removed no-longer-needed break-reason string tracking in `StopVisitor`.

### Test updates (`SingleModeSpec`)
- Reduced sample size from `0.5` to `0.25` for faster CI/runtime.
- Removed brittle RouteDumper-based request/response instrumentation plumbing from the spec.
- Tightened drive-transit assertions to catch regressions:
  - `driveTransitDepartures > 0`
  - `driveShareVsWalk > 0.30`
  - `driveArrivalRate > 0.85`
- Hardened bike-transit assertions to avoid map-key exceptions and fail with clearer diagnostics.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3925)

### DIFF
--- a/src/main/scala/beam/router/BeamRouter.scala
+++ b/src/main/scala/beam/router/BeamRouter.scala
@@ -265,7 +265,7 @@ class BeamRouter(
     case odSkimmerReady: ODSkimmerReady =>
       odSkimmer = Some(odSkimmerReady.odSkimmer)
     case routingResp: RoutingResponse =>
-      if (shouldWriteR5Routes(currentIteration) || beamScenario.beamConfig.beam.routing.writeRoutingStatistic)
+      if (shouldWriteR5Routes(currentIteration))
         eventsManager.processEvent(RouteDumper.RoutingResponseEvent(routingResp))
 
       val updatedRoutingResponse: RoutingResponse = odSkimmer
--- a/src/main/scala/beam/router/BeamRouter.scala
+++ b/src/main/scala/beam/router/BeamRouter.scala
@@ -265,7 +265,7 @@ class BeamRouter(
     case odSkimmerReady: ODSkimmerReady =>
       odSkimmer = Some(odSkimmerReady.odSkimmer)
     case routingResp: RoutingResponse =>
-      if (shouldWriteR5Routes(currentIteration))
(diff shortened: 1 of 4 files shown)


zneedell:
Test!

PR #3928, opened 2026-03-31
tour vehicle handling improvements

zneedell:
## Summary
- Fix tour vehicle handling so personal vehicles are preserved across replanning and home-boundary transitions instead of being dropped too early.
- Retry final `drive_transit` / `bike_transit` egress routing before abandoning the held vehicle, reducing false vehicle drops at the end of a tour.
- Harden tour-vehicle cleanup and release logic, including better recovery/clearing of stale vehicle references and safer handling of shared teleportation vehicles.
- Added code comments clarifying an intentional vehicle-precedence rule during replanning: `currentTourPersonalVehicle` may continue to represent an inherited parent-tour vehicle even when the current subtour strategy is reset, so failed subtour replanning does not force abandonment of the enclosing tour vehicle.

## Testing
- Added regression coverage for final intermodal egress retry behavior.
- Added regression coverage for keeping a personal vehicle available for a later tour after returning home.

- - -
This change is [Reviewable](https://reviewable.io/reviews/LBNL-UCB-STI/beam/3928)


zneedell:
Test!

